PR #5, opened 2021-09-03
Fix panic on Alsa host when playback with minimum config fails

### LINKED ISSUE #1: Showcase example: OutputBuilder::build panicks with ALSA device
Running showcase example on ALSA device, OutputBuilder::build panicks

```rust
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: BackendSpecific { err: BackendSpecificError { description: "ALSA function 'snd_pcm_hw_params_set_buffer_size' failed with error 'EINVAL: Invalid argument'" } }', src/io.rs:81:10
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/panicking.rs:515:5
   1: core::panicking::panic_fmt
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/panicking.rs:92:14
   2: core::result::unwrap_failed
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/result.rs:1355:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/result.rs:1037:23
   4: web_audio_api::io::OutputBuilder::build
             at ./src/io.rs:64:22
   5: web_audio_api::context::AudioContext::new
             at ./src/context.rs:293:22
   6: showcase::main
             at ./examples/showcase.rs:7:19
   7: core::ops::function::FnOnce::call_once
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```
If i set BUFFER_SIZE to 256, the program doesn't panick anymore.

But another bug appears. An assertion fails in debug mode (underflow in release mode)

```rust
thread 'cpal_alsa_out' panicked at 'assertion failed: `(left == right)`
  left: `78`,
 right: `0`', src/graph.rs:117:9
```
I think due to [#495](https://github.com/RustAudio/cpal/issues/495) build_output_stream is not called at fixed rate on ALSA devices.

Jerboas86:
When `OutputBuilder` fails to build a `Stream` with minimum output config on Alsa Host, `OutputBuilder` fallbacks to default config. 

Futher improvements (not included in this PR):
- Homogenize `OutputBuilder::build` and `InputBuilder::build` return type
- Implements a progressive fallback instead of going directly into default config

Partial fix #1


src/io.rs
```diff
@@ -72,11 +74,50 @@ impl OutputBuilder {
...
+                .next()
+                .expect("no supported config?!")
+                .with_max_sample_rate()
+        };
+
+        let sample_format = supported_config.sample_format();
+
+        // determine best buffer size. Spec requires BUFFER_SIZE, but that might not be available
+        let mut buffer_size = match supported_config.buffer_size() {
+            SupportedBufferSize::Range { min, .. } => crate::BUFFER_SIZE.max(*min),
```
orottier, line 106: I'm worried some backends don't support any multiple of 128 and this will fail too.
Maybe in the `default` implementation, we simply do not fiddle with the buffer_size and let cpal choose it. Combined with your other work in #7 that should work 100% of the time

src/io.rs
```diff
@@ -72,11 +74,50 @@ impl OutputBuilder {
...
+        let device = host
+            .default_output_device()
+            .expect("no output device available");
 
-        stream
+        let mut supported_configs_range = device
+            .supported_output_configs()
+            .expect("error while querying configs");
+
+        let supported_config = if cfg!(target_os = "linux") {
```
orottier, line 91: Why switch behaviour per target_os?

src/io.rs
Jerboas86: A large cpal default buffer size will cost a delayed start but after that the rendering should be as responsive as the rendering buffer size namely 128 (if i understand it correctly).

I think you're right we should go directly on default config for every host. 

If cpal makes Fixed behavior more consistent between host in the future, this should be reconsidered.

src/io.rs
Jerboas86: sure, not required

src/io.rs
Jerboas86: I change my mind, i let fixed buffer size. still possible to change if bugs are reported. For now, it works and give the best perfs.\
let me know what you think

Jerboas86:
Is there any works needs to be done for that PR to be merged ?
without it, Alsa host are still panicking for BackendSpecificError reason.
Let me know so i can work on it

orottier:
I'm taking your work and making some changes on it, will update the PR soon. Thanks for your effort

orottier:
I need a bit more time to port the same functionality for the microphone input. Stay tuned

orottier:
Closing in favour of https://github.com/orottier/web-audio-api-rs/pull/14 which builds upon your work. Care to have a look at that one and see if it works on your device?

Jerboas86:
> Care to have a look at that one and see if it works on your device?

I'm finishing my OscillatorNode PR and i will take a look. Sure

Issue #9, opened 2021-09-03
linear_ramp_to_value_at_time applies new value immediatly at endtime


Jerboas86:
**Observed:** 

`linear_ramp_to_value_at_time(v,end)` sets the parameter value **immediatly** at time specified in end argument.

**Expected:** 

from [web audio api spec](https://www.w3.org/TR/webaudio/#:~:text=linearRampToValueAtTime(value%2C%20endTime,the%20given%20value):

> linearRampToValueAtTime(value, endTime)
> 
> Schedules a linear continuous change in parameter value from the previous scheduled parameter value to the given value

Example to reproduce

Run

```rust
use web_audio_api::context::{AsBaseAudioContext, AudioContext};
use web_audio_api::node::{AudioNode, AudioScheduledSourceNode};

fn main() {
    let context = AudioContext::new();

    let osc1 = context.create_oscillator();
    osc1.connect(&context.destination());
    osc1.frequency().set_value_at_time(440., 0.);
    osc1.start();

    osc1.frequency().linear_ramp_to_value_at_time(220., 0.4);
    std::thread::sleep(std::time::Duration::from_millis(1000));
}
```

And run

```rust
use web_audio_api::context::{AsBaseAudioContext, AudioContext};
use web_audio_api::node::{AudioNode, AudioScheduledSourceNode};

fn main() {
    let context = AudioContext::new();

    let osc1 = context.create_oscillator();
    osc1.connect(&context.destination());
    osc1.frequency().set_value_at_time(220., 0.4);
    osc1.start();

    osc1.frequency().set(220., 0.4);
    std::thread::sleep(std::time::Duration::from_millis(1000));
}
```

No difference in scheduling the paramter change is observed.

Jerboas86:
In node.rs

```rust
impl AudioProcessor for OscillatorRenderer {
    fn process(
        &mut self,
        _inputs: &[crate::alloc::AudioBuffer],
        outputs: &mut [crate::alloc::AudioBuffer],
        params: AudioParamValues,
        timestamp: f64,
        sample_rate: SampleRate,
    ) {
        
        [...]

        let freq_values = params.get(&self.frequency);
        let freq = freq_values[0]; // force a-rate processing

       [...]
        }
    }
```

i think that ` linear_ramp_to_value_at_time` doesn't apply cause freq is a scalar value.

orottier:
Fixed in #19

PR #15, opened 2021-09-23
New Oscillator rendering based on wavetables and polyBLEP

Jerboas86:
## Updated oscillator renderers
- `SineRenderer` : New Sine renderer based on **wavetable**
-  `SquareRenderer` and `SawtoothRenderer`: New Square and Sawtooth renderers based on **wavetable** and **polyBLEP**

## New oscillator renderers
- `TriangleRenderer` : New Triangle renderer based on **wavetable** and **polyBLEP**
-  `CustomRenderer`: New Custom renderer based on **wavetable**

### Goals
- Wavetable might be a more efficient algorithm than transcendental function (sin,cos,...)
- Wavetable might allowa more efficient implementation of `custom` oscillator type
- polyBLEP algorithm might reduce aliasing behavior for non bandlimited signal (`square`, `triangle`, and `sawtooth`).

## Todos
- `detune` has not been implemented
- Controls of `OscillatorNode` over `CustomRenderer` is partial. `set_periodic_wave` is currently a no-op
-  Algorithm correctness has been tested but not exhaustively, **and** tests are still not included
-  **No benchmarking between the original rendering implementation and the new one has been done yet**

I 'd like to know what you think about this first draft and improve on you remarks.
I don't know how to pass `PeriodicWave` between the controller and renderer nodes. I thought to send the data via a lock-free channel. 

### Docs
- [polyBLEP doc1](https://www.kvraudio.com/forum/viewtopic.php?t=375517)
- [polyBLEP doc2](http://www.martin-finke.de/blog/articles/audio-plugins-018-polyblep-oscillator/)

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -13,7 +13,7 @@ use crate::buffer::{AudioBuffer, ChannelConfigOptions, ChannelCountMode, Channel
 use crate::graph::{NodeIndex, RenderThread};
 use crate::media::{MediaElement, MediaStream};
 use crate::message::ControlMessage;
-use crate::node::{self, AudioNode};
+use crate::node::{self, AudioNode, PeriodicWave, PeriodicWaveConstraints};
 use crate::param::{AudioParam, AudioParamOptions, AutomationEvent};
 use crate::process::AudioProcessor;
 use crate::spatial::{AudioListener, AudioListenerParams};
@@ -154,6 +154,20 @@ pub trait AsBaseAudioContext {
     /// Creates a AnalyserNode
     fn create_analyser(&self) -> node::AnalyserNode {
         node::AnalyserNode::new(self.base(), Default::default())
+    }
+
+    /// Creates a periodic wave
+    //
+    // dictionary PeriodicWaveConstraints {
+    //   boolean disableNormalization = false;
+    // };
+    fn create_periodic_wave(
+        &self,
+        real: Vec<f32>,
+        imag: Vec<f32>,
+        constraints: Option<PeriodicWaveConstraints>,
+    ) -> PeriodicWave {
+        PeriodicWave::new(real, imag, constraints)
     }
 
     /// Create an AudioParam.
--- a/src/node.rs
+++ b/src/node.rs
@@ -22,7 +22,6 @@ use crate::{BufferDepletedError, SampleRate, BUFFER_SIZE};
 use crossbeam_channel::{self, Receiver, Sender};
 
 use lazy_static::lazy_static;
-use realfft::num_traits::Signed;
 
 /// This interface represents audio sources, the audio destination, and intermediate processing
 /// modules.
@@ -223,11 +222,78 @@ lazy_static! {
     };
 }
 
+/// Options for constructing a periodic wave
+pub struct PeriodicWaveConstraints {
+    /// By default PeriodicWave is build with normalization enabled (disable_normalization = false).
+    /// In this case, a peak normalization is applied to the given custom periodic waveform.
+    ///
+    /// If disable_normalization is enabled (disable_normalization = true), the normalization is
+    /// defined by the periodic waveform characteristics (img, and real fields).
+    disable_normalization: bool,
+}
+
+/// PeriodicWave is a setup struct required to build
+/// custom periodic waveform oscillator type.
+#[derive(Clone)]
+pub struct PeriodicWave {
+    /// The real parameter represents an array of cosine terms of Fourrier series.
+    ///
+    /// The first element (index 0) represents the DC-offset.
+    /// This offset has to be given but will not be taken into account
+    /// to build the custom periodic waveform.
+    ///
+    /// The following elements (index 1 and more) represent the fundamental and harmonics of the periodic waveform.
+    real: Vec<f32>,
+    /// The imag parameter represents an array of sine terms of Fourrier series.
+    ///
+    /// The first element (index 0) will not be taken into account
+    /// to build the custom periodic waveform.
+    ///
+    /// The following elements (index 1 and more) represent the fundamental and harmonics of the periodic waveform.
+    imag: Vec<f32>,
+    /// By default PeriodicWave is build with normalization enabled (disable_normalization = false).
+    /// In this case, a peak normalization is applied to the given custom periodic waveform.
+    ///
+    /// If disable_normalization is enabled (disable_normalization = true), the normalization is
+    /// defined by the periodic waveform characteristics (img, and real fields).
+    disable_normalization: bool,
+}
+
+impl PeriodicWave {
+    /// Returns a PeriodicWave
+    ///
+    /// # Arguments
+    ///
+    /// * `real` - The real parameter represents an array of cosine terms of Fourrier series.
+    /// * `imag` - The imag parameter represents an array of sine terms of Fourrier series.
+    /// * `constraints` - The constraints parameter specifies the normalization mode of the PeriodicWave
+    pub(crate) fn new(
+        real: Vec<f32>,
+        imag: Vec<f32>,
+        constraints: Option<PeriodicWaveConstraints>,
+    ) -> Self {
+        if let Some(c) = constraints {
+            Self {
+                real,
+                imag,
+                disable_normalization: c.disable_normalization,
+            }
+        } else {
+            Self {
+                real,
+                imag,
+                disable_normalization: false,
+            }
+        }
+    }
+}
+
 /// Options for constructing an OscillatorNode
 pub struct OscillatorOptions {
     pub type_: OscillatorType,
     pub frequency: f32,
     pub channel_config: ChannelConfigOptions,
+    pub periodic_wave: Option<PeriodicWave>,
 }
 
 impl Default for OscillatorOptions {
@@ -236,6 +302,7 @@ impl Default for OscillatorOptions {
             type_: OscillatorType::default(),
             frequency: 440.,
             channel_config: ChannelConfigOptions::default(),
+            periodic_wave: None,
         }
     }
 }
@@ -277,6 +344,7 @@ pub struct OscillatorNode {
     channel_config: ChannelConfig,
     frequency: AudioParam,
     type_: Arc<AtomicU32>,
+    periodic_wave: Option<PeriodicWave>,
     scheduler: Scheduler,
 }
 
@@ -304,6 +372,12 @@ impl AudioNode for OscillatorNode {
 }
 
 impl OscillatorNode {
+    /// Returns an OscillatorNode
+    ///
+    /// # Arguments:
+    ///
+    /// * `context` - The AudioContext
+    /// * `options` - The Oscillatoroptions
     pub fn new<C: AsBaseAudioContext>(context: &C, options: OscillatorOptions) -> Self {
         context.base().register(move |registration| {
             let sample_rate = context.base().sample_rate().0 as f32;
@@ -322,11 +396,15 @@ impl OscillatorNode {
             let type_ = Arc::new(AtomicU32::new(options.type_ as u32));
             let scheduler = Scheduler::new();
 
-            let sine_renderer = SineRenderer::new(440., sample_rate);
-            let sawtooth_renderer = SawRenderer::new(440., sample_rate);
-            let triangle_renderer = TriangleRenderer::new(440., sample_rate);
-            let square_renderer = SquareRenderer::new(440., sample_rate);
-            let custom_renderer = CustomRenderer::new(440., sample_rate);
+            let sine_renderer = SineRenderer::new(options.frequency, sample_rate);
+            let sawtooth_renderer = SawRenderer::new(options.frequency, sample_rate);
+            let triangle_renderer = TriangleRenderer::new(options.frequency, sample_rate);
+            let square_renderer = SquareRenderer::new(options.frequency, sample_rate);
+            let custom_renderer = CustomRenderer::new(
+                options.frequency,
+                sample_rate,
+                options.periodic_wave.clone(),
+            );
 
             let render = OscillatorRenderer {
                 frequency: f_proc,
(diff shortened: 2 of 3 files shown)


orottier:
Hi! This looks very very good.
My intend for this library is to be correct, performant and readable (in that order). Your work ticks 1 and 3, and we can use profiling to see about performance.
I'm leaving a few comments in the code but by all means please continue

src/node.rs
```diff
@@ -291,23 +425,43 @@ impl OscillatorNode {
...
+    /// set the oscillator type to custom. The oscillator will generate
+    /// a perdioc waveform following the PeriodicWave characteristics
+    //
+    //  TODO: The current implementation doesn't communicate its state
+    //  to the OscillatorRenderer, and so has no effect on the rendering.
+    //  This function should send the updated periodics waveform characteristics
+    //  to the OscillatorRenderer and more specifically to the CustomRenderer
+    pub fn set_periodic_wave(&mut self, _periodic_wave: PeriodicWave) {
+        self.set_type(OscillatorType::Custom);
+        todo!();
```
orottier, line 452: You can use a rendez-vous channel. A bit more complicated setup is used in the AnalyserNode as that requires back and forth communication.
It sets up a `crossbeam_channel::bounded(0)` channel. The user-facing node can send the PeriodicWave (and will block), the render thread will come by and use `.try_recv()` (so not blocking) to check if any new wave needs to be picked up.

src/node.rs
```diff
@@ -357,6 +538,368 @@ impl AudioProcessor for OscillatorRenderer {
...
+        sample -= self.poly_blep(norm_phase);
+        self.phase = (self.phase + self.incr_phase) % TABLE_LENGTH_F32;
+        sample
+    }
+}
+
+struct CustomRenderer {
+    frequency: f32,
+    sample_rate: f32,
+    cplxs: Vec<(f32, f32)>,
```
orottier, line 757: There may be a performance improvement by packing these multiple Vecs as a single Vec of tuples. But let's see what profiling reports

src/node.rs
```diff
@@ -176,11 +178,122 @@ pub trait AudioControllableSourceNode {
     }
 }
 
+const TABLE_LENGTH_USIZE: usize = 2048;
+const TABLE_LENGTH_F32: f32 = TABLE_LENGTH_USIZE as f32;
+
+// Compute one period sine wavetable of size TABLE_LENGTH
+lazy_static! {
```
orottier, line 185: Alternatively, in a build.rs script we can pre-compute these and later `include_bytes()` them. That saves an atomic sync and heap jump on every read. As always, profile before you optimize

src/node.rs
```diff
@@ -176,11 +178,122 @@ pub trait AudioControllableSourceNode {
     }
 }
 
+const TABLE_LENGTH_USIZE: usize = 2048;
```
orottier, line 181: Perhaps move all (private) oscillator internals in a separate mod? The node.rs file is huge already

orottier:
Maybe write a new expample file that rings all oscillator types for 1 second consecutively? Then we can use cargo-flamechart on it for profiling

src/node.rs
```diff
@@ -176,11 +178,122 @@ pub trait AudioControllableSourceNode {
...
+    /// By default PeriodicWave is build with normalization enabled (disable_normalization = false).
+    /// In this case, a peak normalization is applied to the given custom periodic waveform.
+    ///
+    /// If disable_normalization is enabled (disable_normalization = true), the normalization is
+    /// defined by the periodic waveform characteristics (img, and real fields).
+    disable_normalization: bool,
+}
+
+/// PeriodicWave is a setup struct required to build
+/// custom periodic waveform oscillator type.
```
orottier, line 236: Maybe add doc how to construct it?

src/node.rs
Jerboas86: I put all related oscillator code into oscillator.rs

src/node.rs
Jerboas86: - [x] Done

Jerboas86:
> Maybe write a new example file that rings all oscillator types for 1 second consecutively? Then we can use cargo-flamechart on it for profiling

- [x] Done

Jerboas86:
I profiled heap allocations with heaptrack and found that the bottleneck is AudioParamProcessor::tick method which seems to allocate at each call ( at least for A-rate parameter).

```rust
impl AudioParamProcessor {
    pub fn value(&self) -> f32 {
        if self.value.is_nan() {
            self.default_value
        } else {
            self.value.clamp(self.min_value, self.max_value)
        }
    }

    fn tick(&mut self, ts: f64, dt: f64, count: usize) -> Vec<f32> {
        // store incoming automation events in sorted queue
        for event in self.receiver.try_iter() {
            self.events.push(event);
        }

        // setup return value buffer
        let a_rate = self.automation_rate == AutomationRate::A;
        let mut result = if a_rate {
            // empty buffer
           ////////// I think this is were the allocation happens //////////////////////////
            Vec::with_capacity(count)
        } else {
            // filling the vec already, no expensive calculations are performed later
            vec![self.value(); count]
        };

```
I am learning the profiling tools in the process, so what i am saying should be double check to be sure.

orottier:
I've created a separate issue #16 for the audioparam allocations. Definitely something to look into

Jerboas86:
The polyBLEP algorithm is not working correctly. I'm investigating.

src/node/oscillator.rs
```diff
@@ -0,0 +1,1006 @@
...
+        }
+
+        for (r, incr_ph) in self.interpol_ratios.iter_mut().zip(self.incr_phases.iter()) {
+            *r = incr_ph - incr_ph.floor();
+        }
+
+        self.computed_freq = new_comp_freq;
+    }
+
+    fn set_periodic_wave(&mut self, periodic_wave: PeriodicWave) {
```
orottier, line 872: Ideally, these computations (and allocations) could be done on the control thread.
You could extend the `OscMsg` to include the pre-calculated fields

src/node/oscillator.rs
```diff
@@ -0,0 +1,1006 @@
...
+            1 => Square,
+            2 => Sawtooth,
+            3 => Triangle,
+            4 => Custom,
+            _ => unreachable!(),
+        }
+    }
+}
+
+enum OscMsg {
```
orottier, line 230: What's the motivation for a single discriminant enum?

orottier:
Progress is looking very good 👍

Jerboas86:
- PolyBLEP implementation should be now correct (but i need to make futher measurements to be sure)
- Remove wavetable synthesis for square, triangle, and sawtooth signal for performance reason and because it doesn't work with polyBLEP
- Each OscillatorType signal is now generated by a unique renderer called `OscRendererInner`
- A bunch of optimizations
- Supposedly fix #9 (i need to check  futher to be sure)

To be continued...

src/node/oscillator.rs
Jerboas86: In case i need to add other type of messages.

I had planned to make parameters computations in the control thread and pass them to render thread. like you suggest it in another comment

> Ideally, these computations (and allocations) could be done on the control thread.
> You could extend the OscMsg to include the pre-calculated fields

and i see reading it that you thought about it too. We are on the same page.

Jerboas86:
Why mix polyBLEP and wavetable was not a good idea ?
1- Wavetable didn't add performance over computed synthesis.
2 - Linear interpolation and polyBLEP don't work quite well together. Another type of interpolation like truncation could have work better. But i didn't test it since i had better perfs with computed synthesis.

src/param.rs
```diff
@@ -289,10 +289,13 @@ impl AudioParamProcessor {
                         let end_index_clipped = end_index.min(count);
                         let n_values = end_index_clipped - start_index;
 
+                        let mut val = self.value;
```
orottier, line 292: I have cherry picked this commit, made a small change and added a regression test in #19

Jerboas86:
I think next optimization should be to pass some computation on the control thread side. Let me know if you have other suggestions.

I let lazy_static as a way to initialize wavetable as it is easy to use and doesn't appear to be a bottleneck.

Jerboas86:
* I would like to build a "reactive" audioparam to computedOscFreq but i have no idea how to do that. Any idea to where i could start and if it is a good idea in the first place ?
* I tried to add some tests. From what i understand there is no CPAL backend when testing, would you agree to limit the no backend CPAL config to only CI and give the possibility to write test with a CPAL renderthread when running on local machine ?

src/node/oscillator.rs
```diff
@@ -272,13 +272,15 @@ impl OscillatorNode {
     ///
     /// * `context` - The AudioContext
     /// * `options` - The Oscillatoroptions
-    pub fn new<C: AsBaseAudioContext>(context: &C, options: OscillatorOptions) -> Self {
+    pub fn new<C: AsBaseAudioContext>(context: &C, options: Option<OscillatorOptions>) -> Self {
```
orottier, line 275: nitpick: Ideally we only take one level of `Option<>` so maybe make this argument mandatory en let the members still be optional

examples/oscillators.rs
```diff
@@ -0,0 +1,83 @@
...
+    // Create a custom periodic wave
+    let periodic_wave = context.create_periodic_wave(options);
+
+    // Oscillator needs to be started explicitily
+    osc.start();
+
+    println!("Sine tone sweep playing...");
+
+    // Sine tone sweep
+    for i in 0..1000 {
```
orottier, line 33: Now linear ramp to value is fixed fbis could be written more idiomatic

orottier:
Except for the merge conflict this is ready to merge? Or was there anything else you wanted to add?

src/context.rs
```diff
@@ -359,7 +364,16 @@ impl Drop for AudioContextRegistration {
 
         if !magic {
             let message = ControlMessage::FreeWhenFinished { id: self.id.0 };
-            self.context.inner.render_channel.send(message).unwrap();
+            let send_result = self.context.inner.render_channel.send(message);
+
+            if cfg!(test) {
```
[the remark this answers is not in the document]
Jerboas86: Thanks, i will do that

src/node/oscillator.rs
Jerboas86: I think we should keep the option optional cause:
- compliance reason: this interface follows the web audio api interface
- Ergonomic reason: `options: None `vs `options: Options { field1: None, field2: None}`

The only drawback i see is that we have to deal with the option during construction.

Jerboas86:
Performance report

% represents stack trace frequency
- New: generate_{sine,sawtooth,triangle,square,custom}
- Old: AudioProcessor::process

|          | New    | Old    |
|----------|--------|--------|
| Sine     | 12.11% | NA     |
| Sawtooth | 8.92%  | 15.19% |
| Triangle | 5.68%  | NA     |
| Square   | 5.26%  | 4.65%  |
| Custom   | 12.11% | NA     |

The performance bottleneck is by far Audioparam processing  x4 worse perfs compared to AudioNode processing. So i think we should focus on that next.

Program used to make measurements

```rust
//! This example plays each oscillator type sequentially
use web_audio_api::context::{AsBaseAudioContext, AudioContext};
use web_audio_api::node::{AudioNode, AudioScheduledSourceNode, OscillatorType};

fn main() {
    // Create an audio context where all audio nodes lives
    let context = AudioContext::new();

    // Create an oscillator node with sine (default) type
    let mut osc = context.create_oscillator();

    // Connect osc to the destination node which is the default output device
    osc.connect(&context.destination());

    // Oscillator needs to be started explicitily
    osc.start();
    osc.set_type(OscillatorType::Square);
    // Select Sawtooth as the PeriodicWave type

    //     // defined periodic wave characteristics
    //     let options = Some(PeriodicWaveOptions {
    //         real: Some(vec![0., 0.5, 0.5]),
    //         imag: Some(vec![0., 0., 0.]),
    //         disable_normalization: Some(true),
    //     });

    //     // Create a custom periodic wave
    //     let periodic_wave = context.create_periodic_wave(options);

    //     osc.set_periodic_wave(periodic_wave);
    //     osc.set_type(OscillatorType::Custom);

    std::thread::sleep(std::time::Duration::from_secs(30));
    // enjoy listening
}
```

orottier:
Two unused fields:
```
warning: field is never read: `norm_factor`
   --> src/node/oscillator.rs:523:5
    |
523 |     norm_factor: Option<f32>,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default

warning: field is never read: `disable_normalization`
   --> src/node/oscillator.rs:524:5
    |
524 |     disable_normalization: bool,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: 2 warnings emitted```

What's your plan for integration tests?

File: src/context.rs (hunk 11 above)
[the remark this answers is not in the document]
Comment (orottier) [reply src/context.rs:369]: This should still be reverted

File: src/node/oscillator.rs (hunk 9 above)
Comment (orottier) [reply src/node/oscillator.rs:275]: It's fine for now, later we can have a look at alle node options and figure out the best way.
For ergonomics you could use `options: Default::default()`

File: src/node/oscillator.rs [hunk 12]
```diff
@@ -0,0 +1,1238 @@
...
+impl OscillatorNode {
+    /// Returns an `OscillatorNode`
+    ///
+    /// # Arguments:
+    ///
+    /// * `context` - The `AudioContext`
+    /// * `options` - The Oscillatoroptions
+    pub fn new<C: AsBaseAudioContext>(context: &C, options: Option<OscillatorOptions>) -> Self {
+        context.base().register(move |registration| {
+            let sample_rate = context.base().sample_rate().0 as f32;
```
Jerboas86, line 318: Would it be possible to replace the SampleRate(u32) by an f32 ?
It would:
- avoid castings (with possible loss of precision)
- be compliant with the [specs](https://www.w3.org/TR/webaudio/#dom-baseaudiocontext-samplerate)

Issue #16, opened 2021-09-24
Prevent allocations on every quantum in AudioParamProcessor


orottier:
I profiled heap allocations with heaptrack and found that the bottleneck is AudioParamProcessor::tick method which seems to allocate at each call ( at least for A-rate parameter).

```rust
impl AudioParamProcessor {
    pub fn value(&self) -> f32 {
        if self.value.is_nan() {
            self.default_value
        } else {
            self.value.clamp(self.min_value, self.max_value)
        }
    }

    fn tick(&mut self, ts: f64, dt: f64, count: usize) -> Vec<f32> {
        // store incoming automation events in sorted queue
        for event in self.receiver.try_iter() {
            self.events.push(event);
        }

        // setup return value buffer
        let a_rate = self.automation_rate == AutomationRate::A;
        let mut result = if a_rate {
            // empty buffer
           ////////// I think this is were the allocation happens //////////////////////////
            Vec::with_capacity(count)
        } else {
            // filling the vec already, no expensive calculations are performed later
            vec![self.value(); count]
        };

```
I am learning the profiling tools in the process, so what i am saying should be double check to be sure.

_Originally posted by @Jerboas86 in https://github.com/orottier/web-audio-api-rs/issues/15#issuecomment-926664460_

orottier:
I think `AudioParamProcessor` can have a `values: Vec<f32>` field that can be re-used in every `tick` call. No need to create Vecs as they are only copied as slice when processing

orottier:
This was fixed in #17

PR #17, opened 2021-09-27
Avoid allocation during AudioParamProcessor::tick calls

Jerboas86:
Remove `Vec` allocations at each call to `tick`.

Instead the `Vec` is created at instantion of `AudioParamProcessor` an reused at each `tick` call

With this optimization, memory allocation is greatly reduced 1MB (fixed) instead of 80MB (linearly growing with execution time).

Allocation measurements have been made with Heaptrack.

#16


orottier:
Very good, thanks!

Issue #8, opened 2021-09-03
Context::current_time is not updated correctly


Jerboas86:
Context::current_time always returns 0

Runs the following example to reproduce:

```rust
use web_audio_api::context::{AsBaseAudioContext, AudioContext};
use web_audio_api::node::{AudioNode, AudioScheduledSourceNode};

fn main() {
    let context = AudioContext::new();

    let osc1 = context.create_oscillator();
    osc1.connect(&context.destination());
    osc1.frequency().set_value_at_time(440., 0.);
    osc1.start();

    loop {
        println!("Time: {}", context.current_time());
    }
}
```

The example keeps printing 0:
```
Time: 0
Time: 0
Time: 0
Time: 0
```

from [web audio api spec](https://www.w3.org/TR/webaudio/#:~:text=Thus%2C%20for%20a%20running%20context%2C%20currentTime%20increases%20steadily%20as%20the%20system%20processes%20audio%20blocks%2C%20and%20always%20represents%20the%20time%20of%20the%20start%20of%20the%20next%20audio%20block%20to%20be%20processed.):

> Thus, for a running context, currentTime increases steadily as the system processes audio blocks, and always represents the time of the start of the next audio block to be processed.

Jerboas86:
`AudioContext::current_time()` is derived from `BaseAudioContextInner.frames_played` field. But this field is not incremented at each render quantum.

```
struct BaseAudioContextInner {
    /// sample rate in Hertz
    sample_rate: SampleRate,
    /// number of speaker output channels
    channels: u32,
    /// incrementing id to assign to audio nodes
    node_id_inc: AtomicU64,
    /// message channel from control to render thread
    render_channel: Sender<ControlMessage>,
    /// number of frames played
    frames_played: AtomicU64,
    /// AudioListener fields
    listener_params: Option<AudioListenerParams>,
}
```

On the other hand, `RenderThread` struct got a frames_played field which is updated on each render quantum. But i don't think that this value can be accessed by `AudioContext`

```rust
pub(crate) struct RenderThread {
    graph: Graph,
    sample_rate: SampleRate,
    channels: usize,
    frames_played: AtomicU64,
    receiver: Receiver<ControlMessage>,
}
```

orottier:
Yes, you are right. The `frames_played` should be wrapped in an `Arc` and should be identical to the `RenderThread`'s `frames_played`.

Jerboas86:
I will try to make a PR tomorrow ;)

[...]

PR #18, opened 2021-09-27
current_time shared between BaseAudioContext and Renderthread

Jerboas86:
frames_played is created by `RenderThread` and pass to `BaseAudioContext` during its instantiation.

frames_played has the following lock-free thread shared data structure: `Arc<AtomicU64>`

The `build_output` declaration looks like that now. it 's not ideal :)

`fn build_output() -> (Stream, StreamConfig, Arc<AtomicU64>, Sender<ControlMessage>)`

#8

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -51,7 +51,7 @@ struct BaseAudioContextInner {
     /// message channel from control to render thread
     render_channel: Sender<ControlMessage>,
     /// number of frames played
-    frames_played: AtomicU64,
+    frames_played: Arc<AtomicU64>,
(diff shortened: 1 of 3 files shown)


orottier:
Indeed the code organization could be improved between `io`, `graph` and `context` modules.
Thanks for the work, I'm merging it and adding a few small refactors

orottier:
Added https://github.com/orottier/web-audio-api-rs/commit/2d86f3d863b5b38c999a15f5f837745a88fc0340 to main to make it a bit less clumsy

Issue #12, opened 2021-09-14
Add BiquadFilterNode


Jerboas86:
Add the `BiquadFilterNode` to `web-audio-api-rs` to be compliant with [(web audio api spec)](https://www.w3.org/TR/webaudio/#:~:text=BiquadFilterNode%20is%20an%20AudioNode%20processor%20implementing%20very%20common%20low-order%20filters)

**Solution:**
I have not yet written a prototype, but i think i can reach at least partial compliance for this node

May i begin to work on it ?

orottier:
Sure, go ahead! The [biquad crate](https://crates.io/crates/biquad) seems very useful but maybe the algorithm is simple enough to build it yourself

PR #27, opened 2021-10-05
BiquadFilterNode

Jerboas86:
Implementation of the BiquadFilterNode.

### Done
- Api (supposedly)  Compliant with the web audio api specs
- Add an example biquad.rs
- Add docs
- Implemention details: transposed Direct Form II (best implementation for float arithmetic)

### Todos
- [x] A-rate controls
- [x] Optimizations
- [ ] Add tests

### Optional
- Remove complex-num dep

related to #12


src/node/biquad_filter.rs
```diff
@@ -0,0 +1,1180 @@
...
+    #[allow(clippy::cast_possible_truncation)]
+    pub fn get_frequency_response(
+        &self,
+        frequency_hz: &mut [f32],
+        mag_response: &mut [f32],
+        phase_response: &mut [f32],
+    ) {
+        self.validate_inputs(frequency_hz, mag_response, phase_response);
+        let (sender, receiver) = crossbeam_channel::bounded(0);
+        self.sender
```
Jerboas86, line 294: I am not able to send message with OfflineAudioContext. It is the correct behavior or it is a bug ?

src/node/biquad_filter.rs
```diff
@@ -0,0 +1,1180 @@
...
+    /// Mock of `get_frequency_response`
+    /// This function is the same as `get_frequency_response` except it never send the `CoeffsReq`.
+    /// In tests, we use `OfflineAudioContext` and in this context the `CoeffsReq` is not sendable.
+    ///
+    /// # Arguments
+    ///
+    /// * `frequency_hz` - frequencies for which frequency response of the filter should be calculated
+    /// * `mag_response` - magnitude of the frequency response of the filter
+    /// * `phase_response` - phase of the frequency response of the filter
+    #[cfg(test)]
```
Jerboas86, line 367: `#[cfg(not(test))]` is not well supported by rust analyzer. This is why i made a mock function to use `#[cfg(test)]` instead

src/node/biquad_filter.rs
orottier: It's tricky. The OfflineAudioContext is inactive until you call `start_rendering`. So putting any messages in the rendezvouz channel will hang the main thread since there is no rendering thread picking up the input.

An option would be to let `start_rendering` start a separate thread, and return a `JoinHandle<AudioBuffer>`. This is also more spec compliant. Please do not do this in this PR though, we can make a separate issue

src/node/biquad_filter.rs
```diff
@@ -0,0 +1,1180 @@
...
+        mag_response: &mut [f32],
+        phase_response: &mut [f32],
+    ) {
+        self.validate_inputs(frequency_hz, mag_response, phase_response);
+        let (sender, receiver) = crossbeam_channel::bounded(0);
+        self.sender
+            .send(CoeffsReq(sender))
+            .expect("Sending CoeffsReq failed");
+
+        loop {
```
orottier, line 298: I believe this loop will cause a 100% CPU usage spike. You could use try_recv_timeout and give up after the timeout

src/node/biquad_filter.rs
```diff
@@ -0,0 +1,1180 @@
...
+    ) {
+        assert_eq!(
+            frequency_hz.len(),
+            mag_response.len(),
+            " InvalidAccessError: All paramaters should be the same length"
+        );
+        assert_eq!(
+            mag_response.len(),
+            phase_response.len(),
+            " InvalidAccessError: All paramaters should be the same length"
```
orottier, line 347: Spelling: parameter

src/node/biquad_filter.rs
```diff
@@ -0,0 +1,1180 @@
...
+        let g_values = params.get(&self.gain);
+        let det_values = params.get(&self.detune);
+        let freq_values = params.get(&self.frequency);
+        let q_values = params.get(&self.q);
+
+        self.filter(input, output, g_values, det_values, freq_values, q_values);
+    }
+
+    fn tail_time(&self) -> bool {
+        false
```
orottier, line 446: Todo: true! Spec has some recommendations

src/node/biquad_filter.rs
```diff
@@ -0,0 +1,1180 @@
...
+                Err(crossbeam_channel::TryRecvError::Disconnected) => {
+                    println!("Receiver Error: disconnected type");
+                    continue;
+                }
+                Err(crossbeam_channel::TryRecvError::Empty) => {
+                    println!("Receiver Error: empty type");
+                    continue;
+                }
+                Ok([b0, b1, b2, a1, a2]) => {
+                    println!("received...");
```
orottier, line 309: this debugging statement can be dropped

src/node/biquad_filter.rs
```diff
@@ -0,0 +1,1271 @@
...
+        mag_response: &mut [f32],
+        phase_response: &mut [f32],
+    ) {
+        self.validate_inputs(frequency_hz, mag_response, phase_response);
+        let (sender, receiver) = crossbeam_channel::bounded(0);
+        self.sender
+            .send(CoeffsReq(sender))
+            .expect("Sending CoeffsReq failed");
+
+        match receiver.recv_timeout(Duration::from_millis(10)) {
```
Jerboas86, line 331: Why 10 ms ? because less than 10 ms is too short and more than 10 ms is too long.
There is certainly a better value to put here.
 Just for my understandings, is it Ok to block on the control thread ?

src/node/biquad_filter.rs
Jerboas86: Done :)

src/node/biquad_filter.rs
```diff
@@ -0,0 +1,1180 @@
...
+            6 => Notch,
+            7 => Allpass,
+            _ => unreachable!(),
+        }
+    }
+}
+
+// the naming comes from the web audio specfication
+#[allow(clippy::module_name_repetitions)]
+pub struct BiquadFilterOptions {
```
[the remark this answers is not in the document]
Jerboas86: Done :)

src/node/biquad_filter.rs
orottier: That's fine for now, but it has a performance penalty.
Let's make a separate issue to handle tail time for the biquad and iirfilternode

orottier: issue #34

src/node/biquad_filter.rs
orottier: That's a very valid question.
This implementation is fine for now and in line with other functions that need to fetch information from the render thread to the control thread. But if it is the best solution? I don't know.. My goal was to prevent allocations and needless copies in the render thread. But this does block the control thread for max one render quantum. This could be an issue if you are building a visualization with extremely high framerate.

Jerboas86:
I will make another PR to add tests to this node

Issue #23, opened 2021-09-30
Add IIRFilterNode


orottier:
https://webaudio.github.io/web-audio-api/#IIRFilterNode

Jerboas86:
I have the algorithm implementation ready for this one. i need to wrap it into a node structure an i ll send a PR

orottier:
fixed in #31

PR #31, opened 2021-10-07
IirFilterNode implementation

Jerboas86:
Implementation of the IirFilterNode.

### Done
- Api (supposedly)  Compliant with the web audio api specs
- Add an example iir.rs
- Add docs

### Todos
- [x] Optimizations
- [x] Add tests

related to #23


src/node/iir_filter.rs
```diff
@@ -0,0 +1,271 @@
...
+    /// # Arguments
+    ///
+    /// * `input` - Audiobuffer input
+    #[inline]
+    fn tick(&mut self, input: f32) -> f32 {
+        let input = input as f64;
+        let output = self.norm_coeffs[0].0 * input + self.states[0];
+
+        for (idx, (ff, fb)) in self.norm_coeffs.iter().skip(1).enumerate() {
+            self.states[idx] = ff * input - fb * output + self.states.get(idx + 1).unwrap_or(&0.);
```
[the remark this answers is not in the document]
Jerboas86: I think i found another solution. 
`states` length is now the same as `norm_coeffs` length.
this extra state is only read and is always 0

Jerboas86:
i will replace white noise with sine signals that should make it work.

src/node/iir_filter.rs
```diff
@@ -0,0 +1,652 @@
...
+        background.connect(&iir);
+        iir.connect(&context.destination());
+
+        background.start();
+        let output = context.start_rendering();
+
+        // retrieve processed data by removing silence chunk
+        // These silence slices are inserted inconsistently from an test execution to another
+        // Without these processing the test would be brittle
+        let data_ch: Vec<f32> = output
```
Jerboas86, line 638: I retry the test a bunch of time with a constant success.

orottier: I think the inconsistencies come from the fact that you have used a MediaElement, which will perform some buffering that can be non-deterministic in tests. Can you try to use `create_media_stream_source` on the stream? That fixes the problem I hope and gives a bit more confidence in the test

Jerboas86: i will try your proposed alternative tomorrow

Jerboas86: > Can you try to use create_media_stream_source on the stream?

I tried with `create_media_stream_source` and it also fails from time to time if i don't add the counter measures of the actual test code

orottier: That just sounds really strange to me.... But probably unrelated to your work. Let's get it merged

Issue #35, opened 2021-10-21
Add AudioContextOptions to AudioContext constructor


Jerboas86:
In the specs, we can build `audioContext` with an optional argument [`AudioContextOptions`](https://www.w3.org/TR/webaudio/#dictdef-audiocontextoptions)

```js
dictionary AudioContextOptions {
  (AudioContextLatencyCategory or double) latencyHint = "interactive";
  float sampleRate;
};
```

I think we should extend this options like that:

```js
dictionary AudioContextOptions {
  (AudioContextLatencyCategory or double) latencyHint = "interactive";
  float sampleRate;
  unsigned long channels = 2;
};
```
This way, the user can specify the number of output channels that it would like. The current behavior is that the user has only access to default configuration given by CPAL.

Should `AudioContext::new(options: Option<AudioContextOptions>)` be fallible ?

My opinion is that it should not. The method should fallback to CPAL default config if user requested config fails.
And provides supported config as an error message.

related to #28 
related to #32

orottier:
Closed with merging #36

Issue #28, opened 2021-10-05
Sample rate high on ALSA


Jerboas86:
context.sample_rate() displays a sample rate of 384 000 Hz on my ALSA system.
This default sample rate is coming from the CPAL default output config.

Possible solution: try to build a stream with a lower sample rate before defaulting into the CPAL default sample rate.

orottier:
This is because we call [with_max_sample_rate](https://docs.rs/cpal/latest/cpal/struct.SupportedStreamConfigRange.html#method.with_max_sample_rate) on the SupportedStreamConfigRange in `io.rs`. We could try to use [with_sample_rate](https://docs.rs/cpal/latest/cpal/struct.SupportedStreamConfigRange.html#method.with_sample_rate) instead

PR #36, opened 2021-10-26
add AudioContextOptions to AudioContext

Jerboas86:
The user can now specify:
- the number of output channels  
- the sample rate
- the output latency
when building its `AudioContext`.

If the intented options are not met by the audio hardware capabilities, the `AudioContext` is built with `default_config` options, and a warning is printed to the user.

If the user options are missing, the `AudioContext` is built with `primary_config` options. If the `primary_config` options are not met by the audio harware capabilities, the `AudioContext` is built with `default_config` options, and a warning is printed to the user.

This PR modify `DestinationNode`:
- Adds `max_channels_count` method
- Assert `ChannelCountConstraints` on `DestinationNode`

This PR adds merger.rs example to demonstrate new capabilities.

Clippy passes on `context.rs` and `io.rs`

related to #35
related to #28


src/io.rs
```diff
@@ -89,27 +117,83 @@ pub(crate) fn build_output(
...
+    primary_config.buffer_size = cpal::BufferSize::Fixed(buffer_size);
+
+    let mut config: StreamConfig = supported_config.clone().into();
+
+    match options {
+        Some(opts) => {
+            config.sample_rate = CpalSampleRate(opts.sample_rate.unwrap_or(config.sample_rate.0));
+            config.channels = opts.channels.unwrap_or(config.channels);
+
+            match opts.latency_hint {
```
orottier, line 130: Can you move this block into a helper function, makes it a bit more readable and then we can re-use if for the InputBuilder

src/io.rs
```diff
@@ -134,7 +219,8 @@ pub(crate) fn build_output(
     (stream, config, sender)
 }
 
-pub(crate) fn build_input() -> (Stream, StreamConfig, Receiver<AudioBuffer>) {
+/// Builds the input
+pub fn build_input() -> (Stream, StreamConfig, Receiver<AudioBuffer>) {
```
orottier, line 223: You should port the latency hint to the input builder too (can be done in another PR if it's too much work). They should be symmetrical in the end

src/context.rs
```diff
@@ -488,11 +590,15 @@ impl BaseAudioContext {
             outputs: node.number_of_outputs() as usize,
             channel_config: node.channel_config_cloned(),
         };
-        self.inner.render_channel.send(message).unwrap();
+        self.inner
+            .render_channel
+            .send_timeout(message, Duration::from_millis(10))
```
orottier, line 595: Why opt for timeout? If the requested latency `> 10millis`, this will fail for no good reason

src/context.rs
```diff
@@ -575,21 +691,27 @@ impl OfflineAudioContext {
...
             length,
-            render,
+            renderer,
         }
     }
 
+    /// `OfflineAudioContext` doesn't start rendering automatically
+    /// You need to call this function to start the audio rendering
     pub fn start_rendering(&mut self) -> AudioBuffer {
         // make buffer_size always a multiple of BUFFER_SIZE, so we can still render piecewise with
```
orottier, line 701: I think actually this is no longer needed, now we can handle varying buffer lengths.
We can remove this code in another PR

src/context.rs
```diff
@@ -575,21 +691,27 @@ impl OfflineAudioContext {
...
-        let _split = buf.split_off(self.length as u32);
+        let mut buf = self.renderer.render_audiobuffer(buffer_size);
+        let _split = buf.split_off(self.length);
         buf
     }
 
-    pub fn length(&self) -> usize {
+    /// get the length of rendering audio buffer
+    #[must_use]
+    pub const fn length(&self) -> usize {
```
orottier, line 714: no const please

orottier (changes requested):
Interesting work, thanks!

src/io.rs
```diff
@@ -57,84 +85,356 @@ fn spawn_input_stream(
...
-        receiver,
-        frames_played_clone,
-    );
+    /// modify the buffer size config, following user options
+    ///
+    /// # Argument
+    ///
+    /// * `options` - options contains latency hint information
+    ///
+    /// # Warning
```
Jerboas86, line 209: As mentioned here the correctness of the API is not enforce by the type system. I could make it more strict but at the cost of more complexity. What do you think ?

Jerboas86 (commented):
> You should port the latency hint to the input builder too (can be done in another PR if it's too much work)

I didn't port the refactoring to the input stream. leave it as "good first issue" or for another PR, provide that you agree on the refactoring :)

Jerboas86:
> I didn't port the refactoring to the input stream. i leave it as "good first issue" or for another PR
 
is it Ok with you, if i open a good first issue about this topic?

orottier:
Go ahead!

PR #33, opened 2021-10-20
Refact/01

Jerboas86:
I've done some refactoring:
- Each audio node has its own file
- Modify the interface of `AudioBuffer` to better fit the specs
- Add helper traits on `AudioBuffer` and `ChannelData`
- Replace `SampleRate(u32)` with `f32`
- float_eq crate is now in [dependencies] section
-  Modularize `Graph::process` and `Graph::render` 

related to #30

PR #41, opened 2021-10-28
One node per file

Jerboas86:
## Project structure

Each node has its own file

## Notes 

last commits are only renaming files to make them shorter

related to PR #33


orottier:
Sorry I caused another conflict. I intend to merge though so this will soon be over :)

Issue #45, opened 2021-11-04
Possible memory leak?


b-ma:
Hi,

First, thanks for your work on this promising project. 

I was playing with your library to create some Node.js bindings  (using [https://napi.rs/](https://napi.rs/)) and after some minimal work to wrap your code, I wrote this small javascript test. However the sound rapidly starts to be very noisy (a kind of nice noise but not the kind we would expect...), with the script rapidly growing to take almost 100% of a processor:

```js
function triggerSine(audioContext) {
  const sine = audioContext.createOscillator();
  sine.connect(audioContext.destination);

  const now = audioContext.currentTime;
  sine.start(now);
  sine.stop(now + 0.03);
}

(async function() {
  const audioContext = new AudioContext();
  await audioContext.resume();
  setInterval(() => triggerSine(audioContext), 0.05 * 1000);
}());
```

While I agree this may appear as a kind of "agressive" test, this kind of pattern is quite common e.g. for granular synthesis or for generative music.

To make sure, the problem does not come from my own code, I re-wrote the test in pure Rust but ran into the exact same issue:

```rust
use web_audio_api::context::{AsBaseAudioContext, AudioContext};
use web_audio_api::node::{
  AudioNode, AudioScheduledSourceNode, OscillatorType, PeriodicWaveOptions,
};
use std::{thread, time};

fn trigger_sine(audio_context: &AudioContext) {
  let osc = audio_context.create_oscillator();
  osc.connect(&audio_context.destination());

  let now = audio_context.current_time();
  osc.start_at(now);
  osc.stop_at(now + 0.03)
}

fn main() {
  let audio_context = AudioContext::new();

  // mimic setInterval
  loop {
    trigger_sine(&audio_context);
    thread::sleep(time::Duration::from_millis(50));
  }
}
```

As I'm a newbie with Rust, I didn't try to dig more deeply into your code, but I would be happy to help the best I can.

Cheers,
Benjamin

orottier:
Hi @b-ma, thanks for the detailed report. It is very helpful.

I believe your assumptions are correct and your rust code should not cause a memory leak and/or performance issue.
My guess is that we currently do not clean up oscillator nodes from the audio graph as they are stopped. Therefore the audio graph will continue growing and use up memory and CPU.

[...]

PR #46, opened 2021-11-05
Debugging issue #45: failing to clean up finished nodes

orottier:
Work in progress. Even with the OscillatorNode's fixed tail_time, it still seems if not all nodes are properly cleaned up.. I will have to look into it later


Jerboas86:
I think that you could change the `process` function declaration and make it return a boolean. `process` returns `true` only if output is generated. if no output is generated you can drop the node. This is just an idea i didn't try to implement it.

orottier:
Great minds think alike. Your proposal is exactly the spec's suggestion: https://www.w3.org/TR/webaudio/#AudioWorkletProcessor
That should make my implementation less clunky, for sure.
But there must still be some underlying bug. I will try to have a look this weekend.

orottier:
I've found the bug causing the audio graph not cleaning up properly.
The OscillatorNodes are removed from the graph correctly when they are finished playing, but the audio params that feed into the node are not removed.
In `graph.rs`, nodes are cleaned up if they no longer have outputs connected, but nowhere in the code `has_outputs_connected: bool` is actually set to `false` so this never happens

Issue #43, opened 2021-11-02
Implement all up/down-mixing combinations


orottier:
https://www.w3.org/TR/webaudio/#UpMix-sub
https://www.w3.org/TR/webaudio/#down-mix

A few common mixing implementations are already in src/node.rs
`pub fn mix(&mut self, channels: usize, interpretation: ChannelInterpretation)`

PR #47, opened 2021-11-09
implemented up/down mixing combinations (cf. #43)

b-ma:
I hope this is clean enough regarding performance and Rust conventions. Let me know if it's ok for you.

_Aside_: I wondered if renaming the argument `channels` to `computed_number_of_channels` or something, wouldn't be more clear regarding the spec.

### DIFF
--- a/src/alloc.rs
+++ b/src/alloc.rs
@@ -214,10 +214,10 @@ impl AudioBuffer {
     }
 
     /// Up/Down-mix to the desired number of channels
-    pub fn mix(&mut self, channels: usize, interpretation: ChannelInterpretation) {
-        assert!(channels < MAX_CHANNELS);
+    pub fn mix(&mut self, computed_number_of_channels: usize, interpretation: ChannelInterpretation) {
+        assert!(computed_number_of_channels < MAX_CHANNELS);
 
-        if self.number_of_channels() == channels {
+        if self.number_of_channels() == computed_number_of_channels {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


orottier:
Hi @b-ma , this looks very good!
Feel free to adjust variable names to correspond to the spec more.
Also, could you run `cargo clippy` and `cargo format` and implement the suggested changes for `src/alloc.rs` (depending on your rustc version, warnings may appear for other files too)
After that we are ready to merge!

Jerboas86:
These are the clippy flags that i use at the top of my modules to make clippy especially picky. Fell free to use it.

```rust
// for example at the top of alloc.rs
#![warn(
    clippy::all,
    clippy::pedantic,
    clippy::nursery,
    clippy::perf,
    clippy::missing_docs_in_private_items
)]
```

be careful to false positive though

And thank you for your work.

b-ma:
Hey, thanks for tips. I actually only used the `all` and `perf` clippy flags because the other ones gave me hints I'm really not sure I understand enough.

src/alloc.rs
```diff
@@ -1,12 +1,20 @@
...
 
 use arrayvec::ArrayVec;
 use std::cell::RefCell;
 use std::rc::Rc;
 
 use crate::buffer::ChannelInterpretation;
 
-const LEN: usize = crate::BUFFER_SIZE as usize;
+/// size of a buffer, 128 samples per spec
+const BUFFER_SIZE: usize = crate::BUFFER_SIZE as usize;
```
b-ma, line 17: maybe this is weird

orottier, line 17: Not sure how I feel about 'shadowing' this const, but maybe all in all BUFFER_SIZE should just be `usize` always. I'll make a separate issue

Issue #20, opened 2021-09-30
Support all AudioParam automation methods


orottier:
`param.rs`
Relevant section in spec: https://webaudio.github.io/web-audio-api/#AudioParam

#### Implement automation methods

- [x] SetValueAtTime
- [x] LinearRampToValueAtTime
- [x] ExponentialRampToValueAtTime
- [x] SetTargetAtTime
- [x] SetValueCurveAtTime
- [x] CancelScheduledValues
- [x] CancelAndHoldAtTime

#### Improvements / Spec

- [x] **spec** - All `panic` calls should be replaced by clean and compliant `Errors`
- [x] **spec** - Automation methods should return `self` or `Result(self, Error)` to allow method chaining
- [x] **spec** - Sanity check all method arguments (e.g. `times`, `time_constant`)
- [ ] ~~**refactoring** - Review all `AudioNode::new` to remove `set_value` or `set_value_at_time`, default values should be properly initialised at instantiation without calling an automation method explicitly~~ _(this is maybe a bad idea as it would create a confusion between the param default value and user defined default value)_
- [ ] **perf** - `set_target_at_time` can run forever as it as no "real" end time or value, a possible performance improvement would be to internally cancel the event when `intrisic_value` is close enough to the `target` (note that Chrome implements such strategy)

#### To be discussed further

- Maintain `AutomationEventQueue` in control thread instead of audio thread (e.g. `Arc<Mutex<AutomationEventQueue>>`?) - _is maybe necessary for proper Error handling_ (cf. discussion in [https://github.com/orottier/web-audio-api-rs/pull/62#discussion_r764294272](https://github.com/orottier/web-audio-api-rs/pull/62#discussion_r764294272)
- Possibly review `set_value_curve_at_time` signature to use `Vec<32>` instead of `&[f32]` (cf. discussion in [https://github.com/orottier/web-audio-api-rs/pull/62#discussion_r764292652](https://github.com/orottier/web-audio-api-rs/pull/62#discussion_r764292652))

b-ma:
Hey, just to let you know that I've started to work on this one, if ok for you let me know if prefer a PR per item

orottier:
Sure, go ahead! Let me warn you in advance the current setup may not be ideal for SetTargetAtTime so feel free to change the whole architecture. Also, I think only A-rate is properly tested with the existing implementation so some bugs may appear

b-ma:
[...]

PR #56, opened 2021-11-23
Feature/params

b-ma:
Hey,

Here are some updates related to #20, specifically:
- reviewed linear ramp
- implemented exponential ramp

There is one deviation from the spec (related to the implicit insertion of a `set_value` when the first event is a `linear_ramp` or an `exponential_ramp`), because that would require accessing the `audio_context.current_time` and I didn't find how to test that properly. However I followed the strategy chosen by *blink* (e.g. the *Chrome* rendering engine) and this is really a edge case, so I guess it can be fixed later.

Besides, I found new issues regarding clicks and strange noises that I commented in the novel exemple `many_oscillators_with_env`, however I'm not sure if it's a memory leak again (related to #45) or a performance issue, as the cpu usage seems related to the number of nodes (oscillators?) instantiated. I will try to dig into that to make you a better report.


b-ma:
Damn, there is an error in the `test_listener` test that I didn't catch. I guess it is related to this change I made to match the spec (from what I understand of it however): https://github.com/orottier/web-audio-api-rs/pull/56/files#diff-84943c38197e3ea9e61d5c8fb8610679129a1163949c7f512671955d02636746R337, :

> Set [[current value]] to the value of paramIntrinsicValue at the beginning of this render quantum.

(cf. https://www.w3.org/TR/webaudio/#computation-of-value)

I will try to review that on my side, but any input / opinion would be appreciated on that question

Jerboas86:
> performance issue, as the cpu usage seems related to the number of nodes (oscillators?) instantiated.

Last time, i made perf measures the bottleneck was not on the node itself (at least for osc node), but the `AudioParam` processing associated with it. Let's see what you measure

b-ma:
ok thanks for the information

src/param.rs
```diff
@@ -226,19 +262,88 @@ impl AudioParamProcessor {
...
+        self.events.push(event);
+    }
+
     fn tick(&mut self, ts: f64, dt: f64, count: usize) -> &[f32] {
-        // store incoming automation events in sorted queue
-        for event in self.receiver.try_iter() {
-            self.events.push(event);
+        // println!("> tick - ts: {}, dt: {}, count: {}", ts, dt, count);
+        // handle incoming automation events in sorted queue
+        let events: Vec<AutomationEvent> = self.receiver.try_iter().collect();
```
orottier, line 328: Ideally we avoid the allocation (`collect()`)

src/param.rs
```diff
@@ -114,14 +114,15 @@ impl AudioNode for AudioParam {
...
-    value: f32,
+    value: f32, // @todo - rename to intrisic value to match the spec?
     shared_value: Arc<AtomicF64>,
     receiver: Receiver<AutomationEvent>,
     automation_rate: AutomationRate,
     default_value: f32,
     min_value: f32,
     max_value: f32,
     events: BinaryHeap<AutomationEvent>,
+    last_event: Option<AutomationEvent>, // to compute
```
orottier, line 125: Are you sure you need the `last_event`? You're using it only for ExponentialRampToValueAtTime but I don't think it is necessary, you can set last_time and last_value equal to current time and current value, and it will give the same curve.
I may be mistaken though, let me verify this first

orottier (changes requested):
Thanks a lot! AudioParams could really use some extra love so this is very helpful
I need some more time to look into the performance issue.

src/param.rs
b-ma: Yes probably you are this is ok to store only these two values (meanwhile I also updated the linear_ramp to use them)

I will have a look at this tomorrow

src/param.rs
b-ma: ok thanks I didn't know that (I'm still not really confident with how rust works with memory I must confess)

b-ma:
> There is one deviation from the spec (related to the implicit insertion of a set_value when the first event is a linear_ramp or an exponential_ramp), because that would require accessing the audio_context.current_time and I didn't find how to test that properly.

This is now fixed and behaves as it should

src/param.rs
orottier: The reason we try to avoid allocations in the render thread is because the allocator has unreliable performance. An allocator is shared by all threads (typically in an `Arc<Mutex<Allocator>>`) so the audio thread competes with all other threads to get a lock on the allocator. In a CPU-heavy game, for example, this may cause a hiccup in the audio stream.

Interacting with the allocator comes in 3 flavours:

**Allocating:** any function that returns a heap allocated container (Vec, HashMap, Box, Rc etc see https://doc.rust-lang.org/alloc/index.html), for example `Vec/HashMap::new, collect(), split_off(), clone(), to_vec()`

**Deallocating:** any time a heap allocated object goes out of scope

**Reallocating:** If you're adding items to a container (Vec, Map, Set, Heap) and the current max capacity is reached. `push() insert() extend_from_slice() etc`

In the render thread it should be doable to avoid allocating and deallocating. For example see https://github.com/orottier/web-audio-api-rs/pull/17/files

Avoiding reallocations is tough. A strategy would be to pre-allocate all containers before launching the render thread but that would impose a maximum number of audio nodes and connections on the user.

orottier:
Great progress @b-ma . I need some more time to review all of it.
Also, the idea of doing the event bookkeeping in the control thread is interesting. Let's keep that in the back of our minds but let's continue on this path for now

src/param.rs
b-ma: Thanks for the informations! It makes sens

So if I understand well, there might be a problem there:  https://github.com/orottier/web-audio-api-rs/blob/8f8f52ac5ceeb443303c905ca8e51e45a1125ded/src/param.rs#L140         

Because even if `ChannelData` has a "has copy-on-write semantics," as `output` is modified, new memory will be allocated, no? In that case, there should be a proper `ChannelData` allocated once for all for the param `output`.

Or am I missing something?

b-ma:
Ok thanks, I will just try to fix the conflicts that just appeared (I guess related to `ulps`) a some moment, and take a break on that

src/param.rs
b-ma: Just for information, I just tried this (which looks more like an in-place write of `output[0]` to me):

> in `AudioParamProcessor.process()`
```rs
        let output = &mut outputs[0];
        output
            .channel_data_mut(0)
            .copy_from_slice(param_intrisic_value);
        output.add(input, ChannelInterpretation::Discrete);
```

instead of:

```rs
        let mut param_computed_value = inputs[0].clone(); // get new buf
        param_computed_value.force_mono();
        param_computed_value
            .channel_data_mut(0)
            .copy_from_slice(param_intrisic_value);
        // mix incoming signal
        param_computed_value.add(input, ChannelInterpretation::Discrete);

        outputs[0] = param_computed_value;
```

all test are passing and `many_oscillator_with_env` still sounds ok, but I don't if I'm missing something and how to benchmark that tough

orottier: Ah, I forgot to mention, ChannelData is really special.

Have a good look at alloc.rs. If a ChannelData goes out of scope, it is stored in a pool of readily available buffers. Allocating (or invoking copy on write) a new ChannelData is very cheap if the pool is filled. The pool is used exclusively in the render thread so there are no locks involved.
Sorry, should have mentioned that!

(and now thinking, I should probably add an architecture.md file to this repo...)

orottier: Your change in `AudioParamProcessor.process()` looks good though, much more readable!

b-ma: Thanks, I didn't understood the pool pattern indeed, good to know (`alloc.rs` slowly starts to make sens...)

orottier:
Regarding performance and clicks, what is exactly the issue?
Running `cargo run --release --example many_oscillators_with_env`, it sounds great. I have a consistent 2.5% CPU usage and 4.8 MB memory usage

tests/offline.rs
```diff
@@ -143,11 +143,18 @@ fn test_listener() {
         let listener2 = context.listener();
         listener1.position_x().set_value(1.);
         listener2.position_y().set_value(2.);
+
+        println!("listener1.x {}", listener1.position_x().value());
+        println!("listener1.y {}", listener1.position_y().value());
+        println!("listener2.x {}", listener2.position_x().value());
+        println!("listener2.y {}", listener2.position_y().value());
```
orottier, line 150: These `println`s (and the ones below) can be removed now?

src/param.rs
```diff
@@ -31,48 +31,63 @@ pub struct AudioParamOptions {
...
+    automation_rate: AutomationRate, // // treat as readonly for now
+    default_value: f32,              // readonly
+    min_value: f32,                  // readonly
+    max_value: f32,                  // readonly
     value: Arc<AtomicF64>,
     sender: Sender<AutomationEvent>,
 }
 
+#[derive(Clone)]
+pub struct AudioParamRaw(
```
orottier, line 82: Probably better to name these fields instead of tupling them

src/param.rs
```diff
@@ -112,16 +127,140 @@ impl AudioNode for AudioParam {
...
+    // Setting this attribute has the effect of assigning the requested value
+    // to the [[current value]] slot, and calling the setValueAtTime() method
+    // with the current AudioContext's currentTime and [[current value]].
+    // Any exceptions that would be thrown by setValueAtTime() will also be
+    // thrown by setting this attribute.
+    pub fn set_value(&self, value: f32) {
+        let clamped = value.clamp(self.min_value, self.max_value);
+        self.value.store(clamped as f64);
+
+        // this event is ment to update param intrisic value before any calculation is done
```
orottier, line 162: spelling: `meant`

src/param.rs
```diff
@@ -112,16 +127,140 @@ impl AudioNode for AudioParam {
...
+        } else {
+            self.context().pass_audio_param_event(&self.sender, event);
+        }
+    }
+}
+
 #[derive(Debug)]
 pub(crate) struct AudioParamProcessor {
-    value: f32,
+    value: f32, // @todo - rename to `intrisic_value` to match the spec
```
orottier, line 255: Sure, go ahead with renaming!

src/param.rs
```diff
@@ -227,92 +304,303 @@ impl AudioParamProcessor {
...
+        // events with times before or at this time, the value as calculated from
+        // these events. If automation events are removed from a given time range,
+        // then the paramIntrinsicValue value will remain unchanged and stay at its
+        // previous value until either the value attribute is directly set, or
+        // automation events are added for the time range.
         for event in self.receiver.try_iter() {
+            // param intrisic value must be updated from the set_value call
+            // but we don't want to insert these events in the queue
+            if event.event_type == AutomationType::SetValue {
+                let current_value = self.shared_value.load() as f32;
```
orottier, line 323: Are you sure this is correct? Should we not take the value from the `event`?

b-ma:
Hey, according to performances yes I just discovered this "--release" flag yesterday and had no problems neither. It's still possible to make it crash but with far more demanding conditions, e.g. trigger a sine every few ms, which is not something that happen everyday. I will try to implement a granular synth at some point, I guess that would be an interesting test bad

Sorry for crying wolf

src/param.rs
b-ma: Yes that's correct, because `shared_value` is immediately updated when `set_value` is called, and before the `SetValue` event is even dispatched. This allows to make sure the value stays consistant in the control thread before the next tick is called.

But you are right, as the value of the event is exactly the same here, it's more logical and surely better to use it instead of accessing the `AtomicF64`.

orottier:
Great work! I think the code quality in param.rs is better now, and we have `exponential_ramp` working :)

b-ma:
cool and thanks for the help :)

next step : cancelScheduledValues (before the tricky ones...)

PR #57, opened 2021-11-27
Feature/params

b-ma:
Hey, here is the `CancelScheduledValues`

(also created an `AtomicF32` to use instead of `AtomicF64` for `param.current_value`)

### DIFF
--- a/src/param.rs
+++ b/src/param.rs
@@ -1,6 +1,4 @@
 //! AudioParam interface
-
-use std::collections::BinaryHeap;
 use std::sync::Arc;
 
 use crate::alloc::AudioBuffer;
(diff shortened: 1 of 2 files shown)


src/param.rs
```diff
@@ -328,8 +364,50 @@ impl AudioParamProcessor {
...
+                                // Chrome just keeps the current intrisic_value
+                                // The spec is not very clear there, but Firefox
+                                // seems to be the more compliant:
+                                // "Any active automations whose automation event
+                                // time is less than cancelTime are also cancelled,
+                                // and such cancellations may cause discontinuities
+                                // because the original value (**from before such
+                                // automation**) is restored immediately."
+                                if next_event.time >= event.time {
+                                    let last_event = self.last_event.unwrap();
```
orottier, line 391: Is it possible that `self.last_event = None` at this point?

orottier:
Great work. I only have one remark ☝️

src/param.rs
b-ma: Good point, but no after verification that's not possible, because a linear or exponential ramps are always preceded by another event (even a set_value_at_time inserted implicitly), so if the ramp is the next event that means that at least one event has already been processed.

I added a comment in the code, as well as tests to prevent possible regressions. If you feel it's more safe, I can also add a condition, it wouldn't hurt neither

orottier: A test is fine!

Issue #58, opened 2021-11-27
AudioProcessor::process should return a boolean for lifetime control


orottier:
This is currently handled with `AudioProcessor::tail_time`
To conform to the spec of `AudioWorkletProcessCallback`, this should be refactored

Related discussion in https://github.com/orottier/web-audio-api-rs/pull/46#issuecomment-962002536

https://www.w3.org/TR/webaudio/#callback-audioworketprocess-callback

orottier:
Fixed in #59

Issue #34, opened 2021-10-21
Handle non-trivial tail time for some nodes


orottier:
Nodes:

- [x] DelayNode
- [x] BiquadFilterNode
- [x] IIRFilterNode
- [x] WaveShaperNode
- [x] DynamicsCompressorNode
- [x] (wip) ConvolverNode

From the spec at https://webaudio.github.io/web-audio-api/#iirfilternode
```
tail-time 	Yes 	Continues to output non-silent audio with zero input. Since this is an IIR filter, the filter produces non-zero input forever, but in practice, this can be limited after some finite time where the output is sufficiently close to zero. The actual time depends on the filter coefficients. 
```

_Originally posted by @orottier in https://github.com/orottier/web-audio-api-rs/pull/27#discussion_r733349438_

orottier:
Also relevant for the WaveShaperNode

orottier:
All done, great work team

PR #59, opened 2021-11-29
AudioProcessor::process now returns boolean for tail_time

orottier:
Fixing issue #58 - conform to spec

### DIFF
--- a/examples/worklet.rs
+++ b/examples/worklet.rs
@@ -87,7 +87,7 @@ impl AudioProcessor for WhiteNoiseProcessor {
         params: AudioParamValues,
         _timestamp: f64,
         _sample_rate: SampleRate,
-    ) {
+    ) -> bool {
         // single output node
         let output = &mut outputs[0];
 
@@ -104,10 +104,8 @@ impl AudioProcessor for WhiteNoiseProcessor {
                     let rand: f32 = rng.gen_range(-1.0..1.0);
                     *o = *i * rand
                 })
-        })
-    }
+        });
 
-    fn tail_time(&self) -> bool {
         true // source node will always be active
     }
 }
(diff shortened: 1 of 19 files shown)


b-ma:
Hey, looks good to me as far as I can say. Just one related remark (which could maybe go into #34) about that all `AudioScheduledSourceNode`s should probably have the same strategy that the one implemented in the `OscillatorNode`

orottier:
Thanks for having a look @b-ma 

What do you mean exactly? As I understand, the 
`impl<R: MediaStream> AudioProcessor for MediaStreamRenderer<R>`
is slightly more complex than the `OscillatorRenderer` as it can also quit due to errors or an exhausted media stream.
After your comment I did clean up the code a bit

b-ma:
Hey, I was thinking more specifically about `AudioBufferSourceNode`, `ConstantSourceNode` and `OscillatorNode` (which derive from the `AudioScheduledSourceNode`), not about the stream sources which as you say have to deal with more complex issues. 

Probably this is another discussion, and I might clearly be wrong as I didn't dig enough into the implementation, but at first sight reasoning about `AudioBufferSourceNode` as a stream reader seems quite strange to me (as it reads PCM data from an `AudioBuffer` that is fully loaded in memory), while it would be correct for `MediaElementAudioSourceNode` or `MediaStreamAudioSourceNode`. But, this is maybe also just a terminology problem.

Anyway, even if relevant, this discussion is certainly way beyond your current changes which make the code far more simple to understand!

orottier:
I see your point and will amend the ticket about smarter tail time implementations.
For audiobuffer sources, the input needs to be cut in pieces of BUFFER_SIZE samples. It could definitely be optimised

PR #62, opened 2021-12-03
Feature/params - implemented `AudioParam::CancelAndHoldAtTime`

b-ma:


orottier:
Impressive work! The tests looks very good. I'm having a more thorough look soon

b-ma:
`SetValueCurveAtTime` is done too

src/param.rs
```diff
@@ -340,6 +359,31 @@ impl AudioParam {
...
+    }
+
+    pub fn set_value_curve_at_time(&self, values: &[f32], start_time: f64, duration: f64) {
+        // @todo - An InvalidStateError MUST be thrown if this attribute is a
+        // sequence<float> object that has a length less than 2.
+
+        // When this method is called, an internal copy of the curve is
+        // created for automation purposes.
+        // let arr = values.try_into();
+        let copy = Vec::<f32>::try_from(values).unwrap();
```
orottier, line 376: You can simply call `values.to_vec()`

src/param.rs
```diff
@@ -507,9 +573,9 @@ impl AudioParamProcessor {
             // handle CancelScheduledValues events
             // cf. https://www.w3.org/TR/webaudio/#dom-audioparam-cancelscheduledvalues
             if event.event_type == AudioParamEventType::CancelScheduledValues {
-                // we don't want to `sort()` before `peek` here, because what we
-                // actually need to check is if a Ramp has already started, therefore
-                // we want the actual `peek()` at the end of the last block.
+                // peek currevent event before inserting new events, and possibly sort
```
orottier, line 576: spelling: current

src/param.rs
```diff
@@ -802,8 +937,7 @@ impl AudioParamProcessor {
                             // handle end of event during this block
                             } else {
                                 // event has been cancelled
-                                #[allow(clippy::float_cmp)] // no computation has been done
-                                if end_time != event.time {
+                                if event.cancel_time != None {
```
orottier, line 938: `is_none()` is more idiomatic, I believe

src/param.rs
```diff
@@ -300,6 +344,46 @@ impl AudioParam {
...
+            time_constant: None,
+            cancel_time: None,
+            duration: None,
+            values: None,
+        };
+
+        self.send_event(event);
+    }
+
+    pub fn set_value_curve_at_time(&self, values: &[f32], start_time: f64, duration: f64) {
```
orottier, line 369: If we need ownership of values, let's just require a `Vec<f32>`? The spec says we should make a copy indeed but that's because JS lacks ownership semantics

src/param.rs
```diff
@@ -28,28 +30,38 @@ pub struct AudioParamOptions {
...
+// ordered (using stable sort), some operation may break this ordering (e.g. `push`)
+// in which cases `sort` must be called explicitely.
+// In the current implementation of the param rendering, `sort` is called once after
+// all events have been inserted in the queue at each `tick` (with the exception
+// `CancelAndHoldAtTime` which needs a clean queue to find its neighbors, but this
+// occurs during the insertion of events)
+// After this point, the queue should be considered sorted and no operations that
+// breaks the ordering should be done.
 #[derive(Debug)]
 struct AudioParamEventTimeline {
```
orottier, line 66: I have some suggestions to put the `sorted/dirty` state into the type system here (to avoid panics and avoid duplicate sorts) but I will just put them in a PR later. The timeline is surely fine for now

src/param.rs
```diff
@@ -470,10 +621,138 @@ impl AudioParamProcessor {
...
+            // containing other events, but it’s ok to schedule a value curve exactly
+            // at the time of another event.
+            if event.event_type == AudioParamEventType::SetValueCurveAtTime {
+                // check if we don't try to insert at the time of another event
+                let start_time = event.time;
+                let end_time = start_time + event.duration.unwrap();
+
+                for queued in self.event_timeline.iter() {
+                    if queued.time > start_time && queued.time < end_time {
+                        panic!(
```
orottier, line 724: Before releasing a 1.0 version of the lib, we should probably avoid all panics and make these functions return `Result`-like values

orottier (changes requested):
Loads of great work here. Thanks! I've made a few small comments, but it looks very shippable

src/param.rs
b-ma: I guess I see what you mean, but I'm not completely sure that's what we want here

For example,  in JS that would be very simple to end up writing something like that (e.g. granular synthesis w/ fancy dynamic window):

```js
const someCurve = new Float32Array(/*...*/); // some windowing function

someScheduler.tick(time => {
  const env = audioContext.createGain();
  env.setValueCurveAtTime(someCurve, time, duration);
  // or even (e.g. kind of `slice`?)
  env.setValueCurveAtTime(someCurve.subarray(start, end), time, duration);

  // @note - also, as `setValueCurveAtTime` makes an internal copy of the given values, `someCurve` 
  // could be mutated (e.g. scaling, widening) at any point without affecting already scheduled automations
}, period);
```

If `set_value_curve_at_time` takes ownership of `values`, such pattern wouldn't be possible anymore and the user would have to re-create the window each time, or am I missing something?

src/param.rs
b-ma: ok cool

src/param.rs
b-ma: Yup, I totally agree with that, I just tried to `panic` at places where the spec says an error should be thrown (I missed some of them though) thinking it would be easier to replace them later with a more appropriate error.

I have made a list of todos (spec stuff, optimisations, etc.) while writing the code, I will put all that in the issue #20 (or in a new one if you think it's more appropriate)

orottier: That sounds great

src/param.rs
orottier: Well someone has to clone the values, either the lib or the user. By putting the burden on the user, it will be possible to avoid useless clones (e.g. when the user is not interested in re-using the values). In any case, the current implementation is fine. Maybe something to add to the discussion in #20

orottier:
Great success

PR #63, opened 2021-12-08
Support cyclic audio graph when it contains DelayNodes

### LINKED ISSUE #26: Support cyclic audio graph when it contains DelayNodes
https://webaudio.github.io/web-audio-api/#rendering-loop

Cycles are currently fully muted

orottier:
It's not 100% spec compliant, but it's a decent step to feature completeness.
Check it out with `cargo run  --release --example cyclic_graph`

I've tried to document the trade-offs in `src/node/delay.rs`

Fixes #26

### DIFF
--- a/src/node/delay.rs
+++ b/src/node/delay.rs
@@ -228,6 +228,10 @@ impl AudioProcessor for DelayReader {
         // calculate the delay in chunks of BUFFER_SIZE (todo: sub quantum delays)
         let quanta = (delay * sample_rate.0 as f32) as usize / BUFFER_SIZE;
 
+        // a delay of zero quanta is not allowed (in cycles, we don't know wether the reader or
+        // writer renders first and the ordering may change on every graph update - causing clicks)
+        let quanta = quanta.max(1);
+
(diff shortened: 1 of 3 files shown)


src/node/delay.rs
```diff
@@ -24,15 +28,31 @@ impl Default for DelayOptions {
 }
 
 /// Node that delays the incoming audio signal by a certain amount
+/*
+ * For simplicity in the audio graph rendering, we have made the conscious decision to deviate from
+ * the spec and split the delay node up front in a reader and writer node (instead of during the
+ * render loop - see https://webaudio.github.io/web-audio-api/#rendering-loop )
+ *
+ * This has a drawback: a delay of 0 is no longer possible. This would only be possible if the
```
b-ma, line 36: "a delay of 0 is no longer possible, and the minimum delay is 128 samples (e.g. ~2.9ms at 44.1kHz)."

b-ma:
Quite funny example!

src/node/delay.rs
orottier: Added to the public docstring 👍

PR #65, opened 2021-12-10
Refactoring render quantum & co

b-ma:
There is a bunch of changes here:

- renamed `BUFFER_SIZE` to `RENDER_QUANTUM_SIZE`
- renamed `alloc::AudioBuffer` to `alloc::AudioRenderQuantum`
- renamed `alloc::ChannelData` to `alloc::AudioRenderQuantumChannel`
- moved all channel Config stuff (e.g. ChannelCountMode etc.) from `buffer.rs` to `node/mod.rs`
- moved `graph`, `alloc` and `process` into `renderer` mod

We didn't discuss for the changes about `ChannelConfig` but it felt weird to have that into `buffer.rs`, hope it's ok for you.


orottier:
Very good. Moving ChannelConfig is also fine for now.
I just realized I like `render` better than `renderer` actually. Do you agree?

b-ma:
Hey, cool!

I just put `renderer` instead of `render` because it sounds more as a noun than as a verb to me but, I'm not native english speaker so I might mistaking and, I don't have any strong opinion on that point neither, so I would say "as you prefer" :)

Let me know, I can change that in few minutes and add a commit

orottier:
Yes, thanks. `render` is a bit more succinct

orottier:
I have merged but please have a look at the commit I added in main.
It's always good to have a look at `cargo doc --lib --open`

b-ma:
ok thanks, I will try to be more clean with that (...and to have a closer look at rust documentation syntax)

Issue #55, opened 2021-11-19
Optimization in audio graph traversal


orottier:
Some improvements could be made in graph.rs when rendering an audio quantum:

- [x] Store edges more efficiently (currently a single HashMap that needs to be iterated many times)
- [ ] Use a specialized container type for the nodes. e.g. https://crates.io/crates/intmap
- [x] Avoid the remove and insert calls of the currently processing node. That was necessary for borrow reasons. Wrap nodes in `Cell` or equivalent
- [x] Clear the input buffers when processing of a Node is done (we don't need them anymore and this way they can be reused)
- ~~When a Node has only one outgoing connection in the graph, its outputs can be moved instead of copied to that Node's inputs~~ this is not useful because the inputs are immutable anyway and need to be copied/mutated to outputs
- [ ] Decouple the graph-topology code from the audio-specific code

orottier:
I have a branch ready for item 1 at feature/intmap-for-graph
But I will have a look at the other items first before deciding if this is worth the hassle of an exotic dependency

orottier:
I took a stab at `Avoid the remove and insert calls of the currently processing node.` @b-ma but it's very tricky. We will need a lot of unsafe code to work around it. Not sure if that would be beneficial for the project. I tried a safe, intermediate solution but it had no benefits: 084d8187f7307ab4ca

I will have another look at `Use a specialized container type for the nodes` now

b-ma:
> I tried a safe, intermediate solution but it had no benefits: https://github.com/orottier/web-audio-api-rs/commit/084d8187f7307ab4ca96d5ca95749454aa380eb2

Just tested, no sign of improvement neither sorry... 

I just wonder if we could not also try to bypass the `HashMap` altogether to use some kind of `Vec<Option<Node>>` for nodes and delegating to something like https://docs.rs/index-pool/latest/index_pool/ to manage the indexes. Then in the graph parsing we could just add/remove the nodes like that `let node = self.nodes.swap(index, None)`, or is it silly?

b-ma:
I just re-tested using `RefCell<Node>` in the `HashMap` and managed to have it working this time (slowly understanding some stuff :). It's there https://github.com/orottier/web-audio-api-rs/compare/main...b-ma:web-audio-api-rs:test/graph-render and the perf improvements are quite good (better than chrome on several test cases :)

**before**

```
+ id      | name                                                                     | duration (ms) | Speedup vs. realtime  | buffer.duration (s)
[...]

PR #66, opened 2021-12-13
Store audio graph edges more efficiently, refactor audio graph rendering

orottier:
relates to #55, specifically
- Store edges more efficiently
- Clear the input buffers when processing of a Node is done
- refactoring/splitting the mods

It also has a lot lower CPU usage. `many_oscillators_with_env` goes from 7.3% usage to 4.7%


b-ma:
Cool, looks interesting ! I'll try to have a closer look tomorrow, probably not much to say but interested to understand better how this works (...the AudioBufferSource (and granular synth :) is not far away on my side, probably some PR draft tomorrow)

src/render/graph.rs
```diff
@@ -1,190 +1,22 @@
...
-                // this is the last chunk, and it contained less than RENDER_QUANTUM_SIZE samples
-                let channel_offset = data.len() / self.channels;
-                debug_assert!(channel_offset < RENDER_QUANTUM_SIZE);
-                self.buffer_offset = Some((channel_offset, rendered.clone()));
-            }
-        }
-    }
+/// Connection between two audio nodes
+struct Edge {
+    /// index of the current Nodes input/output port
```
b-ma, line 12: it is not "output port" only?

src/render/graph.rs
```diff
@@ -1,190 +1,22 @@
...
+    self_index: u32,
+    /// reference to the other Node
+    other_id: NodeIndex,
+    /// index of the other Nodes input/output port
+    other_index: u32,
```
b-ma, line 17: it is not "input port" only? e.g. an Edge does have any direction from output to input?

b-ma:
(I don't really understand why sometimes github consider it can be "conversation" and why sometimes it should be a "review comment", anyway...)

All these changes seems pretty good to me. My only comment would be that the logic behind how the graph is parsed still remains a bit magical, that would be really nice to have a few comments to help understand the logic behind.

src/render/graph.rs
orottier: It can be either an incoming or outgoing edge, so I cannot tell at this point!
But do not despair, I'm cooking up a plan to only store the outgoing edges so this confusion will no longer occur. This does mean I need to change the graph topological sorting algorithm which will take some more time

orottier:
Thanks for the review. More PRs will follow and I will definitely put more comments on the Graph code!

Issue #64, opened 2021-12-09
`AudioBuffer`s remarks, questions and some proposals


b-ma:
Hey, I open this issue because reading the code and trying to understand it, there are still some aspects that I struggle with and that I think could be clarified and maybe improved.

**disclaimer:** I just go through my understanding to try to be clear, I don't pretend anything is ground truth here

So, my impression is that there should be 3 different types of audio buffers while there are only 2 for now:

1. The buffers used to compute blocks of signal (i.e. `alloc::AudioBuffer`). Nothing special to say about this one except that it could be renamed to `AudioBlock`, `AudioBus` or `AudioRenderQuantum` to avoid confusion with the `AudioBuffer` exposed by the API (which would also lead to renaming `BUFFER_SIZE` to `BLOCK_SIZE` or `RENDER_QUANTUM_SIZE` for consistency)

2. The `AudioBuffer` as defined by the WebAudio API, which is only consumed by the `AudioBufferSourceNode` and the `ConvolverNode` or returned through a `Promise` by `OfflineContext.startRendering`. This one should be completely loaded in memory for very fast and accurate access by the nodes that consume it, `audioContext.decodeAudioData` should therefore work from a asset (e.g. some audio file) completely loaded in memory (retrieved from network (XHR call in JS land) or file system) and perform the following steps:
    - 1. decode to PCM data
    - 2. resample to the actual `audioContext.sampleRate`
    - 3. return an `AudioBuffer` that can be consumed without further processing by e.g. the `AudioBufferSourceNode`

3. Some other audio buffer which is more related to a streaming paradigm (data is received on the fly with questions of buffering, back pressure, etc.) and can be piped to some WebAudio nodes but is defined or related to other specs such as WebRTC (microphone, audio streams from network and `MediaStreamAudioSourceNode`) or HTML (`<audio>` tag and `MediaElementAudioSourceNode`). In that case decoding and resampling can only be done on the fly when the chunks are received, and the WebAudio nodes do not expose any `start`, `stop` or `seek` possibilities.

In the current implementation, my impression is that the `AudioBuffer` defined `buffer.rs` try to handle both 2. and 3. paradigms in the same way and, is finally more close to 3. and not really optimized for 2.

Do this makes sens or am I mistaking somewhere here? If you think it's a good idea, I would be happy to try to propose something really dedicated to 2.

#### Related issues

- #30
- #42

orottier:
[...]

Issue #42, opened 2021-11-02
Add DecodeAudioData to AudioContext


Jerboas86:
The implementation should follow this part of the specification (see [DecodeAudioData](https://www.w3.org/TR/webaudio/#dom-baseaudiocontext-decodeaudiodata))

orottier:
How would you translate it to rust primitives? Maybe make it generic over `R: Read`. I don't think the project is mature enough to start with async functionality, let's put that burden on the users of this lib for now

b-ma:
- use https://github.com/pdeljanov/Symphonia for decoding?
- resample decoded audio to current `AudioContext` sample rate

orottier:
I'm contemplating to pull this in before the v1.0 release. Bringing in Symphonia will probably change the way we expose the Wav/OggVorbis-decoder now (we may even drop them in favour of a more generic MediaElement constructor).

I'm self assinging for now and see where it takes us!

orottier:
Implemented in #82

b-ma:
Hum, I have the impression there are some issues with the new decoding: 

- in `sample.wav` there is a click at end of the file that was not there before (you can ear it in the `trigger_soundfile` example)
- in `sample.ogg` there are some weird artifacts everywhere (that may be due to resampling because the file sample rate - 32kHz apparently - does not match my system sample rate).

I have also this error popping up sometimes that I don't think I have seen before (not always tough, I will try to find out in which case it appears):
```
[2022-01-05T14:32:50Z ERROR web_audio_api::media::decoding] next packet err IoError(Custom { kind: UnexpectedEof, error: "end of stream" })
```

I will also create a small test/example with several encodings of `sample.wav` created with `ffmpeg` or Audacity (so we are pretty sure the encoding is clean) to help investigate that

orottier:
Interesting, nice catch!
I will investigate the click from the wav file. I copied most of the decoding from https://github.com/pdeljanov/Symphonia/blob/master/symphonia/examples/basic-interleaved.rs but I may have made a mistake.
The ringing in sample.ogg may be due to the low quality `Resampler` that is currently included in src/buffer.rs, let's find out

orottier:
The error is a known/accepted problem: https://github.com/pdeljanov/Symphonia/issues/62 and we handle it correctly this way (I should improve the log message though)

The search for the click continues

b-ma:
[...]

PR #67, opened 2021-12-14
compliant AudioBufferSourceNode and AudioBuffer

b-ma:
Hey,

Here is a working draft, there are still a number of points that needs to be improved but the thing is working and have a clean user-facing API.

Most of the new code lies in `src/audio_buffer.rs` and `src/node/audio_buffer_source.rs`, but I had to add some few stuff here and there.

I have put a lot of comments in the code but to summarise the things I'm not very happy with:
- some types and struct are not really clean for now, but I'm pretty sure you will have some ideas on how to improve that
- `decodeAudioData` only parse wav files and don't resample, but it's ok for testing and validating the principles
- I didn't wrote much unit tests against the `AudioBufferSourceNode` but from the examples I'm quite confident there is no huge problem

I have put 2 new examples:
- `cargo run --release --example trigger_soundfile` which showcases most of the API
- `cargo run --release --example granular` which shows that it works quite well (stable ~2.4% cpu on my computer, I'm quite happy with this one :)

Let me know what you think


orottier:
Hi @b-ma stay tuned, I need some more time

src/audio_buffer.rs
```diff
@@ -0,0 +1,482 @@
+use std::rc::Rc;
+use std::sync::Arc;
+
+use crate::SampleRate;
+
+// @todo - put it there, but should leave in `BaseAudioContext`
+// let's just go with Wav files for now
+//
+// @note - what about using https://github.com/pdeljanov/Symphonia ?
```
orottier, line 9: I really like this suggestion, we should definitely check out this crate and use it

src/audio_buffer.rs
```diff
@@ -0,0 +1,482 @@
...
+// seems quite complete and efficient
+// @note - should also be async, but that's not a big deal neither for now
+pub fn decode_audio_data(file: std::fs::File) -> AudioBuffer {
```
orottier, line 12: Later (in a different PR) we should not take a file but a generic `R: Read` as input!

src/audio_buffer.rs
```diff
@@ -0,0 +1,482 @@
...
+// Having the `Arc` at the level of the single channel, will probably lead to a
+// bit more bookkeeping in the node side, but allows to be a bit more efficent in
+// terms of memory, as we can discard only one channel at a time.
+//
+// @note - spec does not define "audio asset" which can lead to confusion
+//
+// @see - <https://webaudio.github.io/web-audio-api/#acquire-the-content>
+#[derive(Clone)]
+pub(crate) struct AudioBufferData {
+    channels: Rc<Vec<Arc<Vec<f32>>>>,
```
orottier, line 112: Mixing Rc en Arc is really bad, as the result is neither thread safe nor efficient.
Just ask yourself, does this data need to cross thread boundaries (e.g. from control to render)? Then use Arc

src/audio_buffer.rs
```diff
@@ -0,0 +1,482 @@
...
+impl From<Vec<Vec<f32>>> for AudioBufferData {
+    fn from(decoded: Vec<Vec<f32>>) -> Self {
+        let number_of_channels = decoded.len();
+        let mut channels = Vec::<Arc<Vec<f32>>>::with_capacity(number_of_channels);
+
+        for channel in decoded.iter() {
+            // @note - is there a better way than copy using `to_vec`?
+            // basically, is there any way to grab the `decoded.channel` reference and
+            // put it where we want it to be without memory allocation?
+            // (sometimes JS feels so simple... :)
```
orottier, line 146: You can use `decoded.into_iter()` since you have taken ownership. The iterator Item is the actual Vec in that case

src/audio_buffer.rs
```diff
@@ -0,0 +1,482 @@
...
+
+    pub fn sample_rate(&self) -> SampleRate {
+        self.sample_rate
+    }
+
+    pub fn duration(&self) -> f64 {
+        (self.length / self.sample_rate.0 as usize) as f64
+    }
+
+    // @note - not sure if we can handle default arguments in another way
```
orottier, line 211: This is fine, or take an Optional argument

src/audio_buffer.rs
```diff
@@ -0,0 +1,482 @@
...
+    }
+
+    // [spec] According to the rules described in acquire the content either allow writing
+    // into or getting a copy of the bytes stored in [[internal data]] in a new Float32Array
+    //
+    // @note - that's really not clear and kind of "do whatever you want unless it
+    // breaks something...", so just return a copy to make sure nothing can be messed up
+    pub fn get_channel_data(&self, channel_number: usize) -> Vec<f32> {
+        let channel = &self.internal_data.channels[channel_number];
+        channel.to_vec()
```
orottier, line 289: Should we really copy the data? We can probably get away with your `get_channel_clone`, it will be way more efficient

orottier: Or just return the full slice?

src/node/audio_buffer_source.rs
```diff
@@ -0,0 +1,673 @@
...
+use super::{AudioNode, ChannelConfig, ChannelConfigOptions};
+
+// attributes shared between the node and the renderer
+//
+// @note - kind of pack both Scheduler and Controller with an additionnal
+//  `duration` entry and `seek` renamed to `offset`
+// @note - this could probably be generated by a macro! (`Vec<{name, default<type>}>`)
+//  maybe another macro could also help for Getters and Setters on the node itself
+#[derive(Clone, Debug)]
+struct SharedAttributes {
```
orottier, line 20: Is is a reimplementation of `src/control.rs`

src/node/audio_buffer_source.rs
```diff
@@ -0,0 +1,673 @@
...
+use crate::render::{AudioParamValues, AudioProcessor, AudioRenderQuantum};
+use crate::{AtomicF64, SampleRate, RENDER_QUANTUM_SIZE};
+
+use super::{AudioNode, ChannelConfig, ChannelConfigOptions};
+
+// attributes shared between the node and the renderer
+//
+// @note - kind of pack both Scheduler and Controller with an additionnal
+//  `duration` entry and `seek` renamed to `offset`
+// @note - this could probably be generated by a macro! (`Vec<{name, default<type>}>`)
```
orottier, line 17: I'd rather avoid macro's unless really necessary. It's like magic and I think we should avoid magic!

src/node/audio_buffer_source.rs
```diff
@@ -0,0 +1,673 @@
...
+// channels, duration
+// @todo - define if we could pass an `AudioBuffer` directly
+struct BufferChannelsMessage(Vec<Arc<Vec<f32>>>, f64);
+
+pub struct AudioBufferSourceNode {
+    registration: AudioContextRegistration,
+    attributes: SharedAttributes,
+    channel_config: ChannelConfig,
+    sender: Sender<BufferChannelsMessage>,
+    detune: AudioParam,        // has constraints, no a-rate
```
orottier, line 134: These two params are not actually used in rendering, am I right (also not in the current implementation)

src/node/audio_buffer_source.rs
```diff
@@ -0,0 +1,673 @@
...
+
+            if let Some(buf) = buffer {
+                node.set_buffer(&buf);
+            }
+
+            (node, Box::new(renderer))
+        })
+    }
+
+    pub fn set_buffer(&mut self, audio_buffer: &AudioBuffer) {
```
orottier, line 256: I dont think the spec allows setting buffer multiple times

src/audio_buffer.rs
b-ma: good

src/node/audio_buffer_source.rs
b-ma: yup that is the @note

src/node/audio_buffer_source.rs
b-ma: makes sens

src/node/audio_buffer_source.rs
b-ma: yes they are see `audio_buffer_source` line 427

src/node/audio_buffer_source.rs
b-ma: the spec allows to do 
```js
const src = audioContext.createBufferSource()
src.buffer = myBuffer;
```

b-ma:
For the other questions I will try to be more constructive tomorrow

b-ma:
But could you try the examples in between because I'm not completely sure we are speaking about the same thing here. If it's possible to do the `granular` example with the current implementation, I totally agree I'm mistaking and I have no problem to trash all the code

b-ma:
ok lot's of things to say, so

> Regarding 1) the current implementation of AudioBuffer is a Vec<Arc<Vec<f32>>> with some metadata. Why did you add the extra Rc? Is it because of the 'acquire the content' operations? I don't think we should port that to rust, as in rust there is the concept of ownership so we do not need to acquire/copy any data.

Several things there:
- I really believe there no huge difference between the `buffer:AudioBuffer` and the `audio_buffer:AudioBuffer`, except the second one follows the spec API. But very probably the 2 ones could be merged somehow, I just made a new thing so that is was more simple for me to think about the requirements and internal logic of the buffer and source.
- The additional `Rc` seems required to me because several sources can share the same `AudioBuffer`, e.g. it's perfectly fine, and important to be able, to do something such as (in _JSish_ pseudo-code):
```js
const buffer = getSomeOneSecondBufferSomehow(); 
const now = audioContext.currentTime;
for (let i = 0; i < 100; i++) {
   const src = audioContext.createBufferSource();
   src.buffer = buffer;
   src.start(currentTime + i * 0.01);
}
```
Here, you have 100 source that access the same `AudioBuffer` at the same time, so somehow the buffer `internal_data` needs to shared between these sources, that's why I put this extra `Rc` instead of copying the `Vec` of channels. Maybe it is not really necessary, I'm really not sure about that, but if you have a more than one channel sound file, it sounds cheaper to me to clone an `Rc` than to copy an entire `Vec`.
- The concept of "acquire the content" as I understand it (that maybe I didn't implement the proper way in `Rust`, but I'm pretty confident in my understanding of the idea here) derives from that possibility. It basically means that once a source started its rendering, the memory it will access and read from cannot be corrupted or changed by a call to `copyToChannel`, e.g.:
```js
const buffer = getSomeOneSecondBufferSomehow(); 
const now = audioContext.currentTime;
for (let i = 0; i < 100; i++) {
   if (i == 50) {
      buffer.copyToChannel(someArray, 0);
      buffer.copyToChannel(someArray, 1);
   }
   const src = audioContext.createBufferSource();
   src.buffer = buffer;
   src.start(currentTime + i * 0.01);
}
```
Here, the 50 first sources (that are still playing) will continue to read the `audioBuffer` data *as it was* when they started their rendering. That's why I added this creation of a new `Arc` and `Vec` in `copy_to_channel`, **and** the manual `clone` of the `Arc` of channels when the source starts. 
As I wrote in the comments, I'm not sure this last step is necessary, but as I couldn't find any clear informations about how `clone` behaves why nested structure, I just did it explicitly (i.e. if you clone an `Rc<Vec<Arc<Vec>>>` do you clone only the `Rc` *or* the `Rc` and the `Arc`s? If the second is true, I guess the code in `AudioBufferSource::acquire_buffer_channels` could perfectly be removed). 
The important thing here, rather than implementation details or language specific concepts, is that: **from the point of view of an `AudioBufferSourceNode`, once it has started to play, the underlying data it access cannot be modified**.

> Regarding 2) I don't directly see what is the difference is with the current implementation. If we are going to add hundreds of new lines of code, I think there should be a very clear benefit to it. The old AudioBufferSourceNode also takes a buffer, ships it to the render thread, and renders/resamples/loops on the render thread.

  The main difference (which are quite important because it's why the `AudioBufferSourceNode` is primarily made for) from what I can see is that:
- the old `AudioBufferSourceNode` works at block rate, cannot pitch because it just renders one block at a time (from what I can see).
- the new one 
   + is sub-sample time accurate (start, stop and loop), 
   + can pitch / detune and even play backward because it has access to the full buffer, 
   + follows the spec (which I guess is important here :), 
   + and do all that in a **very cheap and predictable** way as it is fundamentally a table lookup into the underlying buffer (which is already resampled with the right sample rate) with a bit of interpolation. 

All these features are *really* important if you want to go beyond simply playing an audio file (for which you'd use `MediaElementSourceNode`) and do serious real-time sample-based synthesis (e.g. sequencer, granular synthesis, concatenative synthesis, etc.)

orottier:
Hi @b-ma, thanks for your initial replies.

Edit: I just now see your reply from 11 minutes ago, I need to read that carefully.

Sidenote:
I hope you realize I greatly appreciate the effort you put in contributing to this project. No matter the outcome of this PR, you are making us question the current state of the project and thus pushing quality improvements all over the place. So, thanks!
End sidenote.

I have added a commit with a very rudimentary forward/backward scrubbing example with original implementation

It is very flawed:
- decode_audio_buffer does not exist yet so I need to put it in a media
  element
- this means I need to hardcode the buffer duration (3.0 seconds)
- since there's only 1 media element, I need to stop the previous scrub
  before I can start the next one. Causing clicks and a not nice sound
- something is off with the volume, I almost blew my eardrums out!

I will put some more work in it later, I think I can safely port your decode_audio_buffer code and shoehorn it into the current implementation. Then I can work on making my version of the example better!

orottier:
> The main differences (which are quite important because it's why the AudioBufferSourceNode is primarily made for) from what I can see is that:

>    the old AudioBufferSourceNode works at block rate, cannot pitch because it just renders one block at a time (from what I can see). 
>    the new one  ,,,
      

These are very valid points, I see now. I will get back

b-ma:
Hey, for your volume problem, I guess it's probably related to how the decoding is done in media::WavDecoder, https://github.com/orottier/web-audio-api-rs/blob/af1fe81bc06d6cd02c183254cef8bcf96c3a818b/src/media.rs#L504

which should probably be `u16::MAX as f32`, but I'm not sure to understand why you try do recreate the scrubbing example with the `MediaElementSourceNode`. I don't think this node should be used to try to do such kind of synthesis, it's meant at piping a stream from an html `<audio>` tag to the WebAudio API to add some effects or stuff like that (e.g. reverb, equalization). It handles a different set of use cases with completely different constraints (where I have the impression that the actual implementation does quite well the job). I think we kind of fallback here in the discussion of #64.

src/node/audio_buffer_source.rs
orottier: huh, somehow I missed the code below where you guard for setting it multiple times. Never mind this

orottier:
To keep stuff a bit organized, I propose to discuss 'acquiring the content' in a later issue. Okay? Let's focus on building a good AudioBufferSourceNode renderer in this PR.

b-ma:
ok sounds good to me (`setChannelData` is not something you use everyday anyway), let me know how you would like advance on that

src/audio_buffer.rs
b-ma: sounds good as it would allow to get files from the network I guess (w/ Rust equivalent of XHR?)

src/audio_buffer.rs
b-ma: I let you decide on that point

src/audio_buffer.rs
b-ma: ok cool

src/audio_buffer.rs
b-ma: No real opinion on that (I don't really like having None everywhere but...), probably the important thing is to be consistent within the library

src/audio_buffer.rs
b-ma: It just to make sure nobody can mess up a buffer that is currently played. I would say that we could just keep things like that (possibly putting a warning in the doc), and discuss it more in depth inside the new issue about "acquire the content".

src/node/audio_buffer_source.rs
b-ma: yup that's true but with the additional `duration` attribute and more consistent naming (`seek` -> `offset`)

src/node/audio_buffer_source.rs
b-ma: I think however that would be nice to have kind of `Atomic` parameter bag that could be defined per Node, with proper names and types. Don't really know how to achieve that with Rust (maybe another issue too?)

src/node/audio_buffer_source.rs
b-ma: the `computed_playback_rate` is calculated here  https://github.com/b-ma/web-audio-api-rs/blob/a60ce47ea5faa8fcffcf4b01f4d74e72ded09517/src/node/audio_buffer_source.rs#L431 and used to update the renderer internal state https://github.com/b-ma/web-audio-api-rs/blob/a60ce47ea5faa8fcffcf4b01f4d74e72ded09517/src/node/audio_buffer_source.rs#L576 and https://github.com/b-ma/web-audio-api-rs/blob/a60ce47ea5faa8fcffcf4b01f4d74e72ded09517/src/node/audio_buffer_source.rs#L577

b-ma:
Hey, so to start to clean all that a bit, what I could imagine as a next step would be to:

- move `decode_audio_data` as is into `AudioContext` (with a warning about it's only `.wav` and does not resample for now)
- add `AudioContext::create_buffer`
- create a `buffer` mod that would contain:
   + the new spec compliant `AudioBuffer`
   + the old `AudioBuffer` that could be renamed to `AudioStream`, `StreamBuffer` or `AudioStreamBuffer` ?
   + move `buffer::ChannelData` struct into its own file, because it could very probably be used by (and help to clean) the new `AudioBuffer` and `AudioBufferSourceNode`
- open some new issues regarding the new questions that arose and missing features (decoding & resampling, etc.)

Sounds good to you?

orottier:
I'm slowly starting to realize this issue is not about the actual `AudioBuffer`, but only about the `AudioBufferSourceNode`.

You are completely right that the current `AudioBufferSourceNode` is not good at all. It simply wraps a MediaElement which is a heavy weight object (you shouldn't spawn many of them per second) and is unable to effectively seek and play because it uses pre-computed blocks of samples. We should definitely drop `src/node/audio_buffer.rs` and replace it with your `src/node/audio_buffer_source.rs`

Regarding `AudioBuffer`, your implementation is identical to the current (except for the `Rc`, tbc in #68), so I strongly disagree with keeping both versions.

So I think the steps forward are:
- move decode_audio_data in its current state (only Wav decode) to context.rs and in #42 improve on it
- drop `src/audio_buffer.rs`
- extend `src/buffer.rs` with the functions copyTo/FromChannel
- drop `src/node/audio_buffer.rs` in favour of `src/node/audio_buffer_source.rs`
- change `src/node/audio_buffer_source.rs`
  - to use the existing `AudioBuffer`
  - to use the existing `Scheduler` and `Controller` (which may need to be extended a bit)

I hope you agree with it. I really don't think we need to introduce a `AudioStreamBuffer` as it will be identical to `AudioBuffer`

b-ma:
Ok cool sounds good, you certainly have a better understanding than me of how the current audio buffer is working and how it is currently used, so no problem on my side. 

*Note:* I'm not completely sure on how to share the buffer between several sources without the `Rc`, but I will try something and let you know if I run into problems.

orottier:
Great. I can pick up some of the tasks too if you are unsure how to continue. Just let me know.

In any case, you can always clone an audiobuffer. It has copy on write semantics, so a clone is cheap and the result cannot be altered by other processes. If you then modify the buffer, your clone will detach from the others and only your version is modified

b-ma:
Ok thanks, I'll let you know

> It has copy on write semantics, so a clone is cheap and the result cannot be altered by other processes.

That's something I still struggle with I guess, as much I think I kind of saw and understood a bit how this is implemented in the `AudioRenderQuantumChannel` (`make_mut` and `drop` ?), as much I fail to see where there is any similar logic in `buffer::ChannelData` (maybe another question for #68)

b-ma:
Hey, I guess the thing starts to be in good shape now. 

I tried to simplify to the maximum, using mostly the API of the `AudioBuffer`. 
Regarding the `AudioBuffer`, I just merged with the public API and passed most of the existing methods to `pub(crate)` visibility.

3 small notes:
- `copy_to_channel` will be the one to test in #68
- `get_channel_data` now returns a slice that seems to be read-only, could you confirm that?
- `resample()` should be `pub(crate)`, and as it crashed the doc examples I made the example private too, let me now if you prefer to keep that public for some reason I didn't see.

Also, probably all `ChannelData` should be `pub(crate)` now, what do you think?

I didn't merge your last PRs yet, I will do that once you confirm it looks good for you.

_aside:_ I think I start to understand what you mean regarding the acquiring the resources problem, I have an idea of a small test/example that should help to make things more clear, will try to sketch that in next days.

orottier:
I think this is going in a great direction! I will try to answer your question inline in the diff

src/buffer.rs
```diff
@@ -35,60 +42,126 @@ impl AudioBuffer {
...
 
-    /// Duration in seconds of the AudioBuffer
+    /// Duration in seconds of the `AudioBuffer`
     pub fn duration(&self) -> f64 {
-        self.sample_len() as f64 / self.sample_rate.0 as f64
+        self.length() as f64 / self.sample_rate.0 as f64
+    }
+
+    /// Copy data from a given channel to the given `Vec`
+    pub fn copy_from_channel(&self, destination: &mut Vec<f32>, channel_number: usize) {
```
orottier, line 66: destination can be `&mut [f32]` here, this allows for more flexibility (and I think, corresponds better to the spec). 
Same goes for `copy_from_channel_with_offset`

orottier: Of course, this below no longer holds then
```
        // we use the `capacity` instead of `len` because destination is a write
        // buffer and we don't care of its current values (cf. `copy_to_channel`)```

But that would be a good thing. The caller should decide explicitely how many samples they want to receive

File: src/buffer.rs [hunk 12]
```diff
@@ -157,25 +231,25 @@ impl AudioBuffer {
...
+    // buffer.resample(SampleRate(96_000));
+    //
+    // assert_eq!(
+    //     buffer.get_channel_data(0)[..],
+    //     vec![1., 1., 2., 2., 3., 3., 4., 4., 5., 5.][..]
+    // );
+    //
+    // assert_eq!(buffer.sample_rate().0, 96_000);
+    // ```
+    pub(crate) fn resample(&mut self, sample_rate: SampleRate) {
```
orottier, line 252: We should definitely replace this version later with the library we are already using

src/buffer.rs
```diff
@@ -35,60 +42,126 @@ impl AudioBuffer {
...
+    /// Return a read-only copy of the underlying data of the channel
+    pub fn get_channel_data(&self, channel_number: usize) -> &[f32] {
+        // [spec] According to the rules described in acquire the content either allow writing
+        // into or getting a copy of the bytes stored in [[internal data]] in a new Float32Array
+        self.channel_data(channel_number).as_slice()
     }
 
     /// Channel data as slice
-    pub fn channels(&self) -> &[ChannelData] {
+    pub(crate) fn channels(&self) -> &[ChannelData] {
```
orottier, line 131: Marking these function `pub(crate)` is good for now, as they are not part of the spec

src/buffer.rs
```diff
@@ -35,60 +42,126 @@ impl AudioBuffer {
...
+        // want to be sure we don't access some undefined index (cf. `copy_from_channel`)
+        let src_len = source.len();
+        let max_frame = (self.length() - offset).min(src_len).max(0);
+        let channel = self.channel_data_mut(channel_number).as_mut_slice();
+
+        channel[offset..(max_frame + offset)].copy_from_slice(&source[..max_frame]);
+    }
+
+    /// Return a read-only copy of the underlying data of the channel
+    pub fn get_channel_data(&self, channel_number: usize) -> &[f32] {
```
orottier, line 124: Returning a slice is fine!

orottier:
Let's drop `bench.rs`, it was a silly way to benchmark in the past when I was trying out different data structures for the buffers

b-ma:
Damn, that unfolded another problem I didn't see in `test/media.rs` where the `SlowMedia` iterator directly creates an `AudioBuffer` from `ChannelData`:

```rs
let channel_data = ChannelData::from(vec![self.value; RENDER_QUANTUM_SIZE]);
let buffer = AudioBuffer::from_channels(vec![channel_data], self.sample_rate);
```

I guess a good option here could be that the `SlowMedia` iterator should return a simple `<Vec<[f32]>>` while `media::MediaElement` (or maybe `node::MediaStreamRenderer`) would be the one in charge of creating the `AudioBuffer` for buffering and resampling. I guess that would make "external" media producers more simple and generic, what do you think?

But maybe it's best if you can have a look on that, as I'm not really familiar with these streaming nodes, as you prefer.

orottier:
Sure, I'll have a look. Can you push the changes so far, even though the build/tests will fail?

b-ma:
Ok thanks, just pushed my last changes (w/ failing build):
- `ChannelData` and `AudioBuffer::from_channels` to `pub(crate)`
- deleted `bench.rs`

orottier:
Hey @b-ma I added my changes. Is this ready for final review now or are you still considering new changes?

b-ma:
Hey, thanks ! No I think I'm good on my side

src/buffer.rs
b-ma: you mean the `rubato` lib here?

src/buffer.rs
```diff
@@ -204,8 +257,9 @@ impl AudioBuffer {
 /// Single channel audio samples, basically wraps a `Arc<Vec<f32>>`
 ///
 /// ChannelData has copy-on-write semantics, so it is cheap to clone.
+// @note - should be `pub(crate)` but used in `bench.rs` examples
```
b-ma, line 260: remove

src/buffer.rs
```diff
@@ -227,9 +281,10 @@ impl ChannelData {
         self.data.len()
     }
 
-    pub fn is_empty(&self) -> bool {
-        self.data.is_empty()
-    }
+    // never used
```
b-ma, line 284: delete or keep that around?

b-ma:
I'll do a short pass on the comments tomorrow, there are a few typos and irrelevant/outdated comments here and there

src/control.rs
```diff
@@ -24,6 +26,8 @@ impl Scheduler {
         Self {
```
orottier, line 26: The `Scheduler` is the driver of the `AudioScheduledSourceNode` trait which corresponds to https://www.w3.org/TR/webaudio/#AudioScheduledSourceNode
We should probably move offset & duration to the `Controller`. The Controller (not part of the spec) is also used for MediaElements but I will figure out a way to actually use these fields there later

src/buffer.rs
orottier: yes!

orottier (commented):
Almost ready to merge I think. I left some nitpicks

b-ma:
ok, changes are done!

orottier:
Great work!

b-ma:
Cool ! and thanks for your help and patience on that

PR #69, opened 2021-12-17
More effient graph-edge storage, some docs

orottier:
Hey @b-ma , for your information. Docs are still a work in progress!

### DIFF
--- a/src/analysis.rs
+++ b/src/analysis.rs
@@ -1,3 +1,7 @@
+//! Helpers for time domain and frequency analysis
+//!
+//! These are used in the [`AnalyserNode`](crate::node::AnalyserNode)
+
 use crate::render::AudioRenderQuantumChannel;
 use crate::RENDER_QUANTUM_SIZE;
 
@@ -5,8 +9,10 @@ use realfft::{num_complex::Complex, RealFftPlanner};
 
 use std::f32::consts::PI;
 
-const MAX_QUANTA: usize = 256;
-const MAX_SAMPLES: usize = MAX_QUANTA * RENDER_QUANTUM_SIZE;
+/// FFT size is max 32768 samples, mandated in spec
+const MAX_SAMPLES: usize = 32768;
+/// Max FFT size corresponds to 256 render quanta
+const MAX_QUANTA: usize = MAX_SAMPLES / RENDER_QUANTUM_SIZE;
 
 /// Blackman window values iterator with alpha = 0.16
 pub fn generate_blackman(size: usize) -> impl Iterator<Item = f32> {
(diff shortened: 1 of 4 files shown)


b-ma:
Hey, looks pretty good and comments are really helpful to understand better the graph logic. I ran through the whole `graph.rs` file this time and I have 1 related question / remarks, why is this necessary to loop through all the nodes: 

```rs
// visit all registered nodes, depth first search 
self.nodes.keys().for_each(|&i| {
  self.visit(
    i,
    &mut marked,
    &mut marked_temp,
    &mut ordered,
    &mut in_cycle,
  );
});
```

At first sight, I wonder why this is necessary to loop through all `self.nodes` because as the graph is guaranteed to have one single output (.e.g. `listener`) and `visit` is recursive, something like (without the loop) should do the trick no?

```rs
// visit all registered nodes, depth first search 
self.visit(
  ListenerIndex,
  &mut marked,
  &mut marked_temp,
  &mut ordered,
  &mut in_cycle,
);
```
I would imagine that this could gives you all the nodes that are currently connected, and would more be deterministic as it wouldn't rely anymore on `keys` and as the connections are sorted in `SmallVec`s?

orottier:
> Why is this necessary to loop through all the nodes there?

Great question. I think the main thing to realize is that there may be pieces of the audio graph that are not connected to the destination node.

For example, you may want to use the web-audio-api to create a visual spectogram of a sound input. In this case, you don't want to play the audio on the speakers, but you want to use an `AnalyzerNode` and write the frequency values to a html `canvas`. This means we will never reach the audio input node if we would start from the destination node and visit the connected nodes from there.

My implementation is much inspired by https://webaudio.github.io/web-audio-api/#rendering-loop . Definitely worth reading.

Sidenote, the recursive implementation is a bit strange in rust because of all the mutable borrows I need to carry around. I may rewrite it into a sequential (non recursive) variant. Let me check if that makes it more readable.

b-ma:
Ok good point, really didn't think of these cases (that would make a nice comment in the code :). For the recursive implementation I think it's quite readable actually, there are some `Rust` weirdness for sure... but it's also kind of pedagogical for newbies such as me.

orottier:
> note: Maybe it would be interesting to try to see with the spec people if defining a kind of Output or Sink type (which can be only a few nodes I guess), just as there are well defined Sources wouldn't worth the effort?

Interesting thought, but I think this will be hard in practise. For example, you could set up a microphone input and connect it to a large delay node. Then after a minute or so, you could connect the delay node to the output node. In this case, the mic and delay should really render every frame, even though they are not (yet) connected to the output. Basically any non-pure Node (means: having side effects such as caching, moving data to user side, etc) should be considered a Sink in this case, which could be quite many of them. This would decrease the usefulness in my opinion

PR #70, opened 2021-12-22
Improved `DelayNode`

b-ma:
Hey, here are new features on the `DelayNode` (and probably my last PR before some holiday break :)

So, what's new here:
- sub-sample rendering
- handle tail-time
- partial support of number of channel change in input
- some unit tests

Two examples:
- `cargo run --release --example simple_delay`
- `cargo run --release --example feedback_delay`

The second one is very much an improved version of your `cyclic_graph.rs` example, let me know if you think it worth the case to keep both and / or which one you prefer to keep.

I have the impression the node is now mostly compliant except for:
- the zero delay thing (see note below)
- the fact the up/down mixing of the internal ring buffer when the number of input channels change is hardcoded to `ChannelInterpretation::Speaker`, because I didn't find how to access the current value from the renderer

**About the zero delays problem**
_Note that I didn't try to implement that because it mainly concerns the graph and maybe this has implication I don't see, but I have the impression it could be possible to achieve that at mostly no cost._

The idea would be to create a dummy connection between `DelayReader` and `DelayWriter` in the `DelayNode::new` constructor, this connection would do almost nothing (as basically the writer do not use `output` and the reader do not use `input`) expect guarantee the order of the processing if the delay is not in a loop. In the graph process, if the node is found in a cycle, the graph could just delete this connection and somehow flag the `Reader` as "in_cycle" so that the later would know it must clamp its minimum delay to the quantum duration. (I guess that's not that far from what is described in the spec, and I don't even think this would need to be reversible, e.g. once in a cycle the node behaves like that forever even if the cycle is broken later, which is really a weird edge-case I can't even imagine where users should anyway know what they are doing)

Let me know what you think


orottier:
I only had a quick look but this looks very promising! I'll think about your suggestion about the cycles, it makes a lot of sense. Stay tuned for a more in depth review

orottier:
Your implementation of the DelayNode tail time got me thinking a lot. The current implementation for decomissioning nodes is 
```
// in src/render/graph.rs
/// Determine if this node is done playing and can be removed from the audio graph
fn can_free(&self, tail_time: bool) -> bool {
    // Only drop when the Control thread has dropped its handle.
    // Otherwise the node can be reconnected/restarted etc.
    if !self.free_when_finished {
        return false;
    }

    // Drop, if the node has no outputs connected
    if self.outgoing_edges.is_empty() {
        return true;
    }

    // Drop, when the node does not have any inputs connected,
    // and if the processor reports it won't yield output.
    if !self.has_inputs_connected && !tail_time {
        return true;
    }

    false
}
```

Reading the spec more carefully, specifically thinking about the DelayWriter, this snippet must be dropped
```
    // Drop, if the node has no outputs connected
    if self.outgoing_edges.is_empty() {
        return true;
    }
```
Because any Node may have side effects (such as the DelayWriter, it just caches data and makes it available to another node without connecting to it). It has no outputs connected but should still render every quantum. I'll try to see what happens when I drop that code

orottier:
I encountered a bug in the DelayNode which was probably also present in the earlier version. Added a failing test for it, care to have a look?

src/node/delay.rs
```diff
@@ -29,6 +29,41 @@ impl Default for DelayOptions {
 ///
 /// The current implementation does not allow for zero delay. The minimum delay is one render
 /// quantum (e.g. ~2.9ms at 44.1kHz).
+///
```
orottier, line 32: I really like the style of documenting each node in detail. I will update my panner-node PR in the same fashion, and let's improve over time the other node's docs

b-ma: Ok cool I'll also continue on that vein too then

b-ma:
> I encountered a bug in the DelayNode which was probably also present in the earlier version. Added a failing test for it, care to have a look?

Nice finding! Don't really see why this could happen but of course I will have look at that

b-ma:
Kind of tricky edge case actually, nice you spotted it, should be fixed now

b-ma:
Ok, I added a new check to internally clamp `max_delay_time` to quantum duration. Maybe this will need to be reviewed a bit when allowing delays smaller the quantum duration (w/ some new edge cases I didn't see) but a priori it should behave as expected and gracefully fallback if found in a loop.

orottier:
Thanks for edge case solution!

FYI I'm still working on the lifetime management fix. Apparently, the snippet that I suggested to remove from `can_free` was a necessity to clean up AudioParams feeding into already decommisioned nodes. If I remove the snippet, once again the example file `many_oscillators` has a memory leak. To be continued..

orottier:
Sorry, I got stuck in the graph lifecycle rabbit hole for a while.
I pushed a fix for my faulty logic in the AudioGraph's `can_free` function.

Furthermore I wanted to point out that I think your tail_time logic is not correct. I added a (sometimes, you should run it many time) failing test case that should point out the flaw. In the commit message I added a suggestion for a fix. I think we don't need to keep track of timestamp of silence input, but instead the Reader end should be able to notice if the Writer end has been dropped.

I hope I was able to make clear what is the issue. Do you think you can fix it? Otherwise I can pick it up, no worries.

b-ma:
Yup, I will have a look and let you know if I get lost somewhere.

For the lifecycle management thing maybe it would be helpful to have an issue (or whatever) dedicated to compiling and summarizing all the cases (and possibly the weirdest edge ones) we can think of. 
From what I can imagine right now your changes seems to make real sens (...I actually had an idea of another case relying on checking outputs but it also works the other way around :)

b-ma:
ok, I guess it follows more or less your idea, let me know what you think!

(also made some cleaning here and there)

orottier:
I've merged the work so far and added a new issue for the zero-delay thing.

As you may have noticed I am focussing now on the roadmap for v1.0 release. This is definitely something we can pick up in a 1.* release

Issue #44, opened 2021-11-02
More PannerNode features


orottier:
https://www.w3.org/TR/webaudio/#PannerNode
https://www.w3.org/TR/webaudio/#Spatialization

- [x] `DistanceModelType` option (good first issue)
- [x] options `refDistance`, `maxDistance`,  `rolloffFactor` (good first issue)
- [ ] handle stereo input (currently mono is assumed)
- [x] a-rate processing (take care for `single_valued` AudioRenderQuantum)
- [x] sound cones https://www.w3.org/TR/webaudio/#Spatialization-sound-cones
- [x] `PanningModelType` `HRTF` (lots of work)

orottier:
Closing in favour of #240 and #241

PR #71, opened 2021-12-23
Add cone gain feature to PannerNode

orottier:
regarding #44


b-ma:
Hey, looks pretty good to me (disclaimer: I'm really really not a spatial audio person...), two small details on my side:
- rename `examples/panner-cone.rs` to `panner_cone.rs` to be consistent with other examples files
- would be nice to have a link to https://webaudio.github.io/web-audio-api/#Spatialization-sound-cones in the code

And one question: the example has some noisy artifacts on the edges of the rotation, do you think this is due to the k-rate processing?

orottier:
> And one question: the example has some noisy artifacts on the edges of the rotation, do you think this is due to the k-rate processing?

Interesting. I had not noticed yet. The narrow cone indeed is noisy.
I've tried running the example again with smaller value for `RENDER_QUANTUM_SIZE`. A value of 16 removes the crisp for me. So I think we can conclude it's indeed due to k-rate processing

PR #72, opened 2021-12-25
Implemented `AudioScheduledSourceNode` interface on `ConstantSourceNode`

b-ma:
Sample accurate start / stop in `ConstantSourceNode` 

2 small notes:
- sub-sample accuracy in the computation of values is already handled by the `offset` param
- I renamed the file to be more consistent


orottier:
Very nice, thanks again. If you wish you can open a PR with some more docs and examples but we can also pick that up later.

b-ma:
Yup I'll do it for sure!

Issue #50, opened 2021-11-10
Change sample rate type from SampleRate(u32) to SampleRate(f32)


orottier:
Which conforms to the spec: https://www.w3.org/TR/webaudio/#dom-baseaudiocontext-samplerate

I'd like to keep the newtype wrapper in place so we don't accidentally mix up a f32 sample_rate parameter with a f32 timestamp one. See related discussion https://github.com/orottier/web-audio-api-rs/pull/33#issuecomment-949365714

b-ma:
Just to note, timestamps should probably be `f64`: [cf. https://www.w3.org/TR/webaudio/#dom-baseaudiocontext-currenttime](https://www.w3.org/TR/webaudio/#dom-baseaudiocontext-currenttime) (which make sens for long running applications)

orottier:
Fixed in #80

PR #80, opened 2021-12-30
Change SampleRate type from u32 to f32 (issue #50)

orottier:
To be fair, u32 is the better choice as it disallows negative values, and has better numerical precision 🤷‍♂️


b-ma:
I'm a bit off these days, but I will have a look soon

orottier:
No hurries, enjoy the holidays.

I'm actually reconsidering this PR. To continue using u32 internally but expose sample rate as f32 to the user. This would be conform the spec and still have the benefit of u32. All serious rust audio libs seem to use u32 for sample rates..

b-ma:
Hey, nothing much to say there... I also find the `float` sample rate rather strange (never seen a sample rate with a decimal part), tried to look into the spec issues to see if this has been discussed somewhere and understand the rational but couldn't anything neither...

orottier:
Alright, I have switched back internally to u32. The sample_rate function on the AudioContext still returns f32 but you can now use `sample_rate_raw` to get the high precision value

Issue #77, opened 2021-12-29
"Eat your own dog food" for release candidate


orottier:
Port a few examples/tutorials from MDN/others to a rust version and see what we run into. For example:

- https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API/Using_Web_Audio_API
- https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API/Advanced_techniques
- https://github.com/mdn/voice-change-o-matic
- https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API/Web_audio_spatialization_basics

b-ma:
I have a couple of examples that I generally use for teaching (AM, granular, step sequencer, etc.) that I could port too. What I don't really see is what strategy to use to create interfaces which can become a real burden (I was thinking about OSC + PureData to delegate all interface stuff to Pd, but it requires to install something...), do you have any ideas/preferences here?

orottier:
Anything goes! I'm building a mic input recording, with some output effects. I noticed a few things already which I fixed in #83 
If you need external things you can just create a new `cargo new --bin` with the web-audio-api-rs as a dependency. We don't need to commit all examples into this repo. Is that what you mean?

b-ma:
No not exactly, I was rather wondering which strategy would be best to create interactive examples with GUIs. I know that creating clean and reactive GUIs can rapidly become complicated and I really have no idea of the Rust ecosystem there. So I was wondering if you had any preferences in term of Rust lib for creating GUIs? 

In any case, I realize having some example working with OSC would be a nice thing anyway, I will try to sketch something with PureData, and we will see where it gets us

orottier:
Ah, I understand now. Using a GUI is indeed quite a bit of work and may actually distract from the goal of just building code around the audio API. It can be a lot of fun though!

For my microphone example I will use https://docs.rs/termion/latest/termion/ to let the user input some settings and maybe display some very crude graphs in the terminal. I will try to push some work in progress soon with it

orottier:
I think we have proven our lib works in various use cases, closing for now

PR #82, opened 2022-01-03
Media decoding via Symphonia (for context.decode_audio_data and media streams)

orottier:
I think this should be the way forward. Have a single MediaDecoder instead of structs for Ogg/Wav etc


b-ma:
Cool looks really good !

examples/media_element.rs
```diff
@@ -0,0 +1,25 @@
...
+    let context = AudioContext::new(None);
+    let node = context.create_media_element_source(element);
+
+    // play media
+    node.connect(&context.destination());
+    node.start();
+
+    // enjoy listening
+    std::thread::sleep(std::time::Duration::from_secs(4));
+}
```
b-ma, line 25: Not related to this PR but, in JS this would be written:

```js
const media = new Audio("sample.ogg"); // this is the JS constructor of the html `<audio>` tag
// or
// const media = new Audio();
// media.src = "sample.ogg"
// -> diff - the decoder is hidden in the MediaElement here
const src = audioContext.createMediaElementSource(media);
src.connect(audioContext.destination);

media.start();
// -> diff - this is the media element which holds the start / stop / seek / loop methods, not the WebAudio node
```

I don't say this should be written exactly the same way in Rust, because `Audio` is basically not part of the WebAudio API, but the `MediaElementSourceNode` is maybe more a problem (at least it should be documented as a divergence according to the spec).

src/context.rs
```diff
@@ -82,82 +82,57 @@ pub trait AsBaseAudioContext {
...
-        // and resample it to the sample-rate of the BaseAudioContext if it is
-        // different from the sample-rate of audioData.
-
-        // @todo - resample if needed
-        // if (sample_rate != self.sample_rate()) {}
-
-        let mut channels = Vec::<ChannelData>::with_capacity(number_of_channels);
+    /// The current implementation can decode FLAC, Opus, PCM, Vorbis, and Wav.
+    ///
+    /// Contrary to the official spec, the input parameter can be any byte stream (not just an
```
b-ma, line 89: Maybe the term `Contrary` is misleading here, as the implementation is just more general than the spec

src/context.rs
```diff
@@ -82,82 +82,57 @@ pub trait AsBaseAudioContext {
...
-
-        // [spec] Take the result, representing the decoded linear PCM audio data,
-        // and resample it to the sample-rate of the BaseAudioContext if it is
-        // different from the sample-rate of audioData.
-
-        // @todo - resample if needed
-        // if (sample_rate != self.sample_rate()) {}
-
-        let mut channels = Vec::<ChannelData>::with_capacity(number_of_channels);
+    /// The current implementation can decode FLAC, Opus, PCM, Vorbis, and Wav.
```
b-ma, line 87: Do you think it could support `mp3` as well?

orottier: Definitely, Symphonia supports quite a few more types: https://docs.rs/symphonia/0.4.0/symphonia/#codecs
The current ones are the 'default features', I'll add the mp3-feature too because it's nice to have batteries included. For the others we can also include feature gates as the added code size may not be beneficial for these lesser used codecs

examples/media_element.rs
orottier: That's a valid point.

To be fair I'm struggling a bit with the 'boundaries' of the Web Audio API also in other cases. For #77 I was playing around with microphone input and playback and realized I needed to have something like https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder

There are a few conflicting goals:
- stay as close as possible to the spec (as that is the unique selling point of this project, provide the comfort of not learning a new API when coming from a web background)
- leverage all the power of rust (no nullables, no exceptions, use ownership etc)
- create a small library (it's a lot of work already in its current state), versus:
- include related APIs (https://w3c.github.io/mediacapture-main/ https://w3c.github.io/webrtc-pc/ https://html.spec.whatwg.org/multipage/media.html#the-audio-element )
- or include include a few intefaces acting as building blocks (the `MediaStream` trait)

I will create a separate issue for this. It's probably best to steer away from names such as `MediaStream` and `MediaElement` because as you pointed out these are different from the w3c specs. I do think we should include them in the library because you would need them in many applications anyway

b-ma: My general impression is that the current state is not far from a really satisfying compromise. I will try to summarize more clearly my own impressions, and the deviations I have seen, in your new issue.

src/context.rs
b-ma: Cool!

examples/media_element.rs
orottier: Issue #84

PR #83, opened 2022-01-04
Housekeeping, consistency and thread safe nodes

orottier:
Some trivial stuff, except for the WaveShaperNode and AudioBufferSourceNode:
I noticed these were quite similar so I consolidated their setup. Now they don't need &mut refs so they are thread safe

### DIFF
--- a/src/node/audio_buffer_source.rs
+++ b/src/node/audio_buffer_source.rs
@@ -1,4 +1,6 @@
 use crossbeam_channel::{Receiver, Sender};
+use once_cell::sync::OnceCell;
+use std::sync::atomic::{AtomicBool, Ordering};
 
 use crate::buffer::AudioBuffer;
 use crate::context::{AsBaseAudioContext, AudioContextRegistration, AudioParamId};
@@ -57,8 +59,8 @@ struct AudioBufferMessage(AudioBuffer);
 /// let file = File::open("sample.wav").unwrap();
 /// let audio_buffer = context.decode_audio_data(file);
 /// // play the sound file
-/// let mut src = context.create_buffer_source();
-/// src.set_buffer(&audio_buffer);
+/// let src = context.create_buffer_source();
+/// src.set_buffer(audio_buffer);
 /// src.connect(&context.destination());
 /// src.start();
 /// ```
@@ -75,8 +77,8 @@ pub struct AudioBufferSourceNode {
     sender: Sender<AudioBufferMessage>,
     detune: AudioParam,        // has constraints, no a-rate
     playback_rate: AudioParam, // has constraints, no a-rate
-    buffer: Option<AudioBuffer>,
-    source_started: bool,
+    buffer: OnceCell<AudioBuffer>,
+    source_started: AtomicBool,
 }
 
 impl AudioNode for AudioBufferSourceNode {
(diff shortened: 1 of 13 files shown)


src/node/audio_buffer_source.rs
```diff
@@ -161,100 +152,94 @@ impl AudioBufferSourceNode {
...
-        self.buffer = Some(audio_buffer.clone());
-        // - If start() has previously been called on this node, perform the
-        // operation acquire the content on buffer.
-        // see <https://github.com/orottier/web-audio-api-rs/issues/68> for
-        // more information on the "acquire the content" concept in this context.
-        if self.source_started {
-            self.send_buffer();
-        }
+
+        self.sender
```
orottier, line 202: Notice I'm sending the buffer regardless of `self.source_started`. I think that is just fine, also regarding our discussion resulting in not needing acquire-the-content ops

orottier:
Also, from the last commit message:
> Some other changes:
> - set_buffer now takes an AudioBuffer instead of a reference. Taking a
>   references suggests we don't need ownership, but we clone it
>   internally so this gives off the wrong message
> - AudioBufferSourceNode internal_buffer now has capacity of MAX_CHANNELS
> - add missing `buffer()` (getter) method on AudioBufferSource
> - Both nodes take a comm channel of capacity 1. This is fine and removes
>   the need for WaveShaperNode::set_curve_mock (for tests)

src/node/audio_buffer_source.rs
```diff
@@ -75,8 +77,8 @@ pub struct AudioBufferSourceNode {
     sender: Sender<AudioBufferMessage>,
     detune: AudioParam,        // has constraints, no a-rate
     playback_rate: AudioParam, // has constraints, no a-rate
-    buffer: Option<AudioBuffer>,
-    source_started: bool,
+    buffer: OnceCell<AudioBuffer>,
+    source_started: AtomicBool,
```
b-ma, line 81: Don't understand why we need an `AtomicBool` here?

src/node/audio_buffer_source.rs
b-ma: Yup, looks fine and more simple

src/node/delay.rs
```diff
@@ -51,8 +51,8 @@ impl Default for DelayOptions {
 /// delay.delay_time().set_value(0.5);
 /// delay.connect(&context.destination());
 ///
-/// let mut src = context.create_buffer_source();
-/// src.set_buffer(&audio_buffer);
+/// let src = context.create_buffer_source();
+/// src.set_buffer(audio_buffer);
```
b-ma, line 55: removing all these `mut` is really nice from a user point of view I think

b-ma:
> set_buffer now takes an AudioBuffer instead of a reference. Taking a
references suggests we don't need ownership, but we clone it
internally so this gives off the wrong message

Yup, I agree it sounds more logical

> AudioBufferSourceNode internal_buffer now has capacity of MAX_CHANNELS

Sounds good too (the `2` was really arbitrary here), for consistency it should probably be the same in the `DelayNode` then

> add missing buffer() (getter) method on AudioBufferSource

Nice shot!

src/node/audio_buffer_source.rs
orottier: It's because the following function can now be used in a multithreaded case (takes an `&self` instead of `&mut self`)
```
    pub fn start_at_with_offset_and_duration(&self, start: f64, offset: f64, duration: f64) {
        if self.source_started.swap(true, Ordering::SeqCst) {
            panic!("InvalidStateError: Cannot call `start` twice");
        }
     ...
```

I can't change the bool value if I only have `&self`, but I can change an AtomicBool!
https://doc.rust-lang.org/std/sync/atomic/struct.AtomicBool.html#method.swap
As you can see the atomicbool is mutable without needing `&mut self` because it leverages atomic CPU instructions.

The reason I would like to have all AudioNode methods take `&self` instead of `&mut self` is, as you stated, user friendlyness (it "just works" if you call from different threads) and also resembles the spec better

b-ma: oh ok thanks! (I somehow need to update my mindset regarding threads :)

PR #85, opened 2022-01-05
Waveshaper - fix, tests and (hopefully) tail time

b-ma:
Hey,

This mainly fixes the waveshaper when there is no oversampling (it was basically a pass through in the current implementation), I also added few unit tests and updated the tail time (I don't think `rubato::FftFixedInOut` introduces any latency, at least it's not written in their documentation, but this should probably be confirmed)

I tried to add some unit tests for 2x and 4x resampling but rubato outputs garbage with arbitrary signals such as the one I use in the tests, I guess it comes from the fact this kind of signal is poorly described with an FFT but this should probably be confirmed too, however I'm really not sure neither that FFT can behave well with such small chunk of data (i.e. it's too short in time to have enough information to describe properly low to medium frequencies).


src/node/waveshaper.rs
```diff
@@ -369,11 +371,11 @@ impl WaveShaperRenderer {
         let wave_in = input.channels();
 
         let up_wave_in = self.upsampler_x2.process(wave_in).unwrap();
-        let mut up_wave_out = up_wave_in.clone();
+        let mut up_wave_out = up_wave_in.clone(); // this looks like allocation
```
orottier, line 374: You are correct. We're actually not using the cloned values anyway. We can probably insert a reusable buffer here

src/node/waveshaper.rs
```diff
@@ -349,18 +350,19 @@ impl WaveShaperRenderer {
     }
 
     #[inline]
-    fn no_process(&self, input: &AudioRenderQuantum, output: &mut AudioRenderQuantum) {
+    fn pass_through(&self, input: &AudioRenderQuantum, output: &mut AudioRenderQuantum) {
```
orottier, line 353: This could be written as `*output = input.clone();` I think. The clone does not allocate but increases the `Rc` strong count

src/node/waveshaper.rs
```diff
@@ -349,18 +350,19 @@ impl WaveShaperRenderer {
...
-                *o = i;
-            }
+            o_data.copy_from_slice(&i_data[..]);
         }
     }
 
     #[inline]
-    fn process_none(&self, input: &AudioRenderQuantum, output: &mut AudioRenderQuantum) {
+    fn process_1x(&self, input: &AudioRenderQuantum, output: &mut AudioRenderQuantum) {
+        // this does nothing...
```
orottier, line 361: What do you mean here? We're applying the shape on the input to produce output, no?

src/node/waveshaper.rs
```diff
@@ -349,18 +350,19 @@ impl WaveShaperRenderer {
...
         for (i_data, o_data) in input.channels().iter().zip(output.channels_mut()) {
```
orottier, line 362: Not related to your changes, but I think this is dangerous. We should not make assumption about the shape of the output buffers. They are typically reused every render quantum but they are initialized as a single channel of silence. This means if the input buffer has 2 channels, we're actually only processing 1 of them.

Probably best to use the strategy of the GainNode - make a copy of input and mutate in place
```rust
        // single input/output node
        let input = &inputs[0];
        let output = &mut outputs[0];
        let gain_values = params.get(&self.gain);

        *output = input.clone();

        output.modify_channels(|channel| {
            channel
                .iter_mut()
                .zip(gain_values.iter())
                .for_each(|(value, g)| *value *= g)
        });
```

Or do you think we should reshape the output in the render graph before handing them to the process function, to avoid these types of problems?

orottier:
> however I'm really not sure neither that FFT can behave well with such small chunk of data (i.e. it's too short in time to have enough information to describe properly low to medium frequencies).

Honestly, I don't know.. Time to do some studying

src/node/waveshaper.rs
b-ma: or maybe even mutate `up_wave_in ` in place.

I will try to have a look

src/node/waveshaper.rs
b-ma: Sorry, just a comment for myself I forgot to remove

src/node/waveshaper.rs
b-ma: Yes you are right, I will fix that

> Or do you think we should reshape the output in the render graph before handing them to the process function, to avoid these types of problems?

No strong opinion on that point, maybe it would be a bit more clear and avoid calls  to `output.set_number_of_channels`, but at contrary maybe that would require that the `Graph` have too much knowledge of the nodes (for example would it make sens for the `StereoPannerNode` ?).

b-ma:
From the formula here https://support.ircam.fr/docs/AudioSculpt/3.0/co/FFT%20Size.html , I think that means in our case that the frequency bins of the FFT for a 44.1kHz signal would be 22050 / 128 ~= 172 Hz wide, which is a really really poor frequency resolution, probably meaning the iFFT will just output blocks of noise...

I will fix the things you spotted, and just listen what we get (that's a good test I guess :), and we can also try to use https://docs.rs/rubato/latest/rubato/struct.SincFixedIn.html I suppose.

b-ma:
Hey, here are a couple of changes:

- handled reviews
- de-factorized inline method that was called only once
- removed cloning in 2x and 4x oversampling
- added doc

2 notes:
- I took the liberty to de-factorize the inline methods that were called only once because I find it more readable, but I guess that's really a matter of taste, let me know if you prefer that I revert back on this point
- I also tested the 2x and 4x oversampling in the example and it sounds quite good, so there is definitely something I don't understand here. I guess we can leave it like that for now and wait for some DSP expert expertise :)

orottier:
Changes look very good to me, thanks once again!

PR #87, opened 2022-01-10
Improvements for audio decoding: error handling, empty input, tests

orottier:
btw, the wav decoding is fixed. You can use `cargo update` to get the new version


b-ma:
Very nice! I have updated the decoding example and it works like a charm

One small thing is that the error messages are a bit confusing, nothing really important but maybe it would be better to override them with some generic message (at contrary it would possibly make further debugging more complicated, no strong opinion)
```
> --------------------------------
> Error decoding audio file: "samples/empty_2c.wav"
> IoError(Custom { kind: UnexpectedEof, error: "end of stream" })
> --------------------------------
> --------------------------------
> Error decoding audio file: "samples/corrupt.wav"
> IoError(Custom { kind: UnexpectedEof, error: "end of stream" })
> --------------------------------
> --------------------------------
> Error decoding audio file: "samples/sample.aiff"
> DecodeError("mp3: invalid mpeg audio layer")
> --------------------------------
> --------------------------------
> Error decoding audio file: "samples/sample.webm"
> DecodeError("mp3: invalid main_data_begin")
> --------------------------------
```

I will push the updated `decoding` example in a next PR along with the `waveshaper` example that somehow disappeared in the merge (I renamed the file, so I imagine the order of the merges didn't go well), in any case it must be reviewed because I just realized today that it's not the proper way of implementing an overdrive/distortion effect.

If you prefer that I push it in this PR, let me know

orottier:
Thanks for the comments. I'll open a new ticket for the error messages to pick up in a later stage. This may also require upstream improvements. I know the symphonia peeps are also cooking up some ideas around fatal/non-fatal exceptions so maybe best to wait for that.

Please open a new PR for the fixed waveshaper example :) it's not really my domain so any improvements you can make are very welcome!

PR #90, opened 2022-01-11
Improved `decoding` and `waveshaper` examples

b-ma:
nothing less, nothing more :)

2 notes:
- I still have this error ` ERROR web_audio_api::media::decoding] next packet err IoError(Custom { kind: UnexpectedEof, error: "end of stream" })` in the `waveshaper` example while I don't have it for the same file in `decoding`, any idea why?
- more problematic, I had a panic pointing to `graph::mark_free_when_finished` (something like `cannot unwrap() None`), when running the `waveshaper` example. Just once unfortunately, and before merging your last changes (but I don't think it could be related), I can't really figure out where could it come from.


orottier:
Good stuff. Nice to hear the distortion in effect

orottier:
In decoding.rs, the logger is not initialized so no logs are generated! It needs an `env_logger::init();` call
Let's keep an eye out for the mark_free_when_finished panics. Thinking about it, there could be a race condition which is easily fixable. PR incoming!

orottier:
Okay I celebrated prematurely, I don't see how a race condition could happen.
`can_free` first checks if the control thread has dropped the handle (that is the meaning of free_when_finished = true)
So a node will never be removed when the control node still  has a handle.
So theoretically, it would be impossible to panic on 
```
    pub fn mark_free_when_finished(&mut self, index: NodeIndex) {
        self.nodes.get_mut(&index).unwrap().free_when_finished = true;
    }
```

Interesting..

b-ma:
Yep that's a bit confusing...

b-ma:
Could it comes from the `AudioParam`s ? 

really not sure but: 
- as they also have an `AudioContextRegistration` they should send the `let message = ControlMessage::FreeWhenFinished { id: self.id.0 };` message (in `context.rs`)
- but they are also dropped manually from the render thread (in `graph.rs`, `nodes.retain(|_id, n| !n.outgoing_edges.iter().any(|e| e.other_id == *index));`

(in which case, maybe the weird point would be that we don't see the error more often...)

orottier:
Hah, you are right and it happens only in this very rare occasion:
C = control thread, R = render thread

1. C: A GainNode goes out of scope. It's fields are dropped in order of their definition [1], so the `registration` field is dropped before the `gain: AudioParam` field
2. C: `registration` is dropped which sends a `FreeWhenFinished` event for the GainNode
3. R: a render quantum starts, the event queue so far is read
4. C: the `gain` field is dropped, so a `FreeWhenFinished` event is sent which will be picked up in the next quantum
4. R: render quantum ends. If the gain node has no inputs left, it is removed from the graph. From the logic you mentioned, the audio param is removed too
5. R: next render quantum: the `FreeWhenFinished` event for the gain param is handled, but it was already removed!

Cool stuff.  The fix will be simple (just ignore the case where the node is missing)

[1] https://github.com/rust-lang/rfcs/blob/master/text/1857-stabilize-drop-order.md

PR #93, opened 2022-01-12
Fix possible race condition in node cleanup

### LINKED ISSUE #92: Possible panic in race condition in graph cleanup
C = control thread, R = render thread

1. C: A GainNode goes out of scope. It's fields are dropped in order of their definition [1], so the `registration` field is dropped before the `gain: AudioParam` field
2. C: `registration` is dropped which sends a `FreeWhenFinished` event for the Gain Processor
3. R: a render quantum starts, the event queue so far is read
4. C: the `gain` field is dropped, so a `FreeWhenFinished` event is sent for the Audio Param processor, which will be picked up in the next quantum
4. R: render quantum ends. If the gain node has no inputs left, it is removed from the graph. From the logic you mentioned, the audio param is removed too
5. R: next render quantum: the `FreeWhenFinished` event for the gain param is handled, but it was already removed!

Cool stuff.  The fix will be simple (just ignore the case where the node is missing)

[1] https://github.com/rust-lang/rfcs/blob/master/text/1857-stabilize-drop-order.md

_Originally posted by @orottier in https://github.com/orottier/web-audio-api-rs/pull/90#issuecomment-1011248548_

orottier:
Fixed #92

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /target
 Cargo.lock
 /.vscode
+/.idea
 *.svg
 *.data
 *.data.old
--- a/src/render/graph.rs
+++ b/src/render/graph.rs
@@ -170,7 +170,12 @@ impl Graph {
     }
 
     pub fn mark_free_when_finished(&mut self, index: NodeIndex) {
-        self.nodes.get_mut(&index).unwrap().free_when_finished = true;
+        // Issue #92, a race condition can occur for AudioParams. They may have already been
+        // removed from the audio graph if the node they feed into was dropped.
+        // Therefore, do not assume this node still exists:
+        if let Some(node) = self.nodes.get_mut(&index) {
+            node.free_when_finished = true;
+        }
     }
 
     /// Helper function for `order_nodes` - traverse node and outgoing edges


b-ma:
Cool ! Looks like both a good explanation and solution

However, I'm still a bit confused on one point (but maybe/probably it doesn't change anything on how the thing works, and would only make things more complicated). But just to write it out:
- the `mark_free_when_finished` doesn't make sens for `AudioParam`s as they are (and should) never (be) finished until their related `AudioNode` is dropped
- therefore, would it make sens (or be possible) to have a kind of special implementation of `ContextRegistration::Drop` that wouldn't try to `mark_free_when_finished` when the `AudioParam` is dropped (as we know it will be dropped with the `AudioNode` itself)?

Anyway, beside aesthetic problems, my main concern with this solution is that it could possibly hide another strange unexpected edge-case because nothing can break anymore, so just in case I think it would be nice to just throw some hardcore `panic!` on the `else` side.

orottier:
I see your point, but it would require the `AudioContextRegistration` to know if they belong to an actual AudioNode or an AudioParam. Therefore we would need to add a boolean `is_param` or something. To fill that value, the BaseAudioContext::register function needs to be changed to pass that info. This function is user facing (as you need it for custom AudioWorklet implementations), and that would mean the user could set it erroneously. To prevent that, we would need to separate register into register_audio_node and register_audio_param but I think we are overcomplicating stuff now

Not sure if there is a more elegant solution

b-ma:
Ok, that was my impression but I preferred to have your confirmation, no need to overcomplicate things I agree.

(...and ok... I realize my `panic!` idea was really stupid because it's exactly what the patch is supposed to fix... (so, let's forget that with the head high :))

So, all good!

orottier:
Alright!

PR #94, opened 2022-01-13
Allow method chaining on AudioParam

orottier:
strictly speaking, `set_value` is not part of the spec (since value should be an attribute, but rust does not allow us to override attribute mutation) and therefore should not allow chaining. But I think it is more consistent this way


b-ma:
Yup I agree, the inverse would be really weird. 

I also prefer this solution rather than the `Result<Self, Err>` one

Issue #75, opened 2021-12-29
Decide how to deal with fallible operations (e.g. exceptions thrown)


orottier:
In the spec, exceptions are thrown all over the place (when creating nodes, when updating attributes, when calling methods..)

There a few options for the rust version:

- panic on exception (this is what we do currently)
- all fallible operations must return a `Result<..>`
- all operations must return a `Result<..>` (for consistency, even infallible ones)
- panic on exception but offer a `try_...` method for fallible operations returning a `Result`
- ignore errors (log them)

All variants have pros and cons. We need to strike a balance between correctness and pragmatism.

e.g. https://docs.rs/curl/0.4.41/curl/ returns a Result for every operation meaning your code will be littered with `unwrap`s

The rust stdlib strikes a reasonable balance between (well documented) panics and offering `try_...` functions, in my opinion (e.g. https://doc.rust-lang.org/std/cell/struct.RefCell.html#method.try_borrow )

b-ma:
Tricky one... another point to take into account maybe is how often the fallible methods return a value you actually want to use use. For example, with params most of the time you will write something like:

```rs
my_param.set_value(0., 0.);
my_param.linear_ramp_to_value_at_time(1., 1.);
my_param.exponential_ramp_to_value_at_time(0.0001, 1.);
```

If we followed the spec here (returning `self` for method chaining) while handling `Error`s using `Result`, (if I'm not wrong here...) we could allow writing something like the following while still allowing the first snippet :

```rs
my_param
   .set_value(0., 0.)?
   .linear_ramp_to_value_at_time(1., 1.)?
   .my_param.exponential_ramp_to_value_at_time(0.0001, 1.)?;
```

I don't say that's the best solution at all (and I'm personally really not for the `unwrap()` everywhere solution), but would maybe be interesting to have a clear picture of all the actual fallible operations before having a decision that may have a lot of impact on the interface.

orottier:
Current user facing panics:
- AudioParam: exponential_ramp_to_value_at_time, set_target_at_time (for bad numerical inputs)
- AudioBuffer: from (if channel have unequal len)
- AudioBufferSource (when setting buffer more than once, when starting/stopping more than once)
- WaveShaper (when setting curve more than once)
- AudioNode, when connecting nodes from different contexts. Note: connect_at returns a Result instead of panicing when ports are not available.
- DelayNode: when delay time is <0 or >180s
- ChannelSplitter/MergerNode: when overriding channel config

[...]

PR #95, opened 2022-01-13
AudioNode connect_at function will panic instead of return an Err

orottier:
conform discussion at #75


orottier:
Hah, new rust version with additional clippy checks. Will have a look soon

b-ma:
All good! 

(I already miss the `wave_whaper` a bit but ok... :)

Issue #61, opened 2021-12-02
Support sub-quantum scheduling


orottier:
The following processors only support scheduling per `BUFFER_SIZE`d quantum, this should be improved:

- [x] ConstantSourceRenderer
- [x] AudioBufferSourceRenderer
- [x] OscillatorRenderer
- [x] DelayRenderer
- ~MediaStreamRenderer~ (this does not make much sense)

e.g. in the current implementation the `timestamp` (start of the quantum) is evaluated

```rust
// todo, sub-quantum start/stop
match self.scheduler.state(timestamp) {
    ScheduledState::Active => (),
    ScheduledState::NotStarted => {
        output.make_silent();
        return true; // will output in the future
    }
    ScheduledState::Ended => {
        output.make_silent();
        return false; // can clean up
    }
}
```

b-ma:
Hey, just to let you know I'm fiddling around with the `DelayNode` sub-sample rendering

b-ma:
Hey, looks like this one is actually done (as I don't think `MediaStreamRenderer` should appear in this list), what do you think?

orottier:
Correct!

PR #96, opened 2022-01-15
Oscillator sub-sample scheduling

b-ma:
Related to #61

A bit more complicated than I thought here

also:
- some fixes and tests on `AudioBufferSourceNode` and `ConstantSourceNode`
- a new example `amplitude_modulation`

Let me know what you think


orottier:
Impressive work, again. I will need more time to look at it carefully.
It is very satisfying to see this library maturing and improving week over week!

src/node/oscillator.rs
```diff
@@ -679,555 +325,179 @@ impl AudioProcessor for OscillatorRenderer {
...
+    // computes the `polyBLEP` corrections to apply to aliasing signal
+    // `polyBLEP` stands for `polyBandLimitedstEP`
+    // This basically soften the sharp edges in square and sawtooth signals
+    // to avoid infinite frequencies impulses (jumps from -1 to 1 or inverse).
+    // cf. http://www.martin-finke.de/blog/articles/audio-plugins-018-polyblep-oscillator/
+    //
+    // @note: do not apply in tests so we can avoid relying on snapshots
+    #[inline]
+    fn poly_blep(mut t: f64, dt: f64) -> f64 {
+        if cfg!(test) {
```
orottier, line 498: I think it's good that we drop the 'snapshot' method of testing. Computing the expected values on the fly is a stronger guarantee for correctness than a big blob of numbers.
However, how do we now test the poly_bleb functionality?

b-ma: Hey, I changed a bit the `poly_blep` API to allow testing and added a simple isolated unit test. I'm not sure we should go further with testing this though, as it could be changed or removed later. 
Maybe more interesting would be have some helpers scripts to plot a piece of signal and/or its frequency response from a snapshot. I have already made a small python script to plot a signal from a snapshot (really nothing complicated but nice to have at hand), do you thing I should add it to the repo or maybe it would be better to share such helpers as gists or something as they are really not stable things?

src/node/audio_buffer_source.rs
```diff
@@ -413,6 +413,14 @@ impl AudioProcessor for AudioBufferSourceRenderer {
         // @see <https://webaudio.github.io/web-audio-api/#playback-AudioBufferSourceNode>
         let mut current_time = timestamp;
 
+        // prevent scheduling in the past
+        // If 0 is passed in for this value or if the value is less than
+        // currentTime, then the sound will start playing immediately
+        // cf. https://webaudio.github.io/web-audio-api/#dom-audioscheduledsourcenode-start-when-when
+        if !self.render_state.started && start_time < current_time {
```
orottier, line 420: The spec also says `A RangeError exception MUST be thrown if when is negative.` but I think this is fine.

src/periodic_wave.rs
```diff
@@ -0,0 +1,423 @@
...
+            Self::normalize(&mut wavetable);
+        }
+
+        wavetable
+    }
+
+    fn normalize(wavetable: &mut [f32]) {
+        let mut max = 0.;
+
+        for sample in wavetable.iter() {
```
orottier, line 205: Have a look at https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.max
The previous version used a version with `reduce`, but that is also overly complicated

src/node/oscillator.rs
```diff
@@ -292,7 +124,7 @@ pub struct OscillatorNode {
     /// starts and stops Oscillator audio streams
     scheduler: Scheduler,
     /// channel between control and renderer parts (sender part)
-    sender: Sender<OscMsg>,
+    sender: Sender<OscillatorMessage>,
```
orottier, line 127: Not really related to your work, but I don't see why this is not a `Sender<PeriodicWave>`

src/node/oscillator.rs
```diff
@@ -406,242 +225,77 @@ impl OscillatorNode {
...
-                computed_freq,
-                dyn_wavetable,
-                norm_factor,
-                disable_normalization,
-            })
+            .send(OscillatorMessage(periodic_wave))
             .expect("Sending periodic wave to the node renderer failed");
     }
-
-    /// Generate the wavetable
```
orottier, line 547: Great to see you have decoupled the OscNode from the PeriodicWave stuff. Much more succinct this way

src/node/oscillator.rs
```diff
@@ -679,555 +325,179 @@ impl AudioProcessor for OscillatorRenderer {
...
-        if t < dt {
+    // computes the `polyBLEP` corrections to apply to aliasing signal
+    // `polyBLEP` stands for `polyBandLimitedstEP`
+    // This basically soften the sharp edges in square and sawtooth signals
+    // to avoid infinite frequencies impulses (jumps from -1 to 1 or inverse).
+    // cf. http://www.martin-finke.de/blog/articles/audio-plugins-018-polyblep-oscillator/
+    //
+    // @note: do not apply in tests so we can avoid relying on snapshots
+    #[inline]
+    fn poly_blep(mut t: f64, dt: f64, is_test: bool) -> f64 {
```
orottier, line 497: This works but it is a bit uncommon I think, to pass a `test` boolean. I'll try to suggest something different later

orottier (commented):
All in all very nice. I need some more time for testing so maybe I'll add more comments later!

src/node/audio_buffer_source.rs
b-ma: Good catch, but I guess this should rather be checked in `start_at(when)`. 

This check will need to be done anyway: I had some very rare crash in the `many_oscillators_with_env` because `start_time` was actually lower than `current_time` (coming from some thread concurrency I guess) which completely spoiled the `phase` calculation (...that took a while to debug because I had to wait for 5 to 10mn each time... :).

src/periodic_wave.rs
b-ma: Yep, but as far I understand the previous implementation didn't check for absolute values, which is important here because if your signal is `[-0.5, 0.1]` you want your normalization factor to be `2.`, not `10.`. Maybe it could be written in a more functional style though.

src/node/oscillator.rs
b-ma: Yup you are right, I will update that

src/node/oscillator.rs
b-ma: Yup, I agree, actually I started by just doing:

```rs
#[cfg(test)]
fn poly_blep_test() {
   if t < dt {
     t /= dt;
     t + t - t * t - 1.0
   } else if t > 1.0 - dt {
     0.0
   }
}
```

but I finally tough it would be more error prone

src/node/oscillator.rs
b-ma: Yup, but I have to say that this implementation might be a bit less efficient than the previous one, I don't really know. But there was several issues with the previous implementation (the `real` and `imaginary` components were somehow inverted, and there was a small offset in the phase), so I though it would be better to just stick to the mathematical formula, write the tests against it, so a next iteration for perf improvement (e.g. iFFT or table lookup) would be easier and accurate.

orottier: Correctness is most important obviously. We can always measure and optimize later

src/node/oscillator.rs
orottier: Feel free to add some files to a `scripts` folder or something. We can always drop them later

src/periodic_wave.rs
orottier: Yup, like `wavetable.iter().map(|&v| v.abs()).max().unwrap()`

src/node/oscillator.rs
orottier: Let's leave it like this then!

src/periodic_wave.rs
```diff
@@ -170,8 +170,9 @@ impl PeriodicWave {
     // cf. https://webaudio.github.io/web-audio-api/#waveform-generation
     // note that sines are in the imaginary components
     //
-    // the current implementation is naive and could be improved using inverse FFT
-    // or table lookup on SINETABLE
+    // @note: Current implementation is naive and could be improved using inverse
+    // FFT or table lookup on SINETABLE. Such performance improvements should be
+    // however tested also against this implementation.
```
orottier, line 175: True, but since this does not occur on the render thread it is really fine for now

orottier:
Alright, I'm ready to merge. Or do you plan on adding anything else?

b-ma:
Ok cool, go on! its good on my side, the possible improvements are quite clear and isolated I think, so we can handle them later

PR #97, opened 2022-01-16
Microphone playback example - featuring a terminal UI

orottier:
The code is a hot mess but I had a lot of fun building it. It takes a 4 sec mic recording and then plays it back with effect. Also a frequency diagram is shown.

I hope this gives a few reusable UI components that can be used to build simple examples such as keyboard playing, beat tapping, metronome display and some rudimentary graphing.

Work in progress for now. I need to work on code cleanup and I have a fix in mind for the 100% CPU usages on the UI thread.


b-ma:
Cool, I will try to have a look soon

orottier:
Cool, thanks. The effects are just different filters settings of the same biquad node but maybe I will make it a bit more complex. I'll add some more work, like 'press-and-hold-space-to-record' and the CPU usage thing.
The termion `MouseEvent` look simple to use but to actually build sliders may be tricky, good luck! Let me know if I can help anywhere.

orottier:
A bit more high level text UI framework is https://crates.io/crates/cursive
It has a slider component available

b-ma:
> A bit more high level text UI framework is https://crates.io/crates/cursive
> It has a slider component available

Cool thanks, I will have a look, I spent enough time implementing sliders in JS :), don't really want to spend hours on that

Issue #76, opened 2021-12-29
Check current API surface


orottier:
API surface: `cargo doc --lib --open`

Check for

- [x] consistency
- [x] internals accidentally marked `pub`
- [x] commit to the current `mod` layout?
- [x] commit to the location of the `Error`s and constants?
- [x] happy with the `BaseAudioContext` and `AsBaseAudioContext`?
- [x] `AudioContext::new(AudioContextOptions)` (not wrapped in an Option)
- [x] remove `Scheduler` and `Controller` from public interface, change traits (AudioScheduledSourceNode etc)
- [x] Check all node constructor options
- [x] Check all node methods
- [x] remove public spatial mod
- [x] mark functions that should return a Promise but currently don't `_sync`
- [x] rename `AudioContextOptions.channels -> number_of_channels`
- [x] create_delay and create_channel_(splitter/merger) take optional argument, but we make them required. Same as create_periodic_wave but here we take in an opt struct. 
- [x] Can we drop  AudioNode's  `channel_config_raw` and `channel_config_cloned`?
- [x] Consistency in AudioNode optional arg: `disconnect() , disconnect_from(node)`
- [x] Remove AudioNodeId from public interface
- [x] stream node constructors only on online AudioContext

b-ma:
> commit to the current mod layout?

- My very personal taste from a user point of view would be to have everything flat. As the public API is not that huge and there is possibility of naming collision (in JS everything is basically exposed to `window`), the namespacing feels more like a burden people need to learn than something that helps. 
- However, I also think that it would be nice to have some `extra` (or whatever) namespace that would contain all structs that are not _per se_ part of the API, but are really nice to have at hand (e.g. `Microphone`, `MediaElement`, etc.) and without which some nodes of the API do not make much sense.

> happy with the BaseAudioContext and AsBaseAudioContext?

Not really :), I think the `AsBaseAudioContext` is really weird from a user point of view and that it would be far more simple to understand renamed to `BaseAudioContext`, e.g.: I can understand that I have to `import::context::{AudioContext, BaseAudioContext}`, but I really don't get why i should `import::context::{AudioContext, AsBaseAudioContext}` which is not even described in the spec. 

From an implementation perspective `AsBaseAudioContext` really makes sense, but from a user perspective it really does not.

b-ma:
[...]

PR #99, opened 2022-01-22
proof of concept: delegation.

orottier:
The point of this exploration is to see if we can hide some of the internals of the (base)audiocontext implementation from the user facing code as discussed in #76 
Ideally AsBaseAudioContext should go, as it is not part of the official spec and might surprise users that they have to import that trait. I'm not entirely sure if we can drop it though as it is used in the generics setup. We should be able to drop it from user facing code though
The BaseAudioContext must stay as it is part of the spec

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ float_eq = "0.7"
 smallvec = "1.7.0"
 once_cell = "1.9.0"
 symphonia = { version = "0.4.0", default-features = false }
+delegate-attr = {git = "https://github.com/upsuper/delegate-attr.git", rev ="32cda21a40f6e8a1b069d6178455318d8ecd6803" }
 
 [dev-dependencies]
 rand = "0.8.*"
--- a/examples/amplitude_modulation.rs
+++ b/examples/amplitude_modulation.rs
@@ -1,5 +1,5 @@
 use std::{thread, time};
-use web_audio_api::context::{AsBaseAudioContext, AudioContext};
+use web_audio_api::context::{Context, AudioContext};
 use web_audio_api::node::{AudioNode, AudioScheduledSourceNode};
 
 fn main() {
--- a/examples/amplitude_modulation.rs
+++ b/examples/amplitude_modulation.rs
@@ -1,5 +1,5 @@
 use std::{thread, time};
-use web_audio_api::context::{Context, AudioContext};
+use web_audio_api::context::{AudioContext, Context};
 use web_audio_api::node::{AudioNode, AudioScheduledSourceNode};
 
 fn main() {
--- a/src/node/delay.rs
+++ b/src/node/delay.rs
@@ -127,7 +127,8 @@ impl AudioNode for DelayNode {
             panic!("IndexSizeError: input port {} is out of bounds", input);
         }
 
-        self.registration().connect(self.reader_registration.id(), dest.id(), output, input);
+        self.registration()
+            .connect(self.reader_registration.id(), dest.id(), output, input);
 
         dest
     }
@@ -146,7 +147,8 @@ impl AudioNode for DelayNode {
 
     /// Disconnects all outgoing connections from the AudioNode.
     fn disconnect_all(&self) {
-        self.registration().disconnect_all(self.reader_registration.id());
+        self.registration()
+            .disconnect_all(self.reader_registration.id());
     }
 }
 
--- a/src/node/delay.rs
+++ b/src/node/delay.rs
@@ -1,4 +1,4 @@
-use crate::context::{AsBaseAudioContext, AudioContextRegistration, AudioParamId};
+use crate::context::{AudioContextRegistration, AudioParamId, Context};
 use crate::param::{AudioParam, AudioParamOptions};
 use crate::render::{AudioParamValues, AudioProcessor, AudioRenderQuantum};
 use crate::{SampleRate, RENDER_QUANTUM_SIZE};
@@ -32,13 +32,13 @@ impl Default for DelayOptions {
 ///
 /// - MDN documentation: <https://developer.mozilla.org/en-US/docs/Web/API/DelayNode>
 /// - specification: <https://webaudio.github.io/web-audio-api/#DelayNode>
-/// - see also: [`AsBaseAudioContext::create_delay`](crate::context::AsBaseAudioContext::create_delay)
+/// - see also: [`Context::create_delay`](crate::context::Context::create_delay)
 ///
 /// # Usage
 ///
 /// ```no_run
 /// use std::fs::File;
-/// use web_audio_api::context::{AsBaseAudioContext, AudioContext};
+/// use web_audio_api::context::{Context, AudioContext};
 /// use web_audio_api::node::AudioNode;
 ///
 /// // create an `AudioContext` and load a sound file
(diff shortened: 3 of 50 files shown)


examples/amplitude_modulation.rs
```diff
@@ -1,5 +1,5 @@
 use std::{thread, time};
-use web_audio_api::context::{AsBaseAudioContext, AudioContext};
```
orottier, line 2: Simply renaming `AsBaseAudioContext` to `Context` was not the solution we were aiming for.
Can you make this example compile without needing to import the trait? That is the goal! We want the user facing code to not have imports that are not part of the spec.

In any case, please have a look at the failing CI checks

src/node/delay.rs
```diff
@@ -117,7 +117,7 @@ impl AudioNode for DelayNode {
         output: u32,
         input: u32,
     ) -> &'a dyn AudioNode {
-        if self.context() != dest.context() {
+        if self.registration() != dest.registration() {
```
orottier, line 120: This is very incorrect

orottier: I stand corrected, it is very misleading.  I don't think you should override the equals method on this  struct

orottier:
@tasogare3710 my suggestion for moving forwards would be:

- scratch the current work and start fresh
- start by looking at 1 example, e.g. `example/simple_delay.rs`
- remove the import of `AsBaseAudioContext`
- obviously it won't compile now. Add the required methods to `AudioContext` with the delegate trait. For example delegate `context.create_delay` to the `base()`
- now you can remove the `create_delay` method from `AsBaseAudioContext`
- rinse, repeat, for more methods and examples and apply to `OfflineAudioContext` too
- check if the docs look good for `AudioContext`, `OfflineAudioContext` (`cargo doc --lib --open`)
- check if there is any user facing code still containing `AsBaseAudioContext` (I mean the examples, integration tests and docs)

Cargo.toml
```diff
@@ -24,6 +24,7 @@ float_eq = "0.7"
 smallvec = "1.7.0"
 once_cell = "1.9.0"
 symphonia = { version = "0.4.0", default-features = false }
+delegate-attr = {git = "https://github.com/upsuper/delegate-attr.git", rev ="32cda21a40f6e8a1b069d6178455318d8ecd6803" }
```
orottier, line 27: Why not just `delegate-attr = "0.2.9"`
I'd rather not take in dependencies that are not from crates.io

examples/amplitude_modulation.rs
tasogare3710: In my opinion, AsBaseAudioContext(Context) is used for trait bounds, it would be difficult to remove the traits.
However, you can change it to an enum type, but you still need to import a type that is not in the spec.

https://github.com/tasogare3710/web-audio-api-rs/blob/a163e229bb8ab1c41dd4cc49972ad823ab911029/examples/worklet.rs#L44
etc.

Cargo.toml
tasogare3710: It's a temporary one that I copied and pasted from another project of mine, I was going to remove the delegate-attr when I sent the PR since it's not needed if you don't want to hide BaseAudioContext from the crate users.

src/node/delay.rs
tasogare3710: This is not necessary if you don't want to hide BaseAudioContext. I can revert to context().

examples/amplitude_modulation.rs
tasogare3710: Also, only difference between the Context trait and the spec's BaseAudioContext is the name. If you call the Context trait the BaseAudioContext trait, you can call the BaseAudioContext struct the ConcreteBaseAudioContext struct. But, not Rust-ish.

orottier:
Thanks for the comments, it made me realize any choice we make will have disadvantages..

Summarizing the three options at the table

1. structs `AudioContext`, `OfflineAudioContext` and `BaseAudioContext`.
trait `AsBaseAudioContext` for shared functionality and mimicking the inheritance.
This is the current state.
Advantages: rust-like
Disadvantages: users need to import `AsBaseAudioContext` most of the time, which is not part of the spec

1. structs `AudioContext`, `OfflineAudioContext` and `ConcreteBaseAudioContext`. 
trait `BaseAudioContext` for shared functionality and mimicking the inheritance.
Advantages: no need for strange imports in user facing code, still powerful and not too bad from a rust perspective
Disadvantages: whenever you need to deal with the actual base (this is rare), the struct name may be surprising

1. structs `AudioContext`, `OfflineAudioContext` and `BaseAudioContext`.
no traits. We use delegation/macros to make all shared functionality available on the three structs
Advantages: no names that are not part of spec
Disadvantages: not rust-like. No possibility for function to be generic over a shared trait (we use that in Node constructors now) but we could avoid that issue by forcing to take in a `BaseAudioContext` always

For context, we are using option 2 for `AudioNode` and `AudioScheduledSourceNode` which I think works out well. There is no issue there because they are never used as concrete types.

All in all I think I am in favour of option 2 now - we model inheritance using traits, always.
We could even make the namespace like this:
`crate::context::{AudioContext, OfflineAudioContext, BaseAudioContext, concrete::BaseAudioContext}`

@b-ma any final thoughts?

b-ma:
Hey,

Yup I agree, option 2. seems both the more clean and simple solution. (Then, we will also have to fix the discrepancy of the `AudioScheduledSourceNode` to make it coherent with this rule too, I will have a look)

I dont know for the namespacing `concrete::BaseAudioContext`, it seems quite good but also a bit strange alongside the `BaseAudioContextInner`. 
- Wouldn't it be possible to somehow merge `BaseAudioContextInner` and `ConcreteBaseAudioContext` (I don't really have a clear picture of the possible implications)? 
- If not possible or just complicating things, maybe doing something like `concrete::{BaseAudioContext, BaseAudioContextInner}` so that these two stay coherent?

orottier:
Fortunately `BaseAudioContextInner` is an implementation detail and is not exposed in the public interface.

The reason for `BaseAudioContext { inner: Arc<BaseAudioContextInner> }` is that we clone it many times. All the AudioNodes have static lifetime (meaning, they do not contain references, only owned values) which is great for user friendlyness. The nodes contain an `AudioContextRegistration` which has reference to the base context via that `Arc`

We could rename `BaseAudioContextInner --> BaseAudioFields` if that makes it easier for the eyes?

b-ma:
Yup I see, thanks for the explanation, looks indeed logical to keep this simple

> We could rename BaseAudioContextInner --> BaseAudioFields if that makes it easier for the eyes?

No, `Inner` is good I think and already used here and there if I remember well.

However, I would go for `ConcreteAudioContextInner`, so we have:
- `trait BaseAudioContext` (which retrieve) 
- `ConcreteBaseAudioContext` (which holds) 
- `ConcreteBaseAudioContextInner` 
(not sure of my keyword relationships but you see the idea :)

orottier:
> (Then, we will also have to fix the discrepancy of the AudioScheduledSourceNode to make it coherent with this rule too, I will have a look)

Yes I think we can make that one follow the spec better. Right now the trait exposes the implementation detail of `Scheduler` and then derives the methods from the spec (start, stop etc).

We should change that to just the raw trait, require implementers to supply `start(_at)`, `stop` etc
Internally we can use the `Scheduler`, but we should remove that from the public API (the whole `src/control.rs` file actually)

Was that what you had in mind too @b-ma  ?

b-ma:
> We should change that to just the raw trait, require implementers to supply start(_at), stop etc
Internally we can use the Scheduler, but we should remove that from the public API (the whole src/control.rs file actually)

Yup, I think it looks really more safe to go that path

orottier:
Thanks for moving the discussion forward @tasogare3710 
I'm closing this PR in favour of #102

tasogare3710:
@orottier 

> Thanks for moving the discussion forward @tasogare3710
> I'm closing this PR in favour of #102

Got it. it's my pleasure.

PR #102, opened 2022-01-23
Rename AsBaseAudioContext -> BaseAudioContext

orottier:
and rename the struct to ConcreteBaseAudioContext. This makes user
facing code a bit nicer, you will no longer need to import an item that
is not part of the spec.

See discussion at https://github.com/orottier/web-audio-api-rs/pull/99

### DIFF
--- a/examples/amplitude_modulation.rs
+++ b/examples/amplitude_modulation.rs
@@ -1,5 +1,5 @@
 use std::{thread, time};
-use web_audio_api::context::{AsBaseAudioContext, AudioContext};
+use web_audio_api::context::{AudioContext, BaseAudioContext};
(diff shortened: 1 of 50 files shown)


b-ma:
Just repeating my comment from #99 here, so it's more simple to follow:

> However, I would go for `ConcreteBaseAudioContextInner`, so we have:
> - `trait BaseAudioContext` (which retrieve) 
> - `ConcreteBaseAudioContext` (which holds) 
> - `ConcreteBaseAudioContextInner` 
> (not sure of my keyword relationships but you see the idea :)

Not very important, but I think it could help understand the relations between the abstractions when you are not completely used to the code.

Except that detail, looks very good I think!

orottier:
Good point, I have renamed it

Issue #73, opened 2021-12-29
Finalize all node constructor structs (e.g. ConstantSourceOptions) 


orottier:
For the v1.0 release, structs with public fields must be finalized since altering them would mean a breaking API change.

- [ ] decide on types (f64 for all float values?)
- [ ] add all fields to all constructor structs, even if the implementation is not yet using them
- [ ] add all enum variants (e.g. PanningModelType) even if the implementation is not yet using them
- [ ] how to handle optional fields?
- [ ] add a Default, Clone and Debug implementation on all structs

b-ma:
(will try to put my 2 cents in each issues)

For the type, my impression from the spec is that all that is related to time is `f64` (even if internally casted to f32, e.g. `delayTime`, and this is important to keep timing precision in long running applications), while every other values are more `f32` (and as the difference between `double` and `float` is already made, it is quite simple to follow without interpreting to much there). For count / enumerating attributes, (e.g. `channelCount`, `number_of_channels`), which are generally described as `unsigned long` I guess `usize` is generally the right one choice as these params / arguments are often used as indexes, etc.

orottier:
This was fixed in #103 
Some leftover tasks are listed in #76

PR #103, opened 2022-01-24
Attempt to harmonize options, defaults, etc

b-ma:
Hey, here is an attempt related to #73 

I tried to follow these rules:

- follow types given by the spec (few changes from `f32` to `f64`, and `usize` to `u32`)
- optional value -> defined in `Default`
- nullable value -> `Option` that defaults to `None`
- `option` in node constructor should always be given e.g.: 
```rs
OscillatorNode::new(context, Default::default());
// rather than
OscillatorNode::new(context, None);
```

I have the impression it's quite consistent and logical from a user perspective, but there is still an issue with the `required` parameters (e.g. for `IirFilter` or `AudioBuffer`), for which I see two possible solutions:

1. define required param as `Option` that defaults to `None` and `panic` in the constructor if `None`
2. do not provide `Default` whenever there is a required param

(for now its closer to 2. I guess, but only because I did nothing :)

I also cleaned a bit `context.rs` but mainly harmonizing factory methods and sorting them to alphabetical order.

Let me know what you think

### DIFF
--- a/src/node/media_element.rs
+++ b/src/node/media_element.rs
@@ -9,10 +9,14 @@ use super::{
     ChannelConfigOptions, MediaStreamRenderer,
 };
 
-/// Options for constructing a MediaElementAudioSourceNode
+/// Options for constructing a [`MediaElementAudioSourceNode`]
+// dictionary MediaElementAudioSourceOptions {
+//   required HTMLMediaElement mediaElement;
+// };
+// note that `MediaElementAudioSourceOptions` does not extend `AudioNodeOptions`
 pub struct MediaElementAudioSourceNodeOptions {
     pub media: MediaElement,
-    pub channel_config: ChannelConfigOptions,
+    pub channel_config: ChannelConfigOptions, // should not have this
 }
 
 /// An audio source from a [`MediaElement`] (e.g. .ogg, .wav, .mp3 files)
@@ -42,6 +46,7 @@ impl AudioNode for MediaElementAudioSourceNode {
     fn registration(&self) -> &AudioContextRegistration {
         &self.registration
     }
+
     fn channel_config_raw(&self) -> &ChannelConfig {
         &self.channel_config
     }
@@ -49,6 +54,7 @@ impl AudioNode for MediaElementAudioSourceNode {
     fn number_of_inputs(&self) -> u32 {
         0
     }
+
     fn number_of_outputs(&self) -> u32 {
         1
     }
--- a/src/node/panner.rs
+++ b/src/node/panner.rs
@@ -4,34 +4,76 @@ use std::sync::Arc;
 use crate::context::{AsBaseAudioContext, AudioContextRegistration, AudioParamId};
 use crate::param::{AudioParam, AudioParamOptions};
 use crate::render::{AudioParamValues, AudioProcessor, AudioRenderQuantum};
-use crate::{AtomicF32, SampleRate};
+use crate::{AtomicF64, SampleRate};
 
 use super::{
     AudioNode, ChannelConfig, ChannelConfigOptions, ChannelCountMode, ChannelInterpretation,
 };
 
-/// Options for constructing a PannerNode
+#[derive(Copy, Clone, Debug)]
+pub enum PanningModelType {
+    EqualPower,
+    HRTF,
+}
+
+#[derive(Copy, Clone, Debug)]
+pub enum DistanceModelType {
+    Linear,
+    Inverse,
+    Exponential,
+}
+
+/// Options for constructing a [`PannerNode`]
+// dictionary PannerOptions : AudioNodeOptions {
+//   PanningModelType panningModel = "equalpower";
+//   DistanceModelType distanceModel = "inverse";
+//   float positionX = 0;
+//   float positionY = 0;
+//   float positionZ = 0;
+//   float orientationX = 1;
+//   float orientationY = 0;
+//   float orientationZ = 0;
+//   double refDistance = 1;
+//   double maxDistance = 10000;
+//   double rolloffFactor = 1;
+//   double coneInnerAngle = 360;
+//   double coneOuterAngle = 360;
+//   double coneOuterGain = 0;
+// };
+#[derive(Clone, Debug)]
 pub struct PannerOptions {
+    #[allow(dead_code)]
+    pub panning_model: PanningModelType,
+    #[allow(dead_code)]
+    pub distance_model: DistanceModelType,
     pub position_x: f32,
     pub position_y: f32,
     pub position_z: f32,
     pub orientation_x: f32,
     pub orientation_y: f32,
     pub orientation_z: f32,
-    pub cone_inner_angle: f32,
-    pub cone_outer_angle: f32,
-    pub cone_outer_gain: f32,
+    #[allow(dead_code)]
+    pub ref_distance: f64,
+    #[allow(dead_code)]
+    pub max_distance: f64,
+    pub cone_inner_angle: f64,
+    pub cone_outer_angle: f64,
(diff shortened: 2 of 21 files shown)


orottier:
This is perfect. I agree that this is the best way to model the optional parameters.

Let's change the constructor of AudioContext too!

Regarding option-structs with required fields, we could offer
`IirFilterOptions::from(feedforward: Vec<f64>, feedback: Vec<f64>)`
`AudioBufferOptions::from(length: usize, sample_rate: SampleRate)`
filling the other fields with their default values.

src/node/media_element.rs
```diff
@@ -9,10 +9,14 @@ use super::{
     ChannelConfigOptions, MediaStreamRenderer,
 };
 
-/// Options for constructing a MediaElementAudioSourceNode
+/// Options for constructing a [`MediaElementAudioSourceNode`]
+// dictionary MediaElementAudioSourceOptions {
+//   required HTMLMediaElement mediaElement;
+// };
+// note that `MediaElementAudioSourceOptions` does not extend `AudioNodeOptions`
```
orottier, line 16: Very sharp! Let's drop it (and fill with defaults/panics in the `AudioNode`'s required methods

orottier:
Great too see your sticking with
`#[derive(Copy, Clone, Debug)]` for pub enums
`#[derive(Clone, Debug)]` for pub structs

I notice `AudioBufferOptions` has `Copy` too. Let's drop that

src/node/panner.rs
```diff
@@ -4,34 +4,76 @@ use std::sync::Arc;
...
-use crate::{AtomicF32, SampleRate};
+use crate::{AtomicF64, SampleRate};
 
 use super::{
     AudioNode, ChannelConfig, ChannelConfigOptions, ChannelCountMode, ChannelInterpretation,
 };
 
-/// Options for constructing a PannerNode
+#[derive(Copy, Clone, Debug)]
+pub enum PanningModelType {
```
orottier, line 14: I will spend some precious time tomorrow to check if we now have all enums and options conforming to the spec and we avoid breaking changes in the future 😄 

Probably should mention all unused options in the rustdocs. If we annotate them with `TODO(unimplemented)` or something we can make a nice tracking issue out of them

b-ma:
Ok, cool! I let you have a second pass with a fresh eye then!

orottier:
Why spend 1 hour doing things manually when you can spend 4 hours trying to automate it ;)
I've added the IDL file to do some scripting, will continue with the checks later

src/node/media_element.rs
orottier: Same goes for ConstantSourceOptions

orottier:
Some more notes:

ChannelConfigOptions, should be AudioNodeOptions and the fields should be renamed. But instead of inheriting from it for the option structs, we use the struct as a field. I think that is fine. (decision: Inheritance of pub structs is done via composition?)

AudioContextOptions.channels is not part of the spec (as it will use system defaults), but our implementation is fine (which is: allows users to supply and optional preferred number of channels)

AutomationRate::A should be AutomationRate::ARate, but I think that is not looking good. Let's keep it this way

PanningModelType::EqualPower should be Equalpower, this seems to be a typo in the spec. Let keep it in current state

OverSampleType, 2X is not permitted so we use X2. Fine with me

`create_delay` and `create_channel_(splitter/merger)` take optional argument, we make them required. Same as `create_periodic_wave` but here we take in an opt struct. Todo we should fix this.

`create_media_(element/stream)_source` now lives on base context but must be online context only. Fine for now?

Mark BaseAudioContext `create_audio_param` pub(crate) and introduce `AudioParam::new(opts, node)` ?

How to deal with promises? Accept sync functions for now, add async later and call functions `start_rendering_async`?

`AudioBufferOptions.number_of_channels` is optional in the spec (deafults to 1). We make it required and I think that is fine

Will polish later!

orottier:
Alright, I have checked all constructor options now and applied some fixes.
I will add some more todos to the issue #76 
Let's merge this and have a look at the other tasks?

b-ma:
Hey, cool all this seems pretty good to me and going to the right direction, so let's merge!

Just a few remarks on your remarks:

> ChannelConfigOptions, should be AudioNodeOptions and the fields should be renamed. But instead of inheriting from it for the option structs, we use the struct as a field. I think that is fine. (decision: Inheritance of pub structs is done via composition?)

Seems coherent with modelling inheritance with traits too, in both cases this is composition

> AudioContextOptions.channels is not part of the spec (as it will use system defaults), but our implementation is fine (which is: allows users to supply an optional preferred number of channels)

This will probably appear in the future, and this is nice to have for sure. The problem could maybe come from the naming, maybe `number_of_channel` might be a better choice as it seems to be the preferred term everywhere else? Maybe we should try to dig into the spec v-next issues, see in which direction it seems to go at the moment.

> AutomationRate::A should be AutomationRate::ARate, but I think that is not looking good. Let's keep it this way
> PanningModelType::EqualPower should be Equalpower, this seems to be a typo in the spec. Let keep it in current state
> OverSampleType, 2X is not permitted so we use X2. Fine with me

In any cases, we will have some discrepancies and compromises with the conversion between enum defined as "string" and rust enums (something I already see in the Node.js bindings). The important thing I guess is that we know the deviations, their justification and document them properly. (In the same vein, there are also `type`/`type_` and `loop`/`loop_` for which we can't do much because the are rust keywords...)

Not really important but for `ARate`, even if I'm really not sure I like it neither, but I'm not so sure at contrary that "I don't like it" is good justification (...but it's also a right to decide it is a good one :)

> create_delay and create_channel_(splitter/merger) take optional argument, we make them required. Same as create_periodic_wave but here we take in an opt struct. Todo we should fix this.

Yup, good point

> create_media_(element/stream)_source now lives on base context but must be online context only. Fine for now?

I don't see that as a big issue and this is very probably helpful for testing (same as before I would say, if the deviation has a clear justification and is documented, I really think that's ok)

> Mark BaseAudioContext create_audio_param pub(crate) and introduce AudioParam::new(opts, node) ?

As neither ones are part of the spec, I would say `create_audio_param` is fine. 

If we want the creation of new nodes to be closer to the `AudioWorklet` API maybe we could look toward [https://webaudio.github.io/web-audio-api/#parameterdescriptors](https://webaudio.github.io/web-audio-api/#parameterdescriptors)? But this is probably another discussion, the question of extensibility is both interesting but tricky I thing.

> How to deal with promises? Accept sync functions for now, add async later and call functions start_rendering_async?

Ah... hard one, breaking change in v2? :) (it seems Rust community don't really like breaking changes)

Another possibility could be to already create an alias such as `start_rendering_sync` and add a warning on `start_rendering` saying that it will become `async` in the future. So people who want to be safe would use `start_rendering_sync` and adventurers could do what they want but knowing the risk they take.

> AudioBufferOptions.number_of_channels is optional in the spec (deafults to 1). We make it required and I think that is fine

For now, it seems that every option struct that have a required fields (i.e. `AudioBufferOptions`, `IIRFilterOptions` and `PeriodicWaveOptions`, and I think that's all) do not implement `Default`, I think that's both fine and consistent.

orottier:
Thanks for the feedback, I added the leftover points to #76 
Let me know when you want to pick something up, I will focus now on

- mark functions that should return a Promise but currently don't _sync
- rename AudioContextOptions.channels -> number_of_channels
- AudioContext::new(AudioContextOptions) (not wrapped in an Option)
- remove public spatial mod

b-ma:
Ok ! I will have a look at the `AudioScheduledSourceNode` thing then.

I will also review the PeriodWave constructor, I have the impression there are some misunderstanding there

PR #104, opened 2022-01-27
Mark functions that should return a Promise but currently dont, `_sync`

orottier:


b-ma:
I would have kept the unstable non`_sync` alias I guess, but it's probably more clean like that. 
(I must admit I'm not very conservative with breaking changes as I really think whatever you do software has to evolve and be maintained at the end, but that's a point of view I can understand not everyone agrees with :) 

So, looks good to me!

examples/trigger_soundfile.rs
```diff
@@ -8,7 +8,7 @@ fn main() {
 
     // load and decode buffer
     let file = File::open("samples/sample.wav").unwrap();
-    let audio_buffer = context.decode_audio_data(file).unwrap();
+    let audio_buffer = context.decode_audio_data_sync(file).unwrap();
 
     // @fixme - if only one node in the graph it is never removed even when returning
```
b-ma, line 13: absolutely not related, but this should be rechecked that was kind of strange

src/context.rs
```diff
@@ -523,16 +531,20 @@ impl AudioContext {
         }
     }
 
-    /// Closes the `AudioContext`, releasing the system resources being used. This will not
-    /// automatically release all `AudioContext`-created objects, but will suspend the progression of
-    /// the currentTime, and stop processing audio data.
+    /// Closes the `AudioContext`, releasing the system resources being used.
+    ///
+    /// This will not automatically release all `AudioContext`-created objects, but will suspend
+    /// the progression of the currentTime, and stop processing audio data.
```
b-ma, line 537: So the nodes renderers will never be dropped, if `graph::render` is never called anymore? (that's really not a big deal for now I guess)

examples/trigger_soundfile.rs
b-ma: I will try to figure out how to make a test for that, that's something I've seen with dirty `println!`

src/context.rs
orottier: To be 100% sure we could write a `Drop` implementation for the `GainProcessor` for example that will `println` when it is dropped. Then we can check the simple example and see if what I am saying is actually true ;)

examples/trigger_soundfile.rs
orottier: Thanks, very odd indeed and something we should investigate

orottier:
> I must admit I'm not very conservative with breaking changes

Yeah.. The rust community seems to disagree with that but I'm impartial too. I'm going to merge!

Issue #84, opened 2022-01-05
How to deal with related web concepts (<audio> element, WebRTC, MediaCapture)


orottier:
As mentioned in #82:

To be fair I'm struggling a bit with the 'boundaries' of the Web Audio API also in other cases. For #77 I was playing around with microphone input and playback and realized I needed to have something like https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder

There are a few conflicting goals:
- stay as close as possible to the spec (as that is the unique selling point of this project, provide the comfort of not learning a new API when coming from a web background)
- leverage all the power of rust (no nullables, no exceptions, use ownership etc)
- create a small library (it's a lot of work already in its current state), versus:
- include related APIs (https://w3c.github.io/mediacapture-main/ https://w3c.github.io/webrtc-pc/ https://html.spec.whatwg.org/multipage/media.html#the-audio-element )
- or include include a few intefaces acting as building blocks (the `MediaStream` trait)

I will create a separate issue for this. It's probably best to steer away from names such as `MediaStream` and `MediaElement` because as you pointed out these are different from the w3c specs. I do think we should include them in the library because you would need them in many applications anyway

_Originally posted by @orottier in https://github.com/orottier/web-audio-api-rs/pull/82#discussion_r778097042_

orottier:
I have added this to the v1.0 checklist. We don't need to have an allround solution but we do need to put some thought in keeping/renaming/dropping the MediaStream trait, the MediaElement en the Microphone input

b-ma:
I think a quite clean tradeoff could be:

Keep the `Microphone`, `MediaElement` and `MediaStream` (with eventual renaming: `AudioPlayer` ? `AudioElement` ? `WebAudioStream` ?)

- I really think that the `Microphone` and `MediaElement` are really nice to have at hand without having to install another lib and write manually all the pipes and interfaces for most usage. And if I understand well if one needs to go beyond these helpers, the `MediaStream` trait makes it possible.

- As all these abstraction are in the `media` namespace (I still think `extra` would be more clear but that's really not a big deal), this is really simple to document and explain that they are provided to be helpful but actually belong to other Web Specs and do not aim at implementing all the features from these specs. 

- If someone wants to implement one of these APIs with all their features, the interface between the two libs seems possible thanks to the `MediaStream` trait, so that's really nice

[...]

PR #106, opened 2022-01-28
Attempt to harmonize AudioScheduledSource and make Scheduler and Controller private

b-ma:
Hey, 

This is a work in progress, mostly working (but the "mostly" seems unfortunately quite important there...).

I managed to: 
- mark the `Scheduler` and `Controller` as `pub(crate)` 
- make `ConstantSource`, `Oscillator` and `AudioBufferSourceNode` derive `AudioScheduledSourceNode`

On these points, I have the impression the solution is pretty much where we wanted to go.

But I can't figure out how to deal with `media::MediaElement`. My idea was to port every scheduling/control methods inside the MediaElement, in such way that `MediaStreamRenderer` does not rely anymore on a shared `scheduler` (which seems to work well for the microphone), and that `scheduler` and `controller` are just implementation details of the `MediaElement`. The user code would end up to something like that:

```rs
// examples/media_element.rs
use web_audio_api::context::{AudioContext, BaseAudioContext};
use web_audio_api::media::{MediaDecoder, MediaElement};
use web_audio_api::node::{AudioNode, MediaElementAudioSourceNode, MediaElementAudioSourceOptions};

fn main() {
    env_logger::init();

    // construct the decoder
    let file = std::fs::File::open("samples/major-scale.ogg").unwrap();
    let stream = MediaDecoder::try_new(file).unwrap();
    // Wrap in a `MediaElement` so buffering/decoding does not take place on the render thread
    let media_element = MediaElement::new(stream);

    // pipe media_element in the WebAudio graph

    let node = MediaElementAudioSourceNode::new(&context, MediaElementAudioSourceOptions {
        media_element
    });
    node.connect(&context.destination());

    // play media
    media_element.start();

    // enjoy listening
    std::thread::sleep(std::time::Duration::from_secs(4));
}
```

Which I think is quite clean and more compliant from an API perspective, but of course creates an ownership issue that I don't see how to approach

```
  --> examples/media_element.rs:23:5
   |
12 |     let media_element = MediaElement::new(stream);
   |         ------------- move occurs because `media_element` has type `MediaElement`, which does not implement the `Copy` trait
...
18 |         media_element
   |         ------------- value moved here
...
23 |     media_element.start();
   |     ^^^^^^^^^^^^^^^^^^^^^ value borrowed here after move
```

Would be best if could have an eye on that, before continuing (or dropping it...)

_Disclaimer_ the projet is in a really dirty state for now, with things commented to isolate the problem or waiting to be completely removed.


orottier:
Yeah I see. The way it works now is that the node takes ownership over the media element and the node will delegate start/stop to the stream internally.

It's definitely possible to stick to the spec better and make your example code compile. The solution would be a bit similar to ConcreteBaseAudioContext + Inner. Create an Arc on the inside that allows for shared ownership. Then you can say something like `MediaElementAudioSourceNode::new(&context, &media_element)` and the ownership problem does not occur. In that `new` function, you clone the `Arc` so you no longer need the borrow of the media_element.

I can whip something up if you would appreciate it. I think your proposal is better than the current impl

b-ma:
Ah thanks, I see the idea I think ! I will try do it, seems like a good exercise (and I tell you if I miserably fail :)

b-ma:
Hey, I've managed to make the `MediaElement` working.

Two notes, that are quite important however:
- with current implementation of the Iterator, if the `MediaElement` is passed twice to an  `MediaElementAudioSourceNode`, the file will be played at speed * 2. As a hot fix, I basically added a `panic!` if the `MediaElement` is cloned more that once. This should probably be reviewed, but I'm sure if or how we could achieve something more robust with the current `MediaStream`?
- I also commented `(set_)loop_start` and `(set_)loop_end` in the `MediaElement`, because 1. it is not really part of the MediaElement spec, and it will make any further refactoring / improvement much more complicated, and 2. I think the most important thing to do first would be to keep in memory only small subset of the input stream with a 1 or 2 sec fixed lookahead, instead of the whole file as seems to be done actually. Maybe after that, we could eventually reintroduce the looping point, what do you think?

Let me know what you think of these changes, then I can continue with the renaming discussed in #84 (or in another PR, as you prefer)

b-ma:
I just forgot this one for a while, would be nice if you could have a look!

src/lib.rs
```diff
@@ -135,6 +129,11 @@ impl AtomicF64 {
             .swap(u64::from_ne_bytes(v.to_ne_bytes()), Ordering::SeqCst);
         f64::from_ne_bytes(prev.to_ne_bytes())
     }
+
+    pub fn fetch_add(&self, v: f64) -> f64 {
```
orottier, line 133: This one is a bit dangerous. It's a method that suggests to be thread safe but actually isn't!
In practise this is not an issue because only the render thread (a single thread) will call methods that update the value. The control thread (possibly multi-threaded) only reads the value, which is then safe.
In any case I suggest we start using  https://crates.io/crates/atomic_float

orottier:
Sorry for my delayed response. The work on the `AudioScheduledSourceNode` is great and we should merge ASAP.
However, I realized a few drawbacks in the way we model the media element now. I need some more time to put it in writing so check back later today please! Your implementation is totally sound and safe btw, kudo's for proper usage of Arc, Mutex and friends :)

b-ma:
Cool, nice that you can have a look and take over, I was kind of stuck at this point :)

orottier:
I was drafting a commit so we could merge the work on `AudioScheduledSourceNode` and leave the discussion on the workings of the `MediaElement` in a separate PR. But it is taking more time than expected so I will leave my concerns here already so you have some time to think about it.

With this PR the media element can be used how people would expect it. The following pattern is now possible:
```
// create context
// create media element
{
    // create MediaElementAudioSourceNode
    // connect to DestinationNode
    // start now and play for 1 second
}
// sleep for 1 second so previous node goes out of scope
{
    // create MediaElementAudioSourceNode
    // connect to DestinationNode
    // start at ts=3 and play for 1 second
}
```

With this script, you would expect to hear the first second of the media element output, and the output from t=3 to 4.
But this exposes the big lie of our MediaElement. There is no progression of time when it is not polled by the AudioGraph.
This means the output from t = 3 to 4 is actually the output of t = 1 to 2. And what if you would `suspend` the audio context in the meantime?
You had already noticed this of course, the progression of time is doubled when there are two MediaElementAudioSourceNode present and pulling audio from the MediaElement

In a webbrowser there is no issue of course. All AudioContexts and MediaElements are glued together by a single DOM which has a guaranteed progression of time. In our situation, that is not possible.
This means that the previous implementation, where the MediaElement is absorbed into the MediaElementAudioSourceNode, is more fair to the user because it guarantees progression of time and it makes sure you cannot run into unexpected situations. Because ownership is taken, you cannot add a media element twice and double speed progression is also prevented.

Now this precisely shows how our model breaks down when we run into the boundaries of the API. Both implementations have pro's and cons. Let's think about it some more.

Taking a step back, what are we trying to achieve with all the different nodes:
**AudioBufferSourceNode** nice and simple, advanced playback controls. Load the buffer up front, loop, seek, playback as you want
**implement AudioProcessor yourself** full control over audio output. For experts only
**MediaStream node**: Iterator of `AudioBuffer` that runs on the render thread. This is useful for web-rtc style applications
**MediaElement node**:  Iterator of `AudioBuffer` that buffers from the control thread. It allows looping and pausing. Useful if you want to play (looping) music from a network source for example.

Maybe the use case for MediaElement is too weak now? Especially since the buffering implementation is very dumb. We could leave it out for now and focus on other things? Or leave it in the previous implement and document the shortcomings?

b-ma:
Yup I think you accurately summarised the problem(s) here... I really agree with you that we need to think more about it and have more input / feedback before continuing on this.

To add my 2 cents to the discussion, my impression is that 99.9% of the `MediaElementSourceNode` use cases are simple audio players that play files one by one with visual feedback (e.g. VU Meter, spectrum using `AnalyserNode`), eventually apply some filters for equalisation, and even more eventually add some reverb.

I would personally advocate to keep this version compared to the previous one (ok I'm biased :), but), I think despite its issues (which should of course be properly documented) that's the more "future-safe" tradeoff: 
- It allows to hide some of the internal we want to hide (e.g. `scheduler` and `controller`)
- It doesn't add non-spec methods to the nodes that we would be stuck with after the fact (i.e. implement `AudioScheduledSourceNode` on `MediaElementAudioSourceNode`)
- the API goes to the right direction even if not completely finished / satisfying
- I'm really not sure we can have all this with the previous version of the `MediaElement`

Maybe we could even consider hiding all the boundaries stuff behind a `nightly` or `experimental` flag for v1. That would definitely claim it is something we had a look and aim at supporting but still a work-in-progress, so people can use it and give us feedback but they know it is unstable and will need a bit of maintenance.

Side notes: 
- as a first next step on the `MediaElement`, I think the buffering issue should be the most important thing to do to make it at least usable in real world use cases (i.e. very long files)
- I will also try to dig a bit more on how all this behaves in JS, that's really not a part of the API I use and know very well.

orottier:
Thanks for sharing your point of view.
I decided to rip the bandaid of and drop the MediaElement and MediaElementSourceNode for now. To be continued in #120 (will update it with more info) after the v1 release.
I will do a final pass over this PR tomorrow and then we can safely merge this!

b-ma:
Haha, quite radical but that works indeed :)

Just a final consideration then, to complete the discussion on the `MediaStreamAudioSourceNode` side:

> MediaStream node: Iterator of AudioBuffer that runs on the render thread. 
> MediaElement node: Iterator of AudioBuffer that buffers from the control thread.

Not sure this distinction is the right one actually: it's can be true indeed if the input is a microphone stream, but if the stream is a WebRTC audio stream from the network I'm not sure it can be describe the problem that way anymore. 

I just found this page of examples which is quite interesting here I guess [https://dvcs.w3.org/hg/audio/raw-file/tip/webaudio/webrtc-integration.html](https://dvcs.w3.org/hg/audio/raw-file/tip/webaudio/webrtc-integration.html)

orottier:
Cool link, thanks. Maybe I was not wording it careful enough.

Both the MediaStream node and the MediaElement node are polled on the render thread. The MediaElement however features an internal buffer that is filled by a dedicated decoding thread and it will raise an error when the decoding is slower than the playback. The MediaStream node has no buffer so you will have to write your own wrapper to prevent catastrophic delays. 

A webRTC wrapper (or any audio feed that involves IO) would look a lot like the Microphone setup. Spawn a realtime thread to perform the IO polling and ship the audio buffers via a channel to the `WebRTC` object. The `WebRTC` object can implement the MediaStream trait so it can be passed to the audio context and the object will perform its own buffering/stretching/delaying of samples when the two realtime threads move out of sync.

I realize this makes a lot of sense in my head but perhaps not for others. In style with #77 I could write an example that spawns an UDP connection between two hosts so you can make a WebRTC-style call? Would that help to move the discussion forward?

b-ma:
> I realize this makes a lot of sense in my head but perhaps not for others. In style with https://github.com/orottier/web-audio-api-rs/issues/77 I could write an example that spawns an UDP connection between two hosts so you can make a WebRTC-style call? Would that help to move the discussion forward?

If it's clear for you, then that's perfect! And in any case, yes an example of this kind would be super nice

b-ma:
cool !

PR #107, opened 2022-01-28
PeriodicWave - Reviewed constructor and wavetable size according to spec

b-ma:
Small fixes concerning the `PeriodicWave`

For the wavetable size, that's what I understand from the spec "A conforming implementation MUST support PeriodicWave up to at least 8192 elements." (cf. https://webaudio.github.io/web-audio-api/#PeriodicWave)

Let me know if you read it another way (I tried to check in Chrome and Firefox code source and it's hard to say because their implementation is far more complex using iFFT and bandlimiting strategies from what I understand, but for sure they do not check the size of `real` and `imag` as we were doing).

### DIFF
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -44,7 +44,7 @@ pub use waveshaper::*;
 mod stereo_panner;
 pub use stereo_panner::*;
 
-pub(crate) const TABLE_LENGTH_USIZE: usize = 2048;
+pub(crate) const TABLE_LENGTH_USIZE: usize = 8192;
 pub(crate) const TABLE_LENGTH_BY_4_USIZE: usize = TABLE_LENGTH_USIZE / 4;
 // 2048 casts without loss of precision cause its mantissa is 0b0
 #[allow(clippy::cast_precision_loss)]
--- a/src/periodic_wave.rs
+++ b/src/periodic_wave.rs
@@ -15,14 +15,16 @@ pub struct PeriodicWaveOptions {
     /// This offset has to be given but will not be taken into account
     /// to build the custom periodic waveform.
     ///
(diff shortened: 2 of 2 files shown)


src/node/mod.rs
```diff
@@ -44,7 +44,7 @@ pub use waveshaper::*;
 mod stereo_panner;
 pub use stereo_panner::*;
 
-pub(crate) const TABLE_LENGTH_USIZE: usize = 2048;
+pub(crate) const TABLE_LENGTH_USIZE: usize = 8192;
 pub(crate) const TABLE_LENGTH_BY_4_USIZE: usize = TABLE_LENGTH_USIZE / 4;
 // 2048 casts without loss of precision cause its mantissa is 0b0
 #[allow(clippy::cast_precision_loss)]
```
orottier, line 50: You should check this comment, does it uphold for the new value?

b-ma: good shot, no idea actually... will try to understand what that means exactly
(...I would be quite surprised that we can loose precision with such small numbers but...)

b-ma: Actually, I'm really not a computer scientist, but I don't understand that comment, it seems wrong to me

1. If we agree a floating point number is represented as: `±mantissa * 2^exponent`, if the mantissa is 0 (...or `0b0`), then the represented number is 0 (as far as I know `0 * x` will always be 0, whatever the exponent)
2. The mantissa seems to be generally represented by 23 bits for f32, so _any_ integer up (or down) to `2^23` can be perfectly represented with an exponent of 0 (without even considering that the number we are talking about is a power of two)

Or am I missing something?

b-ma: Actually these directives where not even required by clippy (at least in its CI config) so I just removed them

PR #111, opened 2022-02-14
AudioBuffer & AudioBufferSourceNode resampling

b-ma:
Hey, 

Here is mostly a simple patch on the `AudioBufferSourceNode` so that it can consume `AudioBuffer`s with non matching sample rates (`AudioBuffer` decoded by other contexts with different sample rates).

I also added an example to show the 2 cases with have for `AudioBuffer`s resampling
1. `AudioContext.decode_audio_data` resample to its own sample rate
2. Resampling is done by the `AudioBufferSourceNode` in the edge case explained before

As you can hear in the example, the resampling as performed by `decode_audio_data` introduce a lots of artefacts, actually more than the linear interpolation performed by the `AudioBufferSourceNode`.

I tried to update `AudioBuffer::resample` to use `rubato`, but couldn't manage to make something work. I must say I'm a bit disappointed with this library, the doc is hard to understand and I don't really get how it is supposed work. I've made some small research and found this crate [https://docs.rs/speexdsp-resampler/latest/speexdsp_resampler](https://docs.rs/speexdsp-resampler/latest/speexdsp_resampler), which is a port of a C library (which looks battle tested) created with [c2rust](https://github.com/immunant/c2rust). Its API looks more intuitive and exposes some values we are interested in such as the latency introduced by the resampling.

Would you be ok that we test it or do you have any bad experience using libs ported that way ? (as we only use `rubato` in the `WaveshaperNode` for now, it would still be quite easy to switch I guess).


orottier:
Great that you are looking into this. A few remarks:

- The `Resampler` that we have currently is a total joke. It does not even perform linear interpolation
- Ideally, there is only a single `Resampler` in the whole project that all nodes and the context can call into.
- We can open issues on the rubato project to help improve the project. I like that it is a pure rust solution and that they are taking care of SIMD-friendliness
- I'm following the progress in https://github.com/HEnquist/rubato/pull/29 which will be important for us I think. I do agree it is quite hard to understand what each resampler does and if it allocates/delays/etc
- In the meantime we can of course experiment with the c2rust solution

b-ma:
Ok thanks for the feedback

> Ideally, there is only a single Resampler in the whole project that all nodes and the context can call into.

I'm not sure to understand well what you mean here: do you mean having a wrapper abstraction that allow to isolate the dependency ? or having only one instance (i.e. singleton) ? I don't really think the later case is possible, as, if I understand well, most algorithms apply some sort of lowpass filtering (i.e. the latency introduced) which has an internal memory that would be polluted if we just dropped unrelated signals in only one instance.

> We can open issues on the rubato project to help improve the project. I like that it is a pure rust solution and that they are taking care of SIMD-friendliness
> I'm following the progress in https://github.com/HEnquist/rubato/pull/29 which will be important for us I think. I do agree it is quite hard to understand what each resampler does and if it allocates/delays/etc
> In the meantime we can of course experiment with the c2rust solution

Yup I see your point. I will try to create a more isolated testbed so we can try, test and compare these 2 libs more simply. 

I also wonder if this point should be v1 milestone too, as that's really part of the "basics features" you ask to an audio lib I think.

So, if it's ok for you, let's merge this one and continue on another PR ! I really think we can consider the linear interpolation patch on the `AudioBufferSourceNode` is here to stay, as in my opinion, it's enough to handle the weird edge case its supposed to handle (...I don't even see how this could happen in real life actually, expect by specifically looking for problems :)

orottier:
>  a wrapper abstraction that allow to isolate the dependency 
This one indeed. I'm also in doubt if we should actually expose the `Resampler` in the public API. It's nice and convenient to have (because the number of options via external crates can be overwhelming). In any case it should move to `extra`

A decent linear interpolation Resampler should be fine for v1 yes

orottier:
I added #112 to the v1 Milestone. Agree with that direction?

b-ma:
Yes agreed, seems the right way to go. More safe to expose something later than exposing something we are not sure of and get stuck with it.

PR #113, opened 2022-02-16
Small improvements

orottier:
Hey @b-ma could you check for me your output for
`cargo test -- test_audio_buffer_resampling`

For me it returns an error consistently, but it works in `--release` mode.
Very strange because the test looks sound to me.
It appears to me that in debug mode the node stops returning samples prematurely. So the left hand side of the assert contains many zeroes..

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ rubato = "0.10"
 float_eq = "0.7"
 smallvec = "1.7.0"
 once_cell = "1.9.0"
-symphonia = { version = "0.4.0", default-features = false }
+symphonia = { version = "0.5.0", default-features = false }
 
 [dev-dependencies]
 rand = "0.8.*"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ repository = "https://github.com/orottier/web-audio-api-rs"
 keywords = ["web-audio-api", "audio", "sound"]
 license = "MIT"
 categories = ["multimedia::audio"]
+exclude = ["/samples", "/snapshots"]
 
 [dependencies]
 cpal = "0.13.1"
--- a/src/context.rs
+++ b/src/context.rs
@@ -126,7 +126,7 @@ pub trait BaseAudioContext {
     /// // await result from the decoder thread
     /// let decode_buffer_result = handle.join();
     /// ```
-    fn decode_audio_data_sync<R: std::io::Read + Send + 'static>(
+    fn decode_audio_data_sync<R: std::io::Read + Send + Sync + 'static>(
         &self,
         input: R,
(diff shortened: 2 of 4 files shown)


b-ma:
Damn that's weird, I have absolutely no problems on my side

I really don't see what could cause that. Actually the sampling ratio only applies on the interpolation in the buffer, it doesn't change any timing informations, so I don't think it's related to my last changes.

Could you make these 2 tests by replacing the beginning of `test_audio_buffer_resampling`:

1. see if it occurs when there is no resampling

```rs
[48000].iter().for_each(|sr| {
  let base_sr = 48000;
  // ...
});
```
 
2. change to smaller values so we can eventually have more meaningful output

```rs
[128, 256, 512].iter().for_each(|sr| {
  let base_sr = 256; // make sure we have at least two blocks computed
  // ...
});
```

orottier:
Well this is just the weirdest thing ever.

In release mode
```
[src/node/audio_buffer_source.rs:732] &sr = 22500
[src/node/audio_buffer_source.rs:737] *sr = 22500
[src/node/audio_buffer_source.rs:739] buffer.number_of_channels() = 1
[src/node/audio_buffer_source.rs:740] buffer.length() = 22500
[src/node/audio_buffer_source.rs:741] buffer.sample_rate_raw() = SampleRate(
    22500,
)
[src/node/audio_buffer_source.rs:742] buffer.sample_rate() = 22500.0
[src/node/audio_buffer_source.rs:743] buffer.duration() = 1.0
```
which looks good.

In debug mode:
```
[src/node/audio_buffer_source.rs:732] &sr = 22500
[src/node/audio_buffer_source.rs:737] *sr = 22500
[src/node/audio_buffer_source.rs:739] buffer.number_of_channels() = 1
[src/node/audio_buffer_source.rs:740] buffer.length() = 22500
[src/node/audio_buffer_source.rs:741] buffer.sample_rate_raw() = SampleRate(
    5966384,
)
[src/node/audio_buffer_source.rs:742] buffer.sample_rate() = 5966384.0
[src/node/audio_buffer_source.rs:743] buffer.duration() = 0.003771128375243699
```

Somehow the sample rate turned into garbage. This explain why your test returns many zeroes. It believes it is done playing after a few millis. Wondering if this is a compiler bug

b-ma:
Indeed, that's really weird... could it be the conversion to u32? so changing

```rs
let mut buffer = context.create_buffer(1, buf_sr, SampleRate(buf_sr as u32));
```

to 

```rs
let mut buffer = context.create_buffer(1, buf_sr, SampleRate(u32::try_from(buf_sr).unwrap());
```

would be more safe?

orottier:
Well that solves it, incredible.
`buf_sr` is a usize but with small enough value to fit in a u32. Somehow that works in release mode but not in debug on mac hardware..
Time to enable the lint to deny unsafe casts

Issue #112, opened 2022-02-16
Improve internal `Resampler` and hide from the public API for now


orottier:
The current `Resampler` is of really low quality. It does not even perform linear interpolation and produces audible artifacts.

Two tasks which we deem necessary pre v1 release:
- [ ] rewrite to use linear interpolation
- [ ] drop it from the public API as it is not part of the spec. We may introduce it later in the `extra` mod. Users are advised to use a new `OfflineAudioContext` with `decode_audio_data` if they want to manually resample a buffer.

PR #116, opened 2022-02-18
Feature/resampling

b-ma:
Hey, 

related to #112 
- `Resampler` now lives in its own file in `media`
- `Buffer::resample` does a linear interpolation

The example `cargo run --example resampling` now sounds better with `decode_audio_data`

Between, I have made a rapid test with the `speex_resample` crate, that seems very slow for us I think and I have a strange bug on the right channel which just disappears. 

To be continued... (seems we will have to dig more on the `rubato` side)

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -2,7 +2,7 @@
 use std::sync::Arc;
 
 use crate::render::AudioRenderQuantum;
-use crate::SampleRate;
+use crate::{assert_is_valid_number_of_channels, assert_is_valid_sample_rate, SampleRate};
 
 /// Options for constructing an [`AudioBuffer`]
 #[derive(Clone, Debug)]
@@ -33,9 +33,12 @@ impl AudioBuffer {
     ///
     /// # Panics
     ///
-    /// This function will panic if the channel count is zero.
+    /// This function will panic if:
+    /// - the sample rate is outside the range [8000, 96000]
+    /// - the number of channel is outside the range [1, 32]
     pub fn new(options: AudioBufferOptions) -> Self {
-        assert!(options.number_of_channels > 0);
+        assert_is_valid_sample_rate(options.sample_rate);
+        assert_is_valid_number_of_channels(options.number_of_channels);
 
         let silence = ChannelData::new(options.length);
 
@@ -51,15 +54,19 @@ impl AudioBuffer {
     ///
     /// # Panics
     ///
-    /// This function will panic if `samples` is an empty Vec or any of its items have different
-    /// lengths.
+    /// This function will panic if:
+    /// - the given sample rate is outside the range [8000, 96000]
+    /// - `samples` is an empty Vec
+    /// - any of its items have different lengths
     pub fn from(samples: Vec<Vec<f32>>, sample_rate: SampleRate) -> Self {
-        assert!(!samples.is_empty());
+        assert_is_valid_sample_rate(sample_rate);
+        assert_is_valid_number_of_channels(samples.len());
 
         let channels: Vec<_> = samples.into_iter().map(ChannelData::from).collect();
         if !channels.iter().all(|c| c.len() == channels[0].len()) {
             panic!("Trying to create AudioBuffer from channel data with unequal length");
         }
+
         Self {
             channels,
             sample_rate,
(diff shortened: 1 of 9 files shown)


src/buffer.rs
```diff
@@ -184,12 +181,6 @@ impl AudioBuffer {
         &mut self.channels[index]
     }
 
-    /// Modify every channel in the same way
```
orottier, line 187: For context of this method see #117 
It's fine that this is dropped now

src/buffer.rs
```diff
@@ -233,13 +224,14 @@ impl AudioBuffer {
         AudioBuffer::from_channels(channels, sample_rate)
     }
 
-    /// Resample to the desired sample rate.
```
orottier, line 236: Please leave the `///` in place. If you run rustdoc with `--document-private-items`, it's nice to have this available.
Also put the example usage in three slashes. Not sure why this was not the case already.

src/buffer.rs
```diff
@@ -249,7 +241,7 @@ impl AudioBuffer {
     //
     // assert_eq!(
```
orottier, line 242: Should be `assert_float_eq` now

src/media/resampling.rs
```diff
@@ -0,0 +1,141 @@
+use std::error::Error;
+
+use crate::buffer::{AudioBuffer, AudioBufferOptions};
+use crate::media::MediaStream;
+use crate::SampleRate;
+
+/// Sample rate converter and buffer chunk splitter.
```
orottier, line 7: The docs and example usage are now gone

orottier (changes requested):
Great work. I added a few comments to have a look at and will push a failing test case in a minute

src/lib.rs
```diff
@@ -137,6 +138,39 @@ impl AtomicF64 {
     }
 }
 
+/// Utility functions for arguments sanity check
+pub(crate) fn assert_is_valid_sample_rate(sample_rate: SampleRate) {
```
b-ma, line 142: I introduced these function, I thought it could be nice to have more of them to harmonize checks and errors messages (thinking about AudioParams timing arguments that should be checked at some point). Let me know if you agree with that, can be perfectly be moved in `buffer.rs`.

I tried to add them in `ConcreteBaseAudioContext::new` but that was crashing many tests, maybe for another PR if you think we should continue in this direction.

src/media/resampling.rs
b-ma: done!

src/buffer.rs
b-ma: done (there are ignored from `cargo test --doc` though)

b-ma:
Hey, just added other checks, example and documentation.

As a side note, we have a discrepancy with types here, i.e. `number_channel: usize` and `length: usize`, which should both be `u32` if we strictly follow the spec. Also this is inconsistent with other places of the code (e.g. `ConcreteBaseAudioContext`). I started to update that but it has many implications here and there so I just reverted back.

Not sure what we should decide though: `unsigned long` is `usize` or `u32`? As they are generally used to as indexes, maybe `usize` would make more sens somehow but I'm not really sure of all the implications of such choice. My only opinion here is that we should be very consistent and apply the same rule everywhere. Maybe creating and maintaining a map of the correspondances somewhere would be helpful? (maybe we have a similar problem with `&[f32]` and `Vec<f32>` at some places)

src/lib.rs
orottier: Looks great, let's do it this way indeed.

Few remarks:
- naming: `assert_is_valid_sample_rate` --> `assert_valid_sample_rate` (same for the others)
- could you document the panics in the docblock, so all callers can copy that in their docblock
- I think `8000 to 96000` is needlessly strict. We only need to guard for `>0` to prevent crashes, so let's do only that?

src/lib.rs
```diff
@@ -137,6 +138,48 @@ impl AtomicF64 {
...
+                sample_rate
+            );
+        }
+    }
+}
+
+pub(crate) fn assert_is_valid_number_of_channels(number_of_channels: usize) {
+    if number_of_channels == 0 || number_of_channels > MAX_CHANNELS {
+        panic!(
+            "NotSupportedError - Invalid number of channels: {:?} is outside range [1, 32]",
```
orottier, line 168: Can you replace the harcoded `32` with `{}` and a `MAX_CHANNELS` argument?
In theory, a later version could allow for more channels of course (either by forking and recompiling, or by introducing const generics ourselves)

b-ma: right, it's fixed

src/lib.rs
b-ma: ok done, I have added a comment about "real-life" sample rates in the doc block

orottier:
Great work, thanks!

Issue #117, opened 2022-02-19
Optimizations for AudioRenderQuantum when all channels are equal


orottier:
AudioRenderQuantum has a method `modify_channels` that modifies each channel in the buffer with the same function.
This allows for an optimization when all channels in the buffer are identical. In this case, the map function only needs to be applied once.
This is a useful optimization because buffers with equals channels occur very often inside the audio graph.
E.g. `OscillatorNode -> GainNode -> DestinationNode`. The GainNode will typically already upmix the oscillator output to two channels. The gain will then be applied twice but this is what we can optimize out.

The same optimization could theoretically apply for `AudioBuffer` (e.g. the resample method) but it practise the buffer will never contain channels that are the same `Arc`. There is no user space function that could get to that situation.

orottier:
Some initial work at https://github.com/orottier/web-audio-api-rs/commit/92ba3f03d741a95ce2013929c8d3f7548cda1fbc - showing it might not always be worthwhile

b-ma:
Yup, maybe a bit convoluted if the improvement is small.

One thing I've though about in the same vein, would be to bypass processing when input is silent. I don't think it would work for all nodes, but for many of them it could be quite a simple improvement. I wanted to start a proof of concept on the `GainNode` but failed to reproduce your flamegraph results in #129 (I think we should see the 7% falling down here). The graph I obtain is very different both in terms of results and layout when I run only the granular bench. Could you share your code and the command you used, so that we are on the same page here?

orottier:
> Yup, maybe a bit convoluted if the improvement is small.

We should rewrite the biquad, iir, stream_destination and many other nodes to use this optimization and see if they benefit from it. My theory is that applying gain is so efficient (32 SIMD instruction on already cached data), that we're only seeing noise right now.

> bypass processing when input is silent. 

Dealing with silent input would be exiting, definitely. The spec has a whole section on it that we ignore currently!
https://www.w3.org/TR/webaudio/#AudioNode-actively-processing

> share your code and the command

I used https://github.com/flamegraph-rs/flamegraph on a cpu-optimized (dedicated) 2 CPU droplet on Digital Ocean
Change cargo.toml
```
[profile.release]
debug = true
```
Then run
`cargo flamegraph --example some_example`

The code I used was the minimal runnable version of the "Granular synthesis" example. I don't have the code anymore, maybe we can start a branch and check our numbers and graphs again?

b-ma:
[...]

PR #118, opened 2022-02-21
benchmarks - first draft

b-ma:
Hey,

A first attempt to port some of the benchmarks from [https://github.com/padenot/webaudio-benchmark](https://github.com/padenot/webaudio-benchmark). 
It's in a really early stage, quite dirty in term of interface (I've used `termion` too but with much less success :) and I don't really know if it should live in the repo neither.

Anyway that's quite interesting to have some numbers, feel very free to hack into it!


orottier:
Hey, really cool stuff. It will definitely be beneficial to have a benchmark suite early in the process. Not just for the conference submission but also to see how performance develops over time.
What do you think of the results? Better / worse than you had expected?

A few remarks:
- benchmarks should definitely live in the repo. Either next to the unit tests or in a `benches/` folder
- benches are full class citizens just like tests in rust, see https://doc.rust-lang.org/1.16.0/book/benchmark-tests.html
- you should definitely use the `Bencher` class. Compilers are weird so you never know what you are profiling in your setup. Have a look at the black-box too
- The termion setup is cool but way too complicated I think. Just create dedicated bench function per item you would like to profile. And if you want to validate the outputs, just write a similar unit test.
- We can use the criterion crate if we want to make it look real sweet: https://crates.io/crates/criterion
- You can also leverage macro's for common stuff, check e.g. https://github.com/BurntSushi/rust-csv/blob/master/benches/bench.rs

Also there are some really simple optimizations still on the table: #117 #119, plus we should really do some flamegraph style profiling to check the current bottlenecks. So don't draw any big conclusions from the current results!

b-ma:
Hey, cool you like it! 

> What do you think of the results? Better / worse than you had expected?

Actually both at the same time :) We have of course lot of room for improvement compared to Firefox or Chrome, but nothing to be ashamed neither I guess. And I have the impression it gives some hints / confirms where are our big bottlenecks.

Few remarks on your remarks, I completely agree that we should try to create more proper benchmarks (e.g. using `Bencher`) on specific points (last time I tried it was a `nightly` feature, did it changed?), however I think this approach is also interesting / complementary because:
- It really gives a bird eye view of running "almost real-world" code compared to micro-benchmarks
(I don't think we can run into some weird compiler optimisation here because the output buffer is actually computed just as it would in a real application).
- I find the idea of comparing the offline rendering against real time really clever and interesting, its true that it is very coarse (more an order of magnitude than a precise result) but that's also a very simple and meaningful number.
- As the code is really a direct translation from the JS code (which is still in use in the Firefox CI apparently), we can very simply compare the results directly with the ones from browsers in a way that doesn't seems too biased, i.e. as the only thing that is measured here is the `offline_context.start_rendering` and as the rendered buffers are quite long (2s), hopefully the JS overhead tends to be marginal so we are almost comparing directly with the C++ rendering code.

In any case, I really agree with you that this approach doesn't have to be only one of course. If you think we should merge that at some point, probably we should rename the file to avoid confusion with more proper benches of course.

b-ma:
> hopefully the JS overhead tends to be marginal

Just thinking a good way to confirm that would be to do the same with the NodeJS bindings, will try to do that too if I can find the time.

examples/benchmarks.rs
```diff
@@ -0,0 +1,382 @@
...
+    let mut current_source: Option<AudioBufferSourceNode> = None;
+
+    // that's really dirty
+    let mut bytes = stdin.bytes();
+    loop {
+        let b = bytes.next().unwrap().unwrap();
+
+        match b {
+            // quit
+            b'q' => return,
```
orottier, line 350: Could you also listen for ctrl-c here? Was a bit surprised I could not kill the program.
Other than that, lets merge!

b-ma:
ok cool, I'll try to improve user experience (sort of :) and port all the tests I can, except for the convolution I think we can port all of them.

I let you know when I finished that

examples/benchmarks.rs
b-ma: it's there!
I have also improved the interface and interaction, let me know what you think

b-ma:
Hey, so except for the convolution we have all the cases now!

Seems to me that our main bottlenecks (kind of expected I guess) for now are the `Graph` parsing and the `AudioParam` automations. These 3 tests are interesting to show that, here we have around 206 sources registered in the graph + a gain for each source in the 2 first tests:

```
- 14      | Synth (Sawtooth with Envelope)             | 5601          | 21.4x
- 15      | Synth (Sawtooth with gain - no automation) | 4678          | 25.7x
- 16      | Synth (Sawtooth without gain)              | 2713          | 44.2x
```

The granular synth with `1500 * (osc + gain)` nodes is a quite extreme example too:

```
- 13      | Granular synthesis                         | 16020         | 0.5x          
``` 

Seems that the `AudioBufferSourceNode` has also some room for improvement.

Note that the way the tests are made introduce a kind of artificial bias because all nodes are registered before hand (leading to very huge graphs) which would not happen in a real-time situation where you would schedule your sources and gains on the fly. 

But in any case I think that's really interesting to have these numbers, quite happy with that :)

orottier:
Cool stuff all in all. Let's use some flamegraph profiling to look for current bottlenecks.
I have created a `Performance` label for issues: https://github.com/orottier/web-audio-api-rs/labels/Performance

PR #121, opened 2022-02-23
Remove spatial mod from public interface

orottier:
AudioListener is now in the root namespace

Part of #76 - remove all functions that are not strictly part of the spec


b-ma:
All good to me!

Issue #119, opened 2022-02-22
Render AudioListener only when Panners are present


orottier:
The `AudioListener` and the 9 `AudioParam`s that feed into it are rendered always. This present a performance baseline hit when rendering simple stuff.
When there are no `PannerNode`s present in the audio graph, we can skip the rendering

PR #125, opened 2022-03-11
Track the number of active panners in the audio graph

orottier:
so we can skip rendering the AudioListener and its Params when there are none

This is one way to tackle #119 but it clutters the code base a bit. The audio graph needs to have knowledge of the AudioListener and its param. Previously the AudioGraph was blissfully unaware of the nature of the audio processors.

Alternative solution would be to lazily put the AudioListener in the context whenever the first PannerNode is added. That would make the code very clean, but when the Panner is dropped, this Listener would remain to render

I need to put a bit more thought in that discussion. In any case we can use this to benchmark again.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -76,6 +76,8 @@ struct ConcreteBaseAudioContextInner {
     queued_messages: Mutex<Vec<ControlMessage>>,
     /// number of frames played
     frames_played: Arc<AtomicU64>,
+    /// control msg to add the AudioListener, to be sent when the first panner is created
+    inactive_audio_listener: Mutex<Option<ControlMessage>>,
     /// AudioListener fields
     listener_params: Option<AudioListenerParams>,
 }
@@ -616,6 +618,7 @@ impl ConcreteBaseAudioContext {
             queued_messages: Mutex::new(Vec::new()),
             node_id_inc: AtomicU64::new(0),
             frames_played,
+            inactive_audio_listener: Mutex::new(None),
             listener_params: None,
         };
         let base = Self {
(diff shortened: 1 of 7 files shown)


orottier:
Yes, if you are adding a Panner at some point, you will probably use another one later. You either build immensive 2d/3d audio stuff, or not at all. I'll leave this open for now and will come up with the other implementation later

orottier:
Still in draft because somehow this new implementation does not bring the desired performance benefits

src/context.rs
```diff
@@ -76,6 +76,8 @@ struct ConcreteBaseAudioContextInner {
     queued_messages: Mutex<Vec<ControlMessage>>,
     /// number of frames played
     frames_played: Arc<AtomicU64>,
+    /// control msg to add the AudioListener, to be sent when the first panner is created
+    inactive_audio_listener: Mutex<Vec<ControlMessage>>,
```
b-ma, line 80: I find the naming a bit confusing, `audio_listener_messages`, `audio_listener_registration_messages`?

src/context.rs
```diff
@@ -813,6 +828,32 @@ impl ConcreteBaseAudioContext {
         self.connect(&AudioNodeId(LISTENER_NODE_ID), panner, 7, 8);
         self.connect(&AudioNodeId(LISTENER_NODE_ID), panner, 8, 9);
     }
+
+    /// Add the [`AudioListener`] to the audio graph (if not already)
+    pub(crate) fn release_audio_listener(&self) {
```
b-ma, line 833: same, I find `release` to be counter intuitive, `register`, `attach`, `connect` ?

b-ma:
Hey, appart the naming questions, look quite good to me

I ran the benchmarks again, it seems quite consistent with the previous version (maybe a bit slower, but I think we are in the noise)

Great !

orottier:
Thanks. I hope you like the new naming

b-ma:
great !

PR #128, opened 2022-03-21
Refactor/param

b-ma:
Hey, 

Finally found little time to finish that, this is mostly refactoring: 
- better delimiting a-rate processing (so it should avoid few unneeded computation for k-rate)
- added automation methods argument sanity checks


src/render/processor.rs
```diff
@@ -61,7 +61,14 @@ impl<'a> AudioParamValues<'a> {
...
-    /// For both A & K-rate params, it will provide a slice of length [`crate::RENDER_QUANTUM_SIZE`]
+    /// # Warning
+    ///
+    /// In the current implementation, the slice will always have a length of
+    /// [`RENDER_QUANTUM_SIZE`](crate::RENDER_QUANTUM_SIZE),
+    /// but one should not rely on this. Following the AudioWorklet specification
+    /// (cf. <https://www.w3.org/TR/webaudio/#audioworkletprocess-callback-parameters>),
+    /// a slice of length 1 may be returned in the future if all values in the
+    /// current quantum are identical (e.g. for k-rate processing).
     pub fn get(&self, index: &AudioParamId) -> &[f32] {
```
orottier, line 72: I think in our previous discussion on this signature, we concluded that is very convenient for the end user to always return a slice of size 128.
We could consider to drop the warning and commit to it. It would require a breaking change though then, to change it back but that is a risk I'm willing to make. For now I don't see much drawback in resizing the array to 128 always, this should translate to a memcpy which should be very efficient

src/param.rs
```diff
@@ -11,6 +11,49 @@ use crate::{AtomicF32, SampleRate, RENDER_QUANTUM_SIZE};
 
 use crossbeam_channel::{Receiver, Sender};
 
+// arguments sanity check functions for automation methods
+fn assert_non_negative(value: f64, method_name: &str, arg_name: &str) {
```
orottier, line 15: Maybe add a `#[track_caller]` attribute?

orottier: I will send a PR soon to enable that attribute on our other assertions, and to assert channel counts in more places

orottier:
Looking good btw!

src/render/processor.rs
b-ma: yup you are right, reverted back to the previous doc

b-ma:
> Would be very nice to hook the node bindings on these and run the official test suite

Sure good idea, if I remember well except for `set_value_curve_at_time` I think all the bindings are already there.
I will try to have a look, see what it costs

orottier:
Good to hear! I think these automation events are particularly tricky to get right in all cases, so that would be a perfect start for the testing harness.

PR #130, opened 2022-03-28
Improve ChannelCount/Mode/Interpretation

orottier:
This turned out bigger than expected, I was having a look at
`Can we drop AudioNode's channel_config_raw and channel_config_cloned?`
From #76 

I will annotate with some comments what is going on exactly

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -8,7 +8,7 @@
 )]
 
 use std::ops::Range;
-use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex};
 
 // magic node values
@@ -22,7 +22,7 @@ const LISTENER_PARAM_IDS: Range<u64> = 2..11;
 use crate::buffer::{AudioBuffer, AudioBufferOptions};
 use crate::media::{MediaDecoder, MediaStream};
 use crate::message::ControlMessage;
-use crate::node::{self, AudioNode, ChannelConfigOptions};
+use crate::node::{self, AudioNode, ChannelConfig, ChannelConfigOptions};
 use crate::param::{AudioParam, AudioParamDescriptor, AudioParamEvent};
 use crate::periodic_wave::{PeriodicWave, PeriodicWaveOptions};
 use crate::render::{AudioProcessor, NodeIndex, RenderThread};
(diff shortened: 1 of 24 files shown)


src/node/mod.rs
```diff
@@ -183,10 +217,7 @@ pub trait AudioNode {
     fn id(&self) -> &AudioNodeId {
         self.registration().id()
     }
-    fn channel_config_raw(&self) -> &ChannelConfig;
```
orottier, line 186: Here's the original motivation of the PR:
- drop channel_config_raw
- drop channel_config_cloned
- only leave channel_config

While still allowing for users to apply constraints on the channel config values

src/node/destination.rs
```diff
@@ -59,32 +50,51 @@ impl AudioNode for AudioDestinationNode {
...
+        if v > self.max_channels_count() as usize {
+            panic!(
+                "IndexSizeError: channel count cannot be greater than maxChannelCount ({})",
+                self.max_channels_count()
+            );
+        }
+        AudioNode::set_channel_count(self, v);
     }
-
-    fn channel_count(&self) -> usize {
```
orottier, line 70: The AudioDestinationNode can now change its desired number of output channels (only online AudioContext, not offline)

src/context.rs
```diff
@@ -80,6 +82,8 @@ struct ConcreteBaseAudioContextInner {
     queued_audio_listener_msgs: Mutex<Vec<ControlMessage>>,
     /// AudioListener fields
     listener_params: Option<AudioListenerParams>,
+    /// Denotes if this AudioContext is offline or not
+    offline: bool,
```
orottier, line 86: We need to know this since the AudioDestinationNode has different behaviour for online and offline when changing channel counts

orottier:
Hold on, the following program now gives a stack overflow. I need to investigate 
```rust
    // Create an audio context (default: stereo);
    // Create an oscillator node with sine (default) type
    let osc = context.create_oscillator();
    // Connect osc to the destination node which is the default output device
    osc.connect(&context.destination());
    // Oscillator needs to be started explicitily
    osc.start();
    // Play for 2 seconds
    std::thread::sleep(std::time::Duration::from_secs(2));
    // Switch to mono
    context.destination().set_channel_count(1);
    // Play for 2 seconds
    std::thread::sleep(std::time::Duration::from_secs(2));
```

orottier:
Sorry, there's multiple issues still. Will update tomorrow

src/context.rs
```diff
@@ -635,16 +635,16 @@ impl ConcreteBaseAudioContext {
         render_channel: Sender<ControlMessage>,
         offline: bool,
     ) -> Self {
-        // setup destination node with preferred number of channels
-        let destination_channel_count = Arc::new(AtomicUsize::new(number_of_channels as usize));
+        // placeholder for destination node channel count, will patch later
+        let tmp_dest_channel_count = Arc::default();
```
orottier, line 639: I'm not too happy with the complexity of bootstrapping the audio context. I will rewrite this after the vq release

PR #133, opened 2022-03-31
AudioContext constructor

orottier:


src/context/online.rs
```diff
@@ -65,23 +75,22 @@ impl BaseAudioContext for AudioContext {
...
 
 impl AudioContext {
     /// Creates and returns a new `AudioContext` object.
     /// This will play live audio on the default output
-    // options is passed by value to be conform to the specification interface
     #[allow(clippy::needless_pass_by_value)]
     #[cfg(not(test))]
     #[must_use]
-    pub fn new(options: Option<AudioContextOptions>) -> Self {
+    pub fn new(options: AudioContextOptions) -> Self {
```
orottier, line 88: This is the big change

orottier:
You are right. I think I initially added that option because I did not realize you can update the channel configuration of the destination node. That was recently fixed in #130 so I will remove it from the constructor now

src/context/online.rs
```diff
@@ -30,18 +30,26 @@ pub enum AudioContextLatencyCategory {
...
+/// All fields are optional and will default to the value best suited for interactive playback on
+/// your hardware configuration.
+///
+/// Check the documentation of the [`AudioContext` constructor](AudioContext::new) for usage
+/// instructions.
+#[derive(Clone, Debug, Default)]
 pub struct AudioContextOptions {
     /// Identify the type of playback, which affects
     /// tradeoffs between audio output latency and power consumption
     pub latency_hint: Option<AudioContextLatencyCategory>,
```
b-ma, line 50: Do we need the `Option` here, as it defaults to `interactive`? maybe that would be more consistent to with the rules we followed for the nodes https://github.com/orottier/web-audio-api-rs/pull/103

```
(AudioContextLatencyCategory or double) latencyHint = "interactive";
```

```rs
impl Default for AudioContextOptions {
  fn default() -> Self {
     Self {
       latency_hint: AudioContextLatencyCategory::Interactive,
       sample_rate: None,
     }
  }
}
```

(Actually we probably don't even need to implement as it already seems to be the default)

orottier:
I have applied your suggestion, thanks!

PR #134, opened 2022-04-04
Remove AudioNodeId from public interface

orottier:
Delegate all access via the AudioContextRegistration.

While working on #76 I realized we don't need to expose the `AudioNodeId` in the public interface.
Now, there's only `AudioParamId` and `AudioContextRegistration` inside the context mod that are not part of the API. These two however cannot be dropped because we need them for manual AudioNode construction (audio worklets)

### DIFF
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -35,7 +35,7 @@ const LISTENER_PARAM_IDS: Range<u64> = 2..11;
 ///
 /// Used for internal bookkeeping.
 #[derive(Debug)]
-pub struct AudioNodeId(u64);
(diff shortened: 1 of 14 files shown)


src/context/mod.rs
```diff
@@ -35,7 +35,7 @@ const LISTENER_PARAM_IDS: Range<u64> = 2..11;
 ///
 /// Used for internal bookkeeping.
 #[derive(Debug)]
-pub struct AudioNodeId(u64);
+pub(crate) struct AudioNodeId(u64);
```
orottier, line 38: Here's the main change

b-ma:
nice !

Issue #51, opened 2021-11-15
Rewrite `CpalBackend::build_input` to make it similar to `build_output`


Jerboas86:
Connection to the audio hardware is made through `io::build_output` and `io::build_input` functions.
- `build_output` instantiates the output stream.
- `build_input` instantiates the input stream.

A rewrite of `io::build_output` has been made by PR #36. This PR makes it possible for the end user to change output stream parameters like:
- number of output channels
- latency settings
- sample rate
But this PR stopped there, and the work has not been done for `build_input` function

A rewrite of `io::build_input` should be made to make stream configuration coherent between `build_output` and `build_input` functions. It would also be helpful if both functions used similar type system to define them.

orottier:
Removing the v1 milestone because this is not part of the public API.
The `Microphone` struct however is, and could benefit from the same config settings we'd like to pass to the `io` mod. Let's fix that after the v1 release (and introduce a `Microphone::with_config` constructor)

orottier:
Fixed in #264

Issue #101, opened 2022-01-22
Implement `BaseAudioContext.state`


orottier:
https://www.w3.org/TR/webaudio/#dom-baseaudiocontext-control-thread-state-slot

Before we think about also adding onStateChange handlers ( https://www.w3.org/TR/webaudio/#dom-baseaudiocontext-onstatechange ) we need to think about the architecture for event handling (probably in a separate issue)

orottier:
Once this is implemented, we can fix the panic that occurs when you add nodes to a closed context, e.g.
```
let context = AudioContext::new(None);
context.close();
let delay = context.create_delay(1.);
```
The panic occurs because we attempt to use the closed message channel to the render thread.

As per discussion in https://github.com/WebAudio/web-audio-api/issues/1580, creating nodes on a closed context should not throw an exception (but will also not do anything useful)

orottier:
Another panic that can be fixed is

```rust
let context = AudioContext::new(None);
let analyser = context.create_analyser();
context.close();
analyser.get_float_time_domain_data(vec![1.; 128]);
```

This should just return zeroes when the context is paused/closed

PR #136, opened 2022-04-05
Implement input options

ork821:
Partial implement of #51

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -17,12 +17,12 @@ use crate::{SampleRate, RENDER_QUANTUM_SIZE};
 
 use cpal::{
     traits::{DeviceTrait, HostTrait, StreamTrait},
-    BuildStreamError, Device, SampleFormat, Stream, StreamConfig, SupportedBufferSize,
+    BuildStreamError, Device, SampleFormat, Stream, StreamConfig, SupportedBufferSize, SampleRate as CpalSampleRate
 };
 
 use crate::buffer::AudioBuffer;
 use crate::context::{AudioContextLatencyCategory, AudioContextOptions};
-use crate::media::MicrophoneRender;
+use crate::media::{MicrophoneRender, MicrophoneInitOptions};
 use crate::render::RenderThread;
 
 use crossbeam_channel::{Receiver, Sender};
@@ -393,7 +393,7 @@ pub(crate) fn build_output(
 }
 
 /// Builds the input
-pub fn build_input() -> (Stream, StreamConfig, Receiver<AudioBuffer>) {
+pub fn build_input(options: MicrophoneInitOptions) -> (Stream, StreamConfig, Receiver<AudioBuffer>) {
     let host = cpal::default_host();
     let device = host
         .default_input_device()
@@ -403,10 +403,17 @@ pub fn build_input() -> (Stream, StreamConfig, Receiver<AudioBuffer>) {
     let mut supported_configs_range = device
         .supported_input_configs()
         .expect("error while querying configs");
-    let supported_config = supported_configs_range
-        .next()
-        .expect("no supported config?!")
-        .with_max_sample_rate();
+
+    let supported_config = match options.sample_rate {
+        Some(sample_rate)=> supported_configs_range
+            .next()
+            .expect("no supported config?!")
+            .with_sample_rate(CpalSampleRate(sample_rate)),
+        None => supported_configs_range
+            .next()
+            .expect("no supported config?!")
+            .with_max_sample_rate()
+    };
 
     let sample_format = supported_config.sample_format();
 
--- a/src/media/mic.rs
+++ b/src/media/mic.rs
@@ -30,12 +30,14 @@ use crossbeam_channel::{Receiver, TryRecvError};
 ///
 /// ```no_run
 /// use web_audio_api::context::{BaseAudioContext, AudioContext};
-/// use web_audio_api::media::Microphone;
+/// use web_audio_api::media::{Microphone, MicrophoneInitOptions};
 /// use web_audio_api::node::AudioNode;
 ///
 /// let context = AudioContext::default();
 ///
-/// let stream = Microphone::new();
+/// let stream = Microphone::new(MicrophoneInitOptions {sample_rate: Some(48000)});
+/// // or you can create Microphone with default options
+/// let stream = Microphone::default();
 /// // register as media element in the audio context
 /// let background = context.create_media_stream_source(stream);
 /// // connect the node directly to the destination node (speakers)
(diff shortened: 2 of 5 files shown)


src/media/mic.rs
```diff
@@ -107,7 +109,7 @@ impl Microphone {
 #[cfg(not(test))]
 impl Default for Microphone {
     fn default() -> Self {
-        Self::new()
+        Self::new(MicrophoneInitOptions {sample_rate: None})
```
orottier, line 112: I would prefer
`Self::new(MicrophoneInitOptions::default())`

src/media/mic.rs
```diff
@@ -142,6 +144,15 @@ impl Iterator for Microphone {
     }
 }
 
+/// Specify the configuration for the [`Microphone`] constructor.
+///
+/// Field is optional and will be set to hardcoded value.
+#[derive(Clone, Debug, Default)]
+pub struct MicrophoneInitOptions {
```
orottier, line 151: Maybe the name `MicrophoneOptions` would be more consistent with `AudioContextOptions`

src/media/mic.rs
orottier, line 112: I would prefer
`Self::new(MicrophoneInitOptions::default())`

src/io.rs
```diff
@@ -403,10 +403,17 @@ pub fn build_input() -> (Stream, StreamConfig, Receiver<AudioBuffer>) {
...
-    let supported_config = supported_configs_range
-        .next()
-        .expect("no supported config?!")
-        .with_max_sample_rate();
+
+    let supported_config = match options.sample_rate {
+        Some(sample_rate)=> supported_configs_range
+            .next()
+            .expect("no supported config?!")
+            .with_sample_rate(CpalSampleRate(sample_rate)),
```
orottier, line 411: Requesting a specific sample rate should be done after determining the fallback config.
The fallback config is put in the variable `default_config` and should only contain values that were actually supplied by cpal, so they should never fail.
Could you try to set the requested sample rate only on the `config` settings?

orottier (changes requested):
Hi @ork821, thanks for contributing! Could you have a look at my comments?
Are you planning to add `latency_hint` too in this PR?

ork821:
> Hi @ork821, thanks for contributing! Could you have a look at my comments? Also the CI checks fail, could you run `cargo fmt` and commit those changes too? Are you planning to add `latency_hint` too in this PR?

src/io.rs
```diff
@@ -425,6 +427,10 @@ pub fn build_input() -> (Stream, StreamConfig, Receiver<AudioBuffer>) {
 
     let mut config: StreamConfig = supported_config.into();
     config.buffer_size = cpal::BufferSize::Fixed(input_buffer_size);
+    if options.sample_rate.is_some() {
```
orottier, line 430: This can be written better as:

```rust
if let Some(sample_rate) = options.sample_rate {
     // do stuff with sample_rate
}
```

This avoids the `panic` path of the `unwrap` call. See also https://doc.rust-lang.org/rust-by-example/flow_control/if_let.html

orottier:
This is going in the right direction :)

I left one nitpick on your code, please have a look.

Also, the build is still failing, you can check it locally with `cargo clippy --all-targets`
It says: `error: this argument is passed by value, but not consumed in the function body` which I think is a false positive and can be ignored with `#[allow(clippy::needless_pass_by_value)]` similar to https://github.com/orottier/web-audio-api-rs/blob/main/src/context/online.rs#L103

After that we can merge and leave the latency for another time

orottier:
If you'd like I can get you started with some other issues, maybe #44 or #101

ork821:
> This is going in the right direction :)
> 
> I left one nitpick on your code, please have a look.

I fixed this. Thank you for your help. This is very important for my skills.

> Also, the build is still failing, you can check it locally with `cargo clippy --all-targets` It says: `error: this argument is passed by value, but not consumed in the function body` which I think is a false positive and can be ignored with `#[allow(clippy::needless_pass_by_value)]` similar to https://github.com/orottier/web-audio-api-rs/blob/main/src/context/online.rs#L103

Also fixed this. `cargo build`, `cargo test` and `cargo clippy --all-targets` runs without warnings.

As for other issues, maybe I would try #101. Or will find another issue to fix :)

orottier:
Great. You forgot about `cargo fmt` though, but I fixed it for you + an unrelated fix because rust 1.60 was released yesterday.

orottier:
@ork821 Since we are nearing a v1 release we need to take care of API stability.
Hence I added the latency_hint option for the mic too (but it is left unimplemented). We can fix this within the v1 version
Please have a look at my latest commit :)

ork821:
> @ork821 Since we are nearing a v1 release we need to take care of API stability. Hence I added the latency_hint option for the mic too (but it is left unimplemented). We can fix this within the v1 version Please have a look at my latest commit :)

Hi, @orottier. Thank you for fixing my small `cargo fmt` fault. 

I am not sure that usage same struct `AudioContextOptions` is good desition. 
For example, according to https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/AudioContext we can have only `latencyHint` and `sampleRate`. But according to https://w3c.github.io/mediacapture-main/#dom-mediatrackconstraintset we can have `echoCancellation`, `channelCount`, `deviceId` (and so on) as input device options. If for now it's ok, we can merge.

orottier:
That's a valid point, however I would prefer to keep it this way, because either way we need breaking changes when we add the features that you mention to `AudioInputOptions`.
https://doc.rust-lang.org/cargo/reference/semver.html#struct-add-public-field-when-no-private

Let's op for a new constructor in the future
`fn new_with_config(audio_opts: AudioContextOptions, mic_opts: MicrophoneOptions) -> Microphone`
We can add all mic specific fields to `MicrophoneOptions`. The new constructor is not a breaking change so can be added within the v1 version. This gives us a bit of time to flesh out the complete featureset around mic input

orottier (approved):

ork821:
> That's a valid point, however I would prefer to keep it this way, because either way we need breaking changes when we add the features that you mention to `AudioInputOptions`. https://doc.rust-lang.org/cargo/reference/semver.html#struct-add-public-field-when-no-private
> 
> Let's op for a new constructor in the future `fn new_with_config(audio_opts: AudioContextOptions, mic_opts: MicrophoneOptions) -> Microphone` We can add all mic specific fields to `MicrophoneOptions`. The new constructor is not a breaking change so can be added within the v1 version. This gives us a bit of time to flesh out the complete featureset around mic input

Yes, you are actually right about breaking changes. It would be better to implement mentioned method later if we need it. I think you can merge if everything is okay.

PR #137, opened 2022-04-06
Toy WebRTC example 

orottier:
As promised, part of #77 , a very crude VOIP app.
It nicely shows the media steam source and destination functionality.
It could be improved in many ways (add buffering, smoothing, encryption, connection bootstrapping etc) but the base is solid

### DIFF
--- a/examples/toy_webrtc.rs
+++ b/examples/toy_webrtc.rs
@@ -1,3 +1,22 @@
+/*
+ * This example features a toy VOIP app
+ *
+ * Run `cargo run --release --example toy_webrtc -- server` to run the echo server.
+ * The echo server simply receives the audio packets and ships them back.
+ *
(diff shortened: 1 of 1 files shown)


b-ma:
Works like a charm :), really nice how simple the code actually looks

examples/toy_webrtc.rs
```diff
@@ -0,0 +1,188 @@
...
+                };
+                return Some(Ok(AudioBuffer::new(options)));
+            }
+            Err(e) => panic!("client recv IO error: {}", e),
+        };
+        Some(Ok(deserialize(bytes, self.sample_rate)))
+    }
+}
+
+/// The client records their audio, ships it to the server and plays the received samples
```
b-ma, line 144: "The client 1. records the audio and ships it to the server, and 2. plays back the samples received from the server"

PR #139, opened 2022-04-07
Fix #126: allow microphone to pause/resume/close

### LINKED ISSUE #126: Unable to stop microphone processing
The `Microphone` struct has pause/resume functionality (and will stop when dropped).
But how we intend to use it,

```rust
let stream = Microphone::new();
// register as media element in the audio context
let background = context.create_media_stream_source(stream);
// connect the node to the destination node (speakers)
background.connect(&context.destination());
```

ownership is given to the MediaStreamAudioSourceNode, which cannot be stopped.
So even when you disconnect the node, it will never drop as it will have tailTime true (this is contrary to the spec actually, I should dig into that).

A solution would be similar how we treat MediaStreamAudioDestinationNode:

```rust
let mic = Microphone::new();
let stream = mic.stream();
// register as media element in the audio context
let background = context.create_media_stream_source(stream);
// connect the node to the destination node (speakers)
background.connect(&context.destination());

// mic is still available now, so can be stopped/paused
// when mic Drops, it will stop automatically
```

This will have some caveats
- users should not Drop the Mic when playback is still desired
- you should not call `.stream()` multiple times
- it's quite different from the JS equivalent (it was already).

orottier:
Fixes #126 

Quite happy with the setup, it looks much more similar to https://docs.rs/web-audio-api/0.17.0/web_audio_api/node/struct.MediaStreamAudioDestinationNode.html#method.stream now

### DIFF
--- a/src/media/mic.rs
+++ b/src/media/mic.rs
@@ -7,7 +7,7 @@ use crate::{SampleRate, RENDER_QUANTUM_SIZE};
 #[cfg(not(test))]
 use crossbeam_channel::Sender;
 #[cfg(not(test))]
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 #[cfg(not(test))]
 use crate::buffer::ChannelData;
@@ -23,10 +23,10 @@ use crossbeam_channel::{Receiver, TryRecvError};
 mod private {
     use super::*;
 
-    pub struct StreamHolder(Arc<Stream>);
+    pub struct StreamHolder(Arc<Mutex<Option<Stream>>>);
 
     impl StreamHolder {
-        pub fn new(stream: Arc<Stream>) -> Self {
+        pub fn new(stream: Arc<Mutex<Option<Stream>>>) -> Self {
             StreamHolder(stream)
         }
     }
@@ -42,8 +42,12 @@ use private::StreamHolder;
 
 /// Microphone input stream
 ///
-/// It implements the [`MediaStream`](crate::media::MediaStream) trait so can be used inside a
-/// [`MediaStreamAudioSourceNode`](crate::node::MediaStreamAudioSourceNode)
+/// The Microphone can set up a [`MediaStream`](crate::media::MediaStream) value which can be used
+/// inside a [`MediaStreamAudioSourceNode`](crate::node::MediaStreamAudioSourceNode).
+///
+/// It is okay for the Microphone struct to go out of scope, any corresponding stream will still be
+/// kept alive and emit audio buffers. Call the `close()` method if you want to stop the microphone
+/// input and release all system resources.
 ///
 /// # Warning
 ///
@@ -74,7 +78,7 @@ pub struct Microphone {
     sample_rate: SampleRate,
 
     #[cfg(not(test))]
-    stream: Arc<Stream>,
+    stream: Arc<Mutex<Option<Stream>>>,
 }
 
 impl Microphone {
@@ -89,7 +93,7 @@ impl Microphone {
 
         // shared ownership for the stream, because the Microphone is allowed to go out of scope
         // but all corresponding streams should still yield output
-        let stream = Arc::new(stream);
+        let stream = Arc::new(Mutex::new(Some(stream)));
 
         Self {
             receiver,
@@ -110,7 +114,9 @@ impl Microphone {
     /// * For a `BackendSpecificError`
     pub fn suspend(&self) {
         #[cfg(not(test))] // in tests, do not set up a cpal Stream
-        self.stream.pause().unwrap()
+        if let Some(stream) = self.stream.lock().unwrap().as_ref() {
+            stream.pause().unwrap()
+        }
     }
 
     /// Resumes the input stream that has previously been suspended/paused.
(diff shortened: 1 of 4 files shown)


src/media/mic.rs
```diff
@@ -16,10 +19,35 @@ use cpal::{traits::StreamTrait, Sample, Stream};
 
 use crossbeam_channel::{Receiver, TryRecvError};
 
+#[cfg(not(test))]
+mod private {
```
orottier, line 23: I'm reasonably sure this setup is safe.
cpal tells us a `Stream` cannot be moved across threads (it is not `Send`) due to some platform constraints. However, I want to move an `Arc<....<Stream>>` into the render thread so the control thread Microphone object can go out of scope while the Stream should continue to run. This private mod prevents any calls (e.g. pause/resume) on the stream so this should make it safe to `Send`

ork821:
> @ork821 if you'd like to help out, could you check if `RUST_LOG=debug cargo run --release --example microphone` works fine on your hardware? Use headphones to prevent feedback loops, start always with extremely low volume to protect your ears in case of bugs
Hi, @orottier! I just execute example. It works fine. My ears are ok, thank you for the warning.

If you are interested, my debug output is:
[2022-04-07T19:18:30Z INFO  web_audio_api::io] Host: Alsa
[2022-04-07T19:18:30Z INFO  web_audio_api::io] Output device: Ok("default")
[2022-04-07T19:18:30Z INFO  web_audio_api::render::thread] Audio render thread has been dropped
[2022-04-07T19:18:30Z WARN  web_audio_api::io] Output stream build failed with prefered config: A backend-specific error has occurred: ALSA function 'snd_pcm_hw_params_set_buffer_size' failed with error 'EINVAL: Invalid argument'
[2022-04-07T19:18:30Z INFO  web_audio_api::io] Input device: Ok("default")
[2022-04-07T19:18:30Z DEBUG web_audio_api::media::mic] Microphone input has been dropped
[2022-04-07T19:18:30Z WARN  web_audio_api::io] Output stream failed to build: BackendSpecific { err: BackendSpecificError { description: "ALSA function 'snd_pcm_hw_params_set_buffer_size' failed with error 'EINVAL: Invalid argument'" } }, retry with default config StreamConfig { channels: 1, sample_rate: SampleRate(384000), buffer_size: Default }
[2022-04-07T19:18:30Z DEBUG web_audio_api::media::mic] Input StreamConfig { channels: 1, sample_rate: SampleRate(384000), buffer_size: Fixed(128) }
Playback for 2 seconds
Pause mic for 2 seconds
Resume mic for 2 seconds
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
[2022-04-07T19:18:34Z DEBUG web_audio_api::media::mic] input frame dropped
Close mic - halting stream
[2022-04-07T19:18:36Z DEBUG web_audio_api::media::mic] Microphone input has been dropped
[2022-04-07T19:18:36Z DEBUG web_audio_api::node] Stream finished
[2022-04-07T19:18:38Z INFO  web_audio_api::render::thread] Audio render thread has been dropped

orottier:
Thanks @ork821, the output looks good.
However your device seems to be affected by #28 too (which is not critical). I will have a look soon

PR #141, opened 2022-04-09
Remove `context.channels()` method, not part of the spec

orottier:
And in fact, the value denotes the max available channels. I have changed the naming and comments accordingly


b-ma:
Except my small comment about the doc (I took the sentence from the spec which is more clear I think), all good for me !

orottier:
> Except my small comments (for the doc I adapted the sentence from the spec which is more clear I think), all good for me !

Thanks, but I think you need to hit the "finish review" button because I cannot see it :)

b-ma:
Better like that?

Issue #100, opened 2022-01-22
Implement `resume` and `suspend` on `OfflineAudioContext`


orottier:
https://www.w3.org/TR/webaudio/#OfflineAudioContext-methods

Notice the `suspend` function has the rather odd feature that you can specify the suspend time. We can implement that (later) as an addiditonal method `suspend_at`

b-ma:
> [disclaimer] Not related to `OfflineAudioContext` and really not sure about that but,

I wonder if we shouldn't have the context `suspended` by default.

From what I understand from the spec, that's not mandatory (https://webaudio.github.io/web-audio-api/#allowed-to-start) because the context can be somehow `allowed to start` (mainly for not breaking early demos and web page I think). But I'm not sure that gives a good hint when you are used to the JS API, where you have to do (except in some magical conditions I've never understood and that may differ between browsers):

```js
const context = new AudioContext();
$someButton.addEventListener('click', async () => {
   await context.resume();
   // now it's safe to use the `context`
});
```

I say that because, I was basically thinking that `resume`, `suspend` and `close` where just not fully implemented or working, when I realized that I didn't need to `resume` first (and true... I didn't took any time to test anything there :)

orottier:
Fixed with #412

PR #142, opened 2022-04-09
Feature: Implement base context state

ork821:
Hi, @orottier. I just started working on #101. 
What do you think about my way of implementation? 
I added `AudioContextState` and support changing it on `suspend`, `resume` and `stop` methods of `online` AudioContext.
Also I particularly support `do nothing` for source nodes when context is closed.

My main question is how to use `ConcreteBaseAudioContext` inside Renderers. In my implementation I used references, but not sure it's a good decision. May we use `Arc<>` or something like that? 

P.S. Don't worry about issues and warnings. I will fix them all after draft PR


orottier:
Hey, very good you are asking for early feedback because I'm seeing some exotic compiler errors occuring and I think we can easily avoid them. I will write some comments inline

src/node/audio_buffer_source.rs
```diff
@@ -323,17 +324,18 @@ impl Default for AudioBufferRendererState {
     }
 }
 
-struct AudioBufferSourceRenderer {
+struct AudioBufferSourceRenderer<'a> {
```
orottier, line 327: You are entering a world of pain by putting a reference with limited lifetime inside the renderers (the objects that implement `AudioProcessor`).

The audio processors are shipped to the render thread, and that thread may live forever so the compiler will complain that the context reference may not live long enough (hence the hints to `'static` in the compiler error)

In typical rust code, you should just make a copy or use shared ownership (via `Arc`) in these situations.

However, for us we don't need any of this. Whenever we close the context, the cpal `Stream` is dropped which closes the entire `RenderThread` object. The `RenderThread` has ownership over the `Graph`, and the `Graph` owns the render objects. Therefore, all these changes are not necessary.

You can check if it works by just calling a `println!(..)` in an Oscillator, and than closing the stream. You will notice it stops printing because the cpal stream is closed

src/context/concrete_base.rs
```diff
@@ -96,6 +98,7 @@ impl ConcreteBaseAudioContext {
             queued_audio_listener_msgs: Mutex::new(Vec::new()),
             listener_params: None,
             offline,
+            state: AudioContextState::SUSPENDED,
```
orottier, line 101: Our context starts running immediately, so this should be `Running`

Technically, the spec requires to start as `Suspended` and only switch to `Running` when the audio render thread is active ( https://www.w3.org/TR/webaudio/#AudioContext-constructors ), but we have not (yet) implemented that. We should probably create a ticket for that or pick it up in this PR

orottier (changes requested):
Thanks for your work! Please have a look at my comments, we can skip a lot of complexity.

Then, could you whip up an example file which showcases the pause, resume and suspend operations and prints the state?

After that, I will guide you to make the `close()` call operate on `&self` (not requiring `mut`).

src/context/concrete_base.rs
orottier: I realise the OfflineAudioContext is Suspended by default, until you run `start_rendering`. So we need different behaviours

src/context/mod.rs
```diff
@@ -49,6 +49,19 @@ impl From<&AudioParamId> for NodeIndex {
     }
 }
 
+/// Describes the current state of the BaseAudioContext
+#[derive(Debug, Copy, Clone)]
+pub enum AudioContextState {
```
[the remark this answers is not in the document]
ork821: Surprisingly, but not `cargo clippy` and not `cargo fmt` told me this. But I explored https://doc.rust-lang.org/1.0.0/style/style/naming/README.html and fix names.

src/context/concrete_base.rs
ork821: In my first implementation, I initialized state with `Running`, but then I explored spec and found `Suspended`. I think start with `Running` is okay.
As for `OfflineAudioContext`, there is no doubt that state is `Suspended` by default and change to `Running` after `start_rendering`

src/context/online.rs
```diff
@@ -163,12 +163,13 @@ impl AudioContext {
     /// * For a `BackendSpecificError`
     // false positive due to #[cfg(not(test))]
     #[allow(clippy::missing_const_for_fn, clippy::unused_self)]
-    pub fn suspend_sync(&self) {
+    pub fn suspend_sync(&mut self) {
```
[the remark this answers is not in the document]
ork821: Forget about breaking changes. So I followed your guide and add `state` field in the `AudioContext` and make it `Mutex`.

ork821:
> Thanks for your work! Please have a look at my comments, we can skip a lot of complexity.
> 
> Then, could you whip up an example file which showcases the pause, resume and suspend operations and prints the state?
> 
> After that, I will guide you to make the `close()` call operate on `&self` (not requiring `mut`).

Hi, @orottier. Thank you for comments and help. 
I just fixed all issues according to your guides. Also I added example file with usage and logging state before and after any operation. Please, check my commits.

examples/audiocontext.rs
```diff
@@ -0,0 +1,36 @@
+use web_audio_api::context::{AudioContext, BaseAudioContext};
```
orottier, line 1: Let's name this file `examples/change_state.rs`

src/context/mod.rs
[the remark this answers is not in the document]
orottier: That's strange, good to know..

src/context/offline.rs
```diff
@@ -64,13 +66,15 @@ impl OfflineAudioContext {
...
+            state: Mutex::new(AudioContextState::Suspended),
         }
     }
 
     /// Given the current connections and scheduled changes, starts rendering audio.
     ///
     /// This function will block the current thread and returns the rendered `AudioBuffer`
     /// synchronously. An async version is currently not implemented.
+    #[allow(clippy::missing_panics_doc)]
     pub fn start_rendering_sync(&mut self) -> AudioBuffer {
```
orottier, line 78: Because we only have an implementation of `start_rendering` that blocks the current thread (hence, `_sync`).
The state after this function is finished is actually `Suspended` again. Or Closed? I'm not actually sure if you should be able to call this function multiple times..

orottier:
Now this is starting to look good. There's one pesky detail still, the `state()` method should live on the `BaseAudioContext` trait (and thus on the `ConcreteBaseAudioContext` struct), not on the `AudioContext` or `OfflineAudioContext`).
Can you move the `state: Mutex<AudioContextState>,` attribute to the  `ConcreteBaseAudioContext`? 

While you are at it, we can change it to a `state: AtomicU32` instead because we do not need a Mutex to change an enum in a thread safe way. Have a look at https://github.com/orottier/web-audio-api-rs/blob/main/src/node/oscillator.rs#L124 for inspiration!

ork821:
> Now this is starting to look good. There's one pesky detail still, the `state()` method should live on the `BaseAudioContext` trait (and thus on the `ConcreteBaseAudioContext` struct), not on the `AudioContext` or `OfflineAudioContext`). Can you move the `state: Mutex<AudioContextState>,` attribute to the `ConcreteBaseAudioContext`?
> 
> While you are at it, we can change it to a `state: AtomicU32` instead because we do not need a Mutex to change an enum in a thread safe way. Have a look at https://github.com/orottier/web-audio-api-rs/blob/main/src/node/oscillator.rs#L124 for inspiration!

I fixed everything we talked about. But I used `AtomicU8`, because I think it's enough for state. Also I don't implement state in `OfflineAudioContext`, because I'm not clearly sure about how correctly should state change and I see that it hasn't `suspend` function, so maybe we should support state change after #100?

orottier:
Great @ork821 ! I took the liberty to push some final polishing, please have a good look at
https://github.com/orottier/web-audio-api-rs/pull/142/commits/0d9290e101a830d790dde463e32894eede7e163a

I will merge this now. 
I you are interested in contributing more, the issue #142 contains two comments around panics that can be fixed now. Maybe you can add those cases to the example file and have a good look at the stack traces. The panics arise from `crossbeam` channels that are no longer alive. We can fix that by checking the state before calling `send` on them.

orottier (approved):

PR #145, opened 2022-04-10
Add currentFrame to AudioProcess.process call

orottier:
We now support all three attributes from
https://www.w3.org/TR/webaudio/#AudioWorkletGlobalScope-attributes

### DIFF
--- a/src/render/processor.rs
+++ b/src/render/processor.rs
@@ -14,7 +14,7 @@ use super::{graph::Node, AudioRenderQuantum, NodeIndex};
 /// This struct currently only contains information about the progress of time. In a future
 /// version, it should be possible to add arbitrary data. For example, multiple processors might
 /// share a buffer defining a wavetable or an impulse response.
-pub struct GlobalScope {
+pub struct Scope {
     pub current_frame: u64,
     pub current_time: f64,
     pub sample_rate: SampleRate,
@@ -50,7 +50,7 @@ pub trait AudioProcessor: Send {
         inputs: &[AudioRenderQuantum],
         outputs: &mut [AudioRenderQuantum],
         params: AudioParamValues,
-        global_scope: GlobalScope,
+        global_scope: Scope,
     ) -> bool;
 }
 
--- a/src/render/processor.rs
+++ b/src/render/processor.rs
@@ -14,7 +14,7 @@ use super::{graph::Node, AudioRenderQuantum, NodeIndex};
 /// This struct currently only contains information about the progress of time. In a future
 /// version, it should be possible to add arbitrary data. For example, multiple processors might
 /// share a buffer defining a wavetable or an impulse response.
-pub struct Scope {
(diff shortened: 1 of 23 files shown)


src/render/processor.rs
```diff
@@ -38,7 +38,8 @@ pub trait AudioProcessor: Send {
         inputs: &[AudioRenderQuantum],
         outputs: &mut [AudioRenderQuantum],
         params: AudioParamValues,
-        timestamp: f64,
+        current_frame: u64,
+        current_time: f64,
         sample_rate: SampleRate,
     ) -> bool;
```
orottier, line 44: My personal limit for number of method arguments is five. We are at six now. I'm wondering if I should combine `current_frame`, `current_time` and `sample_rate` in a `GlobalScope` struct

src/render/processor.rs
```diff
@@ -7,6 +7,18 @@ use crate::SampleRate;
...
 use super::{graph::Node, AudioRenderQuantum, NodeIndex};
 
+#[non_exhaustive] // we may want to add user-provided blobs to this later
+#[derive(Copy, Clone, Debug)]
+/// The execution context of all AudioProcessors in a given AudioContext
+///
+/// This struct currently only contains information about the progress of time. In a future
+/// version, it should be possible to add arbitrary data. For example, multiple processors might
+/// share a buffer defining a wavetable or an impulse response.
+pub struct GlobalScope {
```
orottier, line 17: I organized the three variables in the GlobalScope now, which is loosely based on the https://www.w3.org/TR/webaudio/#audioworkletglobalscope
These concepts of scopes and worklets don't really translate to rust so this is perhaps the best we can do. With the current setup we leave the future ability open (with no breaking changes) for
> Note: The [AudioWorkletGlobalScope](https://www.w3.org/TR/webaudio/#audioworkletglobalscope) may also contain any other data and code to be shared by these instances. As an example, multiple processors might share an ArrayBuffer defining a wavetable or an impulse response.

orottier:
Ready for review but I want to let this simmer for a few days myself. Not sure where the best balance of spec compliance versus decent Rust paradigms lies

b-ma:
Yup, no real strong opinion on my side. My main point would be to say that:

> With the current setup we leave the future ability open (with no breaking changes) for
> > Note: The [AudioWorkletGlobalScope](https://www.w3.org/TR/webaudio/#audioworkletglobalscope) may also contain any other data and code to be shared by these instances. As an example, multiple processors might share an ArrayBuffer defining a wavetable or an impulse response.

is a very good point!

Just one note: maybe `RenderScope` instead of `Scope` would be a bit more precise

orottier:
> Just one note: maybe `RenderScope` instead of `Scope` would be a bit more precise

Great suggestion, that is a lot more fitting.

I also made a minor but quite important change to hand a reference instead of an owned RenderScope to the processors, see 32213b653fbd0055

PR #146, opened 2022-04-12
Change channels, input/output, length integers all to `usize`

### LINKED ISSUE #122: Use `usize` consistently for channel, input/output and length
from #116 

> As a side note, we have a discrepancy with types here, i.e. number_channel: usize and length: usize, which should both be u32 if we strictly follow the spec. Also this is inconsistent with other places of the code (e.g. ConcreteBaseAudioContext). [...]
> 
> Not sure what we should decide though: unsigned long is usize or u32? As they are generally used as indexes, maybe usize would make more sens somehow but I'm not really sure of all the implications of such choice. My only opinion here is that we should be very consistent and apply the same rule everywhere. Maybe creating and maintaining a map of the correspondances somewhere would be helpful? (also we may have a similar problem with &[f32] and Vec<f32> at some places)

orottier:
Some of them were u32 still.

Fixes #122

### DIFF
--- a/examples/mic_playback.rs
+++ b/examples/mic_playback.rs
@@ -47,10 +47,10 @@ impl AudioNode for MediaRecorder {
     fn channel_config(&self) -> &ChannelConfig {
         &self.channel_config
     }
-    fn number_of_inputs(&self) -> u32 {
+    fn number_of_inputs(&self) -> usize {
         1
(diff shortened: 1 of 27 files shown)


b-ma:
cool, this removes a lot of casting!

PR #148, opened 2022-04-13
Add `base_latency` and `output_latency`

### LINKED ISSUE #138: Implement AudioContext latency attributes
https://www.w3.org/TR/webaudio/#AudioContext-attributes
- base latency
- output latency

It's not directly clear what these numbers should be and what `cpal` can provide us.

Some more insights at https://github.com/WebAudio/web-audio-api/issues/2461

b-ma:
Fixes #138

In the same vein, there is `getOutputTimestamp` (https://webaudio.github.io/web-audio-api/#dom-audiocontext-getoutputtimestamp), maybe I can have a shot in this PR as you prefer

### DIFF
--- a/examples/latency_attributes.rs
+++ b/examples/latency_attributes.rs
@@ -14,7 +14,11 @@ fn main() {
     println!("- BaseLatency: {:?}", context.base_latency());
 
     loop {
+        println!("-------------------------------------------------");
+        println!("+ currentTime {:?}", context.current_time());
         println!("+ OutputLatency: {:?}", context.output_latency());
+        println!("+ AudioTimestamp {:?}", context.get_output_timestamp());
+
         std::thread::sleep(std::time::Duration::from_secs(1));
     }
 }
--- a/examples/latency_attributes.rs
+++ b/examples/latency_attributes.rs
@@ -12,11 +12,9 @@ fn main() {
     sine.start();
 
     println!("- BaseLatency: {:?}", context.base_latency());
-    println!("+ OutputLatency: {:?}", context.output_latency());
 
-    std::thread::sleep(std::time::Duration::from_secs(1));
-    println!("+ OutputLatency: {:?}", context.output_latency());
-
-    std::thread::sleep(std::time::Duration::from_secs(1));
-    println!("+ OutputLatency: {:?}", context.output_latency());
+    loop {
+        println!("+ OutputLatency: {:?}", context.output_latency());
+        std::thread::sleep(std::time::Duration::from_secs(1));
+    }
 }
--- a/examples/latency_attributes.rs
+++ b/examples/latency_attributes.rs
@@ -17,7 +17,6 @@ fn main() {
         println!("-------------------------------------------------");
         println!("+ currentTime {:?}", context.current_time());
         println!("+ OutputLatency: {:?}", context.output_latency());
-        println!("+ AudioTimestamp {:?}", context.get_output_timestamp());
 
         std::thread::sleep(std::time::Duration::from_secs(1));
     }
--- a/examples/latency_attributes.rs
+++ b/examples/latency_attributes.rs
@@ -1,0 +1,22 @@
+use web_audio_api::context::{AudioContext, BaseAudioContext};
+use web_audio_api::node::{AudioNode, AudioScheduledSourceNode};
+
+fn main() {
+    println!("AudioContextLatencyCategory::Interactive");
+    let context = AudioContext::new(Default::default());
+
+    let sine = context.create_oscillator();
+    sine.frequency().set_value(200.);
+    sine.connect(&context.destination());
+
+    sine.start();
+
+    println!("- BaseLatency: {:?}", context.base_latency());
+    println!("+ OutputLatency: {:?}", context.output_latency());
+
+    std::thread::sleep(std::time::Duration::from_secs(1));
+    println!("+ OutputLatency: {:?}", context.output_latency());
+
+    std::thread::sleep(std::time::Duration::from_secs(1));
+    println!("+ OutputLatency: {:?}", context.output_latency());
+}
--- a/src/context/concrete_base.rs
+++ b/src/context/concrete_base.rs
@@ -225,17 +225,6 @@ impl ConcreteBaseAudioContext {
         self.inner.sample_rate
     }
 
-    /// This represents the number of seconds of processing latency incurred by
-    /// the `AudioContext` passing the audio from the `AudioDestinationNode`
-    /// to the audio subsystem.
-    // We don't do any buffering between rendering the audio and sending
-    // it to the audio subsystem, so this value is zero. (see Gecko)
-    #[allow(clippy::unused_self)]
-    #[must_use]
-    pub(super) const fn base_latency(&self) -> f64 {
-        0.
-    }
-
     /// The estimation in seconds of audio output latency, i.e., the interval
     /// between the time the UA requests the host system to play a buffer and
     /// the time at which the first sample in the buffer is actually processed
--- a/src/context/concrete_base.rs
+++ b/src/context/concrete_base.rs
@@ -10,7 +10,7 @@ use crate::param::{AudioParam, AudioParamEvent};
 use crate::render::AudioProcessor;
 use crate::spatial::AudioListenerParams;
 
-use crate::{AtomicF64, AudioListener, SampleRate};
+use crate::{AudioListener, SampleRate};
 
 use crossbeam_channel::Sender;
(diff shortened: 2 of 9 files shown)


src/context/concrete_base.rs
```diff
@@ -210,25 +214,45 @@ impl ConcreteBaseAudioContext {
...
+    /// This represents the number of seconds of processing latency incurred by
+    /// the `AudioContext` passing the audio from the `AudioDestinationNode`
+    /// to the audio subsystem.
+    // We don't do any buffering between rendering the audio and sending
+    // it to the audio subsystem, so this value is zero. (see Gecko)
+    #[allow(clippy::unused_self)]
+    #[must_use]
+    pub(super) const fn base_latency(&self) -> f64 {
+        0.
+    }
```
b-ma, line 237: I followed what is done in Gecko / Firefox, but I'm not super sure it is correct so I though it was maybe better to have this API. Let me know what you think

examples/latency_attributes.rs
```diff
@@ -0,0 +1,22 @@
...
+
+    let sine = context.create_oscillator();
+    sine.frequency().set_value(200.);
+    sine.connect(&context.destination());
+
+    sine.start();
+
+    println!("- BaseLatency: {:?}", context.base_latency());
+    println!("+ OutputLatency: {:?}", context.output_latency());
+
```
orottier, line 16: Might be nice to just infinitely `loop` this block? Then you can properly it changing slightly over time

src/context/concrete_base.rs
orottier: Yeah, I think this is about internal buffering, which we do not do. So probably 0 is fine

src/context/concrete_base.rs
```diff
@@ -210,25 +214,45 @@ impl ConcreteBaseAudioContext {
...
+
+    /// The estimation in seconds of audio output latency, i.e., the interval
+    /// between the time the UA requests the host system to play a buffer and
+    /// the time at which the first sample in the buffer is actually processed
+    /// by the audio output device.
+    #[must_use]
+    pub(super) fn output_latency(&self) -> f64 {
+        self.inner.output_latency.load()
```
orottier, line 245: From the spec
> The [outputLatency](https://www.w3.org/TR/webaudio/#dom-audiocontext-outputlatency) attribute value does not change for the context’s lifetime as long as the connected audio output device remains the same.

But I do see it changing, are we just being more accurate than the spec mandates?

src/render/thread.rs
```diff
@@ -124,7 +127,16 @@ impl RenderThread {
     // This code is not dead: false positive from clippy
     // due to the use of #[cfg(not(test))]
     #[allow(dead_code)]
-    pub fn render<S: Sample>(&mut self, mut buffer: &mut [S]) {
+    pub fn render<S: Sample>(&mut self, mut buffer: &mut [S], infos: &OutputCallbackInfo) {
+        // update output latency, this value might change while running (e.g. sound card heat)
+        let timestamp = infos.timestamp();
```
orottier, line 132: Looks good to me

orottier (commented):
Looking good. Let's have a stab at the timestamp too, go ahead!

src/context/concrete_base.rs
b-ma: I realize I generally use the editor's draft in which the spec seems to have evolved on this specific point 

https://webaudio.github.io/web-audio-api/#dom-audiocontext-outputlatency

> The [outputLatency](https://webaudio.github.io/web-audio-api/#dom-audiocontext-outputlatency) attribute value may change while the context is running or the associated audio output device changes. It is useful to query this value frequently when accurate synchronization is required.

So, we are ok here !

examples/latency_attributes.rs
b-ma: right, that's done!

b-ma:
note that I moved the methods from `BaseAudioContext` to `AudioContext`

src/context/concrete_base.rs
b-ma: Yup this is my impression too in our case, (it is just that there is something I don't really understand about the calculation they do in the spec with double buffering)

src/context/concrete_base.rs
orottier: Huh, interesting. Good to know

src/lib.rs
```diff
@@ -46,6 +49,12 @@ pub const RENDER_QUANTUM_SIZE: usize = 128;
 /// Maximum number of channels for audio processing
 pub const MAX_CHANNELS: usize = 32;
 
+/// Global `Instant` to create `AudioTimestamp` returned by `AudioContext::getOutputTimestamp`
+/// on a timeline shared by all contexts. This can be usefull to tightly synchronise
+/// audio rendering with other processes (e.g. graphics)
+// @note: naming inspired by https://www.w3.org/TR/hr-time-3/#dfn-time-origin
+pub(crate) static TIME_ORIGIN: Lazy<Instant> = Lazy::new(Instant::now);
```
b-ma, line 56: Not completely sure this is ok to do that?

src/context/online.rs
```diff
@@ -55,6 +55,17 @@ pub struct AudioContextOptions {
...
 
+/// Struct returned by `get_output_timestamp` that associate estimated playback
+/// context time (i.e. `current_time - output_latency`) to a monotonic global clock
+/// (i.e. <https://doc.rust-lang.org/std/time/struct.Instant.html>)
+#[derive(Clone, Debug, Default)]
+pub struct AudioTimestamp {
+    /// Represents a point in the time coordinate system of BaseAudioContext’s currentTime.
+    pub context_time: f64,
+    /// Represents a point in the time coordinate system of a Performance interface implementation
+    pub performance_time: f64,
```
b-ma, line 66: I strictly followed the spec here, exposing the time in millisecond, but I'm actually not sure this is the best idea.

As I understand it these values are interesting to tightly synchronize other processes (e.g. graphic) to the audio, in such case maybe it would be more handy and useful to directly expose the `std::time::Duration` since `TIME_ORIGIN`, any idea?

src/lib.rs
b-ma: maybe we could make it public too, I don't have any idea how this could be used in user space (cf. next comment)

src/context/online.rs
b-ma: Maybe we could add some non-spec stuff such as:
- `AudioContext::set_time_origin(i: std::time::Instant)` to allow users to define their own time origin, 
- `AudioTimestamp::performance_duration: std::time::Duration` raw field computed from the default or given time origin

src/context/concrete_base.rs
```diff
@@ -210,25 +214,45 @@ impl ConcreteBaseAudioContext {
...
     }
 
+    /// This represents the number of seconds of processing latency incurred by
+    /// the `AudioContext` passing the audio from the `AudioDestinationNode`
+    /// to the audio subsystem.
+    // We don't do any buffering between rendering the audio and sending
+    // it to the audio subsystem, so this value is zero. (see Gecko)
+    #[allow(clippy::unused_self)]
+    #[must_use]
+    pub(super) const fn base_latency(&self) -> f64 {
```
orottier, line 235: There's no need to put this method here, we can simply have it at the OnlineAudioContext

src/context/online.rs
```diff
@@ -136,11 +153,13 @@ impl AudioContext {
 
         let (sender, _receiver) = crossbeam_channel::unbounded();
         let frames_played = Arc::new(AtomicU64::new(0));
+        let output_latency = Arc::new(AtomicF64::new(0.));
 
         let base = ConcreteBaseAudioContext::new(
             sample_rate,
             number_of_channels,
             frames_played,
+            output_latency,
```
orottier, line 162: This attribute should actually be inside the AudioContext, not the ConcreteBaseAudioContext
This means we can leave it out of the offline.rs file too!

orottier (changes requested):
The outputTimestamp is more complicated than I expected, and to be fair I'm not sure if I am able to approve this without really knowing what it is used for in practise.

All in all, a `Lazy<Instant>` just seems very strange to me.

Could you maybe address my two other comments, and then we merge the baseLatency and outputLatency part?

Perhaps we can then find a real world example that utilizes the outputTimestamp, port it to Rust and see where that takes us?

b-ma:
> Perhaps we can then find a real world example that utilizes the outputTimestamp, port it to Rust and see where that takes us?

Yup you're right, seems more reasonable to do like that. Though I have no idea of where to actually find such example.

I handled your comments, everything lives only in the AudioContext now!

Issue #114, opened 2022-02-18
Consistent clippy lints


orottier:
We ran into a nasty bug due to unsafe casts in #113 
There's a group of clippy lints called `unnecessary_cast, cast_lossless/possible_truncation/possible_wrap/precision_loss/sign_loss` which ideally we enable globally.
These are part of the `pedantic` category of clippy. We could enable it but it will give annoying suggestions regarding
- functions not marked as const (this is something I do not want to commit to, also not beneficial for our project)
- naming not following convention (but we need to stick to the spec)

Goal of this ticket
- use a single clippy configuration for the whole project (not per mod as we do now)
- guard for unsafe casts

PR #150, opened 2022-04-14
Chore: add more clippy lints and apply suggestions

orottier:
Working towards #114

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -1,4 +1,8 @@
 //! General purpose audio signal data structures
+#![warn(clippy::all, clippy::pedantic, clippy::nursery, clippy::perf)]
+#![allow(clippy::module_name_repetitions)]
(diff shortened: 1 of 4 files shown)


b-ma:
Hum sorry, I guess there is a misunderstanding on my side on this point, as I constantly remove the per file clippy directives.

As I understood #114, I though we wanted to have a common config at the crate level. Could you elaborate on the methodology you want to follow and on the final set of directive we'd like to have, so I don't constantly break what you try to achieve in the long term?

orottier:
Hey, no worries. I actually don't really have a plan currently so I never complained about it either.
I'll try to come up with a strategy in the linked ticket so we can discuss it there

b-ma:
In the mean time, maybe having some comment (e.g. `// clippy::do-not-remove-clippy!`) on the ones that must be there

PR #156, opened 2022-04-22
Chore: add sub-project for flamegraphs

b-ma:
Not sure that's the best approach, but I thought it maybe make it more simple to test some changes in different branches compared to having a dedicated branch (as proposed in #117), and we don't spoil the example folder with lots of test files.

Let me know what you think!

For the record with the `offline-granular` my flamegraph still looks quite different (maybe because some features are disabled: `dtrace: system integrity protection is on`, I don't know...)

flamegraph


orottier:
Thanks, but I don't really understand the benefit of a subproject.
Ideally, we can collect the cases we are investigating in a `benches/` folder and monitor them over time with `cargo bench`. That way they do not pollute the examples folder. Also, we can just remove the file if we have fixed the underlying performance issue.

Regarding your flamechart. It does looks similar to mine I think, but
- it misses out on a number of layers, suggesting some functions are inlined or your linker is applying magic
- it does not really have high sampling precision

The last case may be fixed by running as root. I don't have any suggestions for the first item

b-ma:
ok thanks, I will try this way then

Issue #129, opened 2022-03-26
Graph::order_nodes bad performance for large number of nodes


orottier:
The example/benchmark.rs "Granular synthesis" program has terrible performance. This program adds 1500 buffer source nodes with increasing start time and a gain envelope.
 
I ran some profiling, which shows time slices
- 58% Graph::order_nodes
- 29% AudioNode processing (17% audio params, 7% gain, 4% buffer source)
- 12% Graph processing overhead

Probably some terrible quadratic complexity is happening in the order_nodes function. (or maybe, we are ordering the nodes over and over again?)
Next up: audio param profiling

(open image in new tab to get mouse hover info)
![flamegraph](https://user-images.githubusercontent.com/5442615/160243651-f1876568-b606-4849-a544-054bf82f7d62.svg)

b-ma:
Hey, I did some (very) dirty logging and here are some impressions I have

> (or maybe, we are ordering the nodes over and over again?)

I might be mistaking, but it seems to me that the ordering is done according the removal of the nodes from the graph after they finished their processing (which is de facto quite often in the granular bench), maybe a bit more than expected (I counted 1827 where I was waiting for 1500) but that could be due to the fact that related AudioBufferSource and GainNode are not always dropped at the exact same time. In any case it does not appear as if it was done on each tick or something.

But, I discovered something that could be a real problem if I'm not mistaking. I just dirty logged the number of nodes in `Graph.order_nodes()`, something like:

```
println!("{:?}", duration);
println!("{:?}", self.nodes.keys().len());
println!("{:?}", self.ordered.len());
```

which this gives me at the very beginning:

```
18.803119ms
7506
7506
```

while the number of nodes in the graph should be `> num nodes 3002` (~ 1500 * (src + gain))

After some struggling, it just appeared to me that the `AudioParam`s seems to be stored in `Graph::nodes` along with the `AudioNode`s, in which case these numbers seems more logical because `GainNode` has 1 param (`gain`) and `AudioBufferSourceNode` has 2 params (`detune` and `playbackRate`). So maybe we are just increasing the complexity the graph ordering by a factor of 2.5 for nothing here. Do you confirm this impression or do you think it's not that relevant regarding the implementation of the algorithm, I can't really figure out?

orottier:
Thanks for digging in!

Indeed, AudioParams are treated just like AudioNodes, and are therefore part of the audio graph.
[...]

PR #157, opened 2022-04-23
Proof of concept: do not sort the audio graph when nodes are dropped

orottier:
Relates to #129

This improves benchmark performance x10:
- 13 | Granular synthesis | 3299 | 2.3x | 7.5
Old speedup was 0.6

Work in progress because we can update the ordering while we are
iterating it, instead of doing it after the main loop.

### DIFF
--- a/src/render/graph.rs
+++ b/src/render/graph.rs
@@ -341,9 +341,16 @@ impl Graph {
             }
         });
 
-        // If there were any nodes decomissioned, clear current graph order
+        // If there were any nodes decomissioned, remove from graph order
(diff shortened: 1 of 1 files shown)


b-ma:
Nice ! 

Just to note improved speedup is more x4 than x10 (0.6 -> 2.3), 7.5 is the duration of the computed buffer

Almost the same on my machine: ~0.5 -> 1,6

```
- 13      | Granular synthesis                                                       | 4689          | 1.6x                  | 7.5
```

Quite impressive with such small change :)

Issue #147, opened 2022-04-12
Implement the concept "actively processing"


orottier:
https://www.w3.org/TR/webaudio/#AudioNode-actively-processing

PR #158, opened 2022-04-25
Fix : accumulate and mix inputs according to the node channel config

b-ma:
A first draft in the direction of #147, let me know what you think. 

For the record here are my flamegraphs for the granular bench example:

before
164681860-ec4dcab3-d746-4825-a75b-a15e40a8ee5d (1)

after
_flame_offline_granular

The code I use for the flamegraphs (I did not manage to make something simply work with a `benches` directory)

```rs
use std::fs::File;
use std::time::Instant;

use rand::Rng;
 
use web_audio_api::context::{BaseAudioContext, OfflineAudioContext};
use web_audio_api::SampleRate;

fn main() {
    let duration = 120. / 16.;
    let sample_rate = SampleRate(48000);

    let length = (duration * sample_rate.0 as f64) as usize;
    let mut context = OfflineAudioContext::new(1, length, sample_rate);
    // let mut context = OfflineAudioContext::new(1, 256, sample_rate);

    let file = File::open("samples/think-mono-48000.wav").unwrap();
    let buffer = context.decode_audio_data_sync(file).unwrap();

    let mut offset = 0.;
    let mut rng = rand::thread_rng();

    // this 1500 sources...
    while offset < duration {
        let env = context.create_gain();
        env.connect(&context.destination());

        let src = context.create_buffer_source();
        src.connect(&env);
        src.set_buffer(buffer.clone());

        let rand_start = rng.gen_range(0..1000) as f64 / 1000. * 0.5;
        let rand_duration = rng.gen_range(0..1000) as f64 / 1000. * 0.999;
        let start = offset * rand_start;
        let end = start + 0.005 * rand_duration;
        let start_release = (offset + end - start).max(0.);

        env.gain().set_value_at_time(0., offset);
        env.gain().linear_ramp_to_value_at_time(0.5, offset + 0.005);
        env.gain().set_value_at_time(0.5, start_release);
        env.gain()
            .linear_ramp_to_value_at_time(0., start_release + 0.05);

        src.start_at_with_offset_and_duration(offset, start, end);

        offset += 0.005;
    }

    let start = Instant::now();
    let buffer = context.start_rendering_sync();
    let duration = start.elapsed();

    println!("> processing duration (ms): {:?}", duration.as_millis());
    println!("> buffer.duration (s): {:?}", buffer.duration());
    println!(
        "> speedup vs. realtime (ms): {:?}",
        (buffer.duration() * 1000.) / duration.as_millis() as f64
    );
}
```

No visible improvement in the bench numbers, but I guess the `visit` takes so much place this is difficult to see any improvement there.

### DIFF
--- a/src/node/gain.rs
+++ b/src/node/gain.rs
@@ -94,16 +94,18 @@ impl AudioProcessor for GainRenderer {
         let input = &inputs[0];
         let output = &mut outputs[0];
 
-        let gain_values = params.get(&self.gain);
-
         *output = input.clone();
 
-        output.modify_channels(|channel| {
-            channel
-                .iter_mut()
-                .zip(gain_values.iter())
-                .for_each(|(value, g)| *value *= g)
-        });
+        if !input.is_silent() {
+            let gain_values = params.get(&self.gain);
+
+            output.modify_channels(|channel| {
+                channel
+                    .iter_mut()
+                    .zip(gain_values.iter())
+                    .for_each(|(value, g)| *value *= g)
+            });
+        }
 
         false
     }
--- a/src/param.rs
+++ b/src/param.rs
@@ -568,20 +568,24 @@ impl AudioProcessor for AudioParamProcessor {
         let period = 1. / scope.sample_rate.0 as f64;
         let param_intrisic_values = self.tick(scope.current_time, period, RENDER_QUANTUM_SIZE);
 
-        let input = &inputs[0]; // single input mode
         let param_computed_values = &mut outputs[0];
 
         param_computed_values
             .channel_data_mut(0)
             .copy_from_slice(param_intrisic_values);
 
-        param_computed_values.add(input, ChannelInterpretation::Discrete);
+        let input = &inputs[0]; // single input mode
+        if !input.is_silent() {
+            param_computed_values.add(input, ChannelInterpretation::Discrete);
+        }
 
         true // has intrinsic value
     }
 }
 
 impl AudioParamProcessor {
+    // [spec] intrinsic parameter value [is the] the value the AudioParam would
+    // normally have without any audio connections
     pub fn intrisic_value(&self) -> f32 {
         if self.intrisic_value.is_nan() {
             self.default_value
--- a/src/param.rs
+++ b/src/param.rs
@@ -574,7 +574,11 @@ impl AudioProcessor for AudioParamProcessor {
             .channel_data_mut(0)
             .copy_from_slice(param_intrisic_values);
 
+        // add input signal to param computed values
+        // @note - maybe we should clamp after that too? not very clear in spec
+        // but it would appear logical
         let input = &inputs[0]; // single input mode
+
         if !input.is_silent() {
             param_computed_values.add(input, ChannelInterpretation::Discrete);
         }
--- a/src/param.rs
+++ b/src/param.rs
@@ -574,12 +574,12 @@ impl AudioProcessor for AudioParamProcessor {
             .channel_data_mut(0)
             .copy_from_slice(param_intrisic_values);
 
-        // add input signal to param computed values
-        // @note - maybe we should clamp after that too? not very clear in spec
-        // but it would appear logical
-        let input = &inputs[0]; // single input mode
+        // add input signal to param computed values if not silent
+        let input = &inputs[0];
 
         if !input.is_silent() {
+            // @note/todo - maybe we should clamp after that too? this is not
+            // very clear in the spec but it would appear quite logical
             param_computed_values.add(input, 1, ChannelInterpretation::Discrete);
         }
 
@@ -588,7 +588,7 @@ impl AudioProcessor for AudioParamProcessor {
 }
 
 impl AudioParamProcessor {
-    // [spec] intrinsic parameter value [is the] the value the AudioParam would
+    // [spec] intrinsic parameter value [is] the value the AudioParam would
     // normally have without any audio connections
     pub fn intrisic_value(&self) -> f32 {
         if self.intrisic_value.is_nan() {
--- a/src/param.rs
+++ b/src/param.rs
@@ -580,7 +580,7 @@ impl AudioProcessor for AudioParamProcessor {
         let input = &inputs[0]; // single input mode
 
         if !input.is_silent() {
-            param_computed_values.add(input, ChannelInterpretation::Discrete);
+            param_computed_values.add(input, 1, ChannelInterpretation::Discrete);
         }
 
         true // has intrinsic value
--- a/src/render/graph.rs
+++ b/src/render/graph.rs
@@ -353,6 +353,8 @@ impl Graph {
 
 #[cfg(test)]
 mod tests {
+    use float_eq::assert_float_eq;
+
     use super::*;
 
     #[derive(Debug, Clone)]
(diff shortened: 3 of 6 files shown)


b-ma:
Hey, digging into this, I found an issue with how connections are mixed in inputs (connection are mixed according to the producer config, not the receiver one which is not right according to [1] and produce unpredictable results). 
I just added a failing test case in `graph.rs` to spot that

It shouldn't be very difficult to fix though, I will do that soon

[1] https://webaudio.github.io/web-audio-api/#channel-up-mixing-and-down-mixing

b-ma:
This is fixed, let me know what you think !

Storing incoming connection (as you did at some point) could simplify the code as we could do all the add and mixing just before the `node.process` in `graph.render`. I can go on with that too if you want

b-ma:
maybe this is best to merge that fix, and continue in actively processing nodes in a dedicated PR (I also realize it would be maybe best to finish #34 before)

orottier:
Good catch in the output mixing! I forgot to refactor that apparently in `def9304f8151353b`
I have not really had the time to look at this PR, there's a lot going on. I will try to put some comments on it later this week.

src/node/gain.rs
```diff
@@ -94,16 +94,18 @@ impl AudioProcessor for GainRenderer {
...
+        if !input.is_silent() {
+            let gain_values = params.get(&self.gain);
+
+            output.modify_channels(|channel| {
+                channel
+                    .iter_mut()
+                    .zip(gain_values.iter())
+                    .for_each(|(value, g)| *value *= g)
+            });
+        }
```
orottier, line 108: You still should fill the output with silence here, otherwise the outputs are unspecified (reused from previous render quantum)

src/param.rs
```diff
@@ -568,20 +568,28 @@ impl AudioProcessor for AudioParamProcessor {
...
 
         param_computed_values
             .channel_data_mut(0)
             .copy_from_slice(param_intrisic_values);
 
-        param_computed_values.add(input, ChannelInterpretation::Discrete);
+        // add input signal to param computed values if not silent
+        let input = &inputs[0];
+
+        if !input.is_silent() {
```
orottier, line 580: This optimization is already present in the `add` implementation of `AudioRenderQuantum` so we can keep this code clean

src/render/render_quantum.rs
```diff
@@ -453,6 +453,12 @@ impl AudioRenderQuantum {
         self.channels.truncate(1);
     }
 
+    /// Check if this `RenderQuantum` contains a single channel of silence
+    #[must_use]
+    pub fn is_silent(&self) -> bool {
```
orottier, line 458: Why require only one channel?

src/render/render_quantum.rs
```diff
@@ -464,15 +470,10 @@ impl AudioRenderQuantum {
...
     }
 
-    /// Sum two `AudioRenderQuantum`s
-    ///
-    /// If the channel counts differ, the buffer with lower count will be upmixed.
-    pub fn add(&mut self, other: &Self, interpretation: ChannelInterpretation) {
-        // mix buffers to the max channel count
-        let channels_self = self.number_of_channels();
-        let channels_other = other.number_of_channels();
-        let channels = channels_self.max(channels_other);
```
orottier, line 474: Taking `max` was a strange choice, great that we are having this fixed

src/render/render_quantum.rs
```diff
@@ -464,15 +470,10 @@ impl AudioRenderQuantum {
...
-
+    /// Sum two `AudioRenderQuantum`s, both quantums will be up/down mix according
+    /// to given number of channels and channel interpretation.
+    /// see <https://webaudio.github.io/web-audio-api/#channel-up-mixing-and-down-mixing>
+    pub fn add(&mut self, other: &Self, channels: usize, interpretation: ChannelInterpretation) {
```
orottier, line 476: I think I would prefer not to take a `channel` argument here, but just assume `self` already has the right number of channels. To me that makes it more clean and also more semantical: if you want to add another buffer to an existing one, the number of channels will not change in the result

Looking at the code in `graph.rs`, the inputs are not already set to the right number of channels, so that should then be changed
```
    pub fn add_node(
        &mut self,
        index: NodeIndex,
        processor: Box<dyn AudioProcessor>,
        inputs: usize,
        outputs: usize,
        channel_config: ChannelConfig,
    ) {
        // todo, allocate on control thread, make single alloc..?
        let inputs = vec![AudioRenderQuantum::new(self.alloc.silence()); inputs];
        let outputs = vec![AudioRenderQuantum::new(self.alloc.silence()); outputs];
```

src/render/graph.rs
```diff
@@ -314,8 +320,27 @@ impl Graph {
                 .for_each(|edge| {
                     let output_node = nodes.get_mut(&edge.other_id).unwrap();
                     output_node.has_inputs_connected = true;
-                    let signal = &node.outputs[edge.self_index];
-                    output_node.inputs[edge.other_index].add(signal, channel_interpretation);
+
+                    let connection = &node.outputs[edge.self_index];
+
+                    if !connection.is_silent() {
```
orottier, line 326: Again this optimization for checking silence is probably not worth is because it is also present in the actual `add` method

orottier (commented):
Sorry it took me a while to have a look at this. Great to see that we are having a few important bugs fixed.

src/node/gain.rs
b-ma: hum I don't understand, has `*output = input.clone();` (line 97) we are just propagating forward the silent input, or did I miss something?

src/render/graph.rs
b-ma: I would argue that's the same problem as for the `AudioParam`, but I have no numbers this time :)

src/render/render_quantum.rs
b-ma: It questioned me too, I'm not completely sure about this one. In https://webaudio.github.io/web-audio-api/#AudioNode-actively-processing, the spec says that: 

> `[AudioNode](https://webaudio.github.io/web-audio-api/#AudioNode)s that are not [actively processing](https://webaudio.github.io/web-audio-api/#actively-processing) output a single channel of silence.`. 

I have the impression (but I'm not sure) that will more simple to reason about in the long run.

src/render/render_quantum.rs
b-ma: Hum I think I see what you mean, but there was something strange I don't recall precisely, my impression was that `add` should be:
- either `add(&self, other)`, with all mixing done by the caller and `add` being completely dumb
- either `add(&self, other, num_channels, interpretation)`, with arbitrary input and `add` should take care of all mixing everything (which appeared more safe in the end than the dumb version)

But I don't really remember why I ended up thinking that, maybe that's a misunderstanding on my side (thinking too much of `add` as a kind of static method) I will review that and tell you if something comes back in the way

src/node/gain.rs
orottier: You are right, sorry I missed that part

src/param.rs
orottier: This is a very tricky call to make. You might improve the current benchmark a bit, but we don't know the effect on all other programs. By checking for silence twice, you add extra work for all programs that handle non-silent inputs

src/render/render_quantum.rs
orottier: I currently do not see how this would relate to the spec text.
In any case, the current name is misleading and should probably be called `is_single_channel_of_silence`.
Also, I do not see any drawbacks in just implementing
```rust
for c in &self.channels {
    if !c.is_silent() { return false }
}
return true
```

src/render/render_quantum.rs
orottier: Cool, let me know. My opinion is that `add` should be allowed to perform some mixing to align the buffers

src/param.rs
b-ma: I would say that `AudioParam` are more often *not* modulated by an audio signal than the other way. But ok, let's keep that for another time with a more precise bench considering the two cases side by side, so we will have a clearer picture of the trade-off

src/render/render_quantum.rs
b-ma: hum, my point in implementing `is_silence()` was really about the `actively processing` stuff. I really think we should consider that an `AudioRenderQuantum` containing several channel of silence is wrong (because of the spec, and because that would mean we do something, e.g. mixing, that is unnecessary), so I think we need a way to track if such problem appears. 
On the other end, I really agree with you that the implementation you propose is more safe. So what about something like (at least in a temporary way):

```rs
if cfg(test) {
   if one_channel_of_silence { 
      return true;
   } else if several_channels_of_silence {
      panic!("this should not happen...");
   }
} else {
    for c in &self.channels {
       if !c.is_silent() { return false; }
    }
    return true;
}
```

This way end-user applications are safe but we can write tests to ensure that the propagation of silence is properly done?

orottier: I see your point, but since the function will be part of the public interface, I think we should stick to the principle of least surprise.
Switching behaviour between test and not-test could be very surprising to end users.
Also, currently, we allow to upmix a single channel of silence to multiple channels. We don't panic at that call site either.

src/render/render_quantum.rs
b-ma: Hey, no I didn't have any time to continue last week, and this week will be probably be a bit busy too so if you have some time, yes that's a good idea to go on with the fix (feel free to push on this branch if you want too)!

PR #159, opened 2022-05-28
Hotfix mixing inputs

orottier:
Hey, this work is very much based on your work in #158 but I have tried to keep the code a bit smaller and interface changes minimal.
The relevant change is that `add(` now takes the full ChannelConfig information.
I'm still thinking how to make a good regression test for this. I dislike the amount of boilerplate is needed inside the graph mod for the test that you have added (which succeeds with my changes though :) ), so I will probably make a big integration test file `tests/mixing.rs` which covers most cases.


b-ma:
Hey, I must admit I lost my ideas a bit on all that... but if the falling test passes I guess it's nice :)

> add() now takes the full ChannelConfig information.

Looks good to me!

> so I will probably make a big integration test file tests/mixing.rs which covers most cases.

Yup good idea, I agree about the amount of boilerplate, maybe not the best place in the mod

So what do we do with #158 ? Drop it and restart the silence stuff on a fresh basis ?

orottier:
Sorry it took me a bit more time. When I was writing the integration tests I discovered a bug (of a missing feature ;)) when you would set the destination channel intepretation to Discrete.
This should fix all the issues.

> So what do we do with https://github.com/orottier/web-audio-api-rs/pull/158 ? Drop it and restart the silence stuff on a fresh basis ?
Probably best. Sorry about causing havoc to your work

b-ma:
> when you would set the destination channel intepretation to Discrete.

Nice that's important indeed

>> So what do we do with https://github.com/orottier/web-audio-api-rs/pull/158 ? Drop it and restart the silence stuff on a fresh basis ?

> Probably best. Sorry about causing havoc to your work

No problem, let's consider that was a draft, a least things are going into the right direction :)

PR #160, opened 2022-06-08
Drop SampleRate type, switch to plain f32

orottier:
Work in progress, the build succeeds but I need to fix some tests still.
Just wanted to let you know I have this in the pipeline. Please base new PRs on this one to avoid merge conflicts.

For the benchmarks in the WAC paper, we could choose to run on this branch already, altough it probably won't make any difference compared to `main`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -64,10 +64,6 @@ mod io;
 mod analysis;
 mod message;
 
-/// Number of samples processed per second (Hertz) for a single channel of audio
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
-pub struct SampleRate(pub u32);
-
 /// Atomic float 32, only `load` and `store` are supported, no arithmetics
 #[derive(Debug)]
 pub(crate) struct AtomicF32 {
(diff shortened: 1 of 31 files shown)


src/lib.rs
```diff
@@ -130,11 +126,13 @@ impl AtomicF64 {
 ///
 #[track_caller]
 #[inline(always)]
-pub(crate) fn assert_valid_sample_rate(sample_rate: SampleRate) {
-    if sample_rate.0 == 0 {
+pub(crate) fn assert_valid_sample_rate(sample_rate: f32) {
+    // 1000 Hertz is a just a random cutoff, but it helps a if someone accidentally puts a
+    // timestamp in the sample_rate variable
+    if sample_rate <= 1000. {
```
b-ma, line 132: We could eventually pick values from the spec here : https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-createbuffer-samplerate

b-ma:
Great work! I think this is really much more clean and simple like that 

> Please base new PRs on this one to avoid merge conflicts.

ok no problem (I'm unfortunately still out of the party for a little moment tough...)

Issue #21, opened 2021-09-30
Add ConvolverNode


orottier:
https://webaudio.github.io/web-audio-api/#ConvolverNode

Jerboas86:
I've got a direct form implementation ready for this one. 
But the state of the art is to mix direct form and FFT to obtain both good latency and computation efficiency. 
I can push a PR with the direct form alg or try to write the DF+FFT alg but it will require more time.

orottier:
Hi @Jerboas86, you are still assigned this issue. Are you interested in publishing your direct form implementation?

orottier:
To be continued with #220

PR #161, opened 2022-06-28
Proof of concept of single-channel convolution engine

orottier:
Getting started with #21 , this is a extremely simple, computationally intensive, single channel convolution engine

~Try it out *with speakers very low because the normalization is not implemented yet!!*~
normalization is now implemented!

cargo run --release --example convolution

My CPU barely keeps up with the larger response buffer,  but this sets a baseline for further improvement


src/node/convolver.rs
```diff
@@ -0,0 +1,353 @@
...
+pub struct ConvolverOptions {
+    /// The desired buffer for the ConvolverNode
+    pub buffer: Option<AudioBuffer>,
+    /// The opposite of the desired initial value for the normalize attribute
+    pub disable_normalization: bool,
+    /// AudioNode options
+    pub channel_config: ChannelConfigOptions,
+}
+
+/// Processing node which applies a linear convolution effect given an impulse response.
```
b-ma, line 67: Maybe add some disclaimer, "This node is still a proof of concept and won't be usable in most real-world use-cases"

src/node/convolver.rs
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    /// Denotes if the response buffer will be scaled with an equal-power normalization
+    pub fn normalize(&self) -> bool {
+        self.normalize
+    }
+
+    /// Update the `normalize` setting. This will only have an effect when `set_buffer` is called.
+    pub fn set_normalize(&mut self, value: bool) {
+        // TODO, use AtomicBool to prevent &mut self?
```
b-ma, line 219: Yup, would be consistent with the other nodes

orottier:
Or port the C++ code to Rust. It is about 5k lines of code..

orottier:
Hey @b-ma, it took me some time to continue with this.
I had a look at both porting or binding existing C++ stuff, but it is not my range of expertise.
Hence for now I opted to do some incremental improvement on the current implementation. Using Frequency Domain Delay Lines there is a nice performance improvement while not overcomplicating stuff.
Could you check if this works on your end?
Up next is: adding tests (trivial impulse response, comparing with IIR, etc) and then the next improvement: Nonuniform partition scheduling
For literature I am looking at https://www.eecs.qmul.ac.uk/~josh/documents/2017/Jillings%20IEEE%20WASPAA%202017.pdf and https://github.com/vtolani95/convolution/blob/master/reverb.py

b-ma:
Hey, nice! I will a look tomorrow

Maybe this [https://www.dspguide.com/ch18.htm](https://www.dspguide.com/ch18.htm) could be of some help too, didn't check it yet and maybe a bit naive in terms of implementation but this book is generally really nice for explaining the concepts

b-ma:
Hey, didn't have time to have a look at the code, but all tests are passing on my side and the example do not crash anymore! Quite nice :)

orottier:
Good to hear. If the code looks good to you I intend to merge this version.
I have created #220  for further improvements

examples/convolution.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    println!("Dry");
+    thread::sleep(time::Duration::from_millis(4_000));
+
+    println!("Small room");
+    convolve.set_buffer(Some(impulse_buffer1));
+    thread::sleep(time::Duration::from_millis(4_000));
+
+    println!("Parking garage");
+    convolve.set_buffer(Some(impulse_buffer2));
```
b-ma, line 37: Maybe cross-fading between the different convolvers would be more clean, but maybe you want to case the ability to change the buffer?

src/node/convolver.rs
```diff
@@ -0,0 +1,568 @@
...
+            &mut self.fft_input,
+            &mut self.fft_output,
+            &mut self.fft_scratch,
+        )
+        .unwrap();
+        &self.fft_output[..]
+    }
+
+    fn inverse(&mut self) -> &[f32] {
+        let fft = self.fft_planner.plan_fft_inverse(self.fft_input.len());
```
b-ma, line 286: maybe `fft` and `fft_inverse` could be created once for all in `new`, I saw `fft_planner` has some cache system but it looks a bit like unnecessary work

b-ma:
Except my small comment, seems pretty good to me. I didn't go into the details of the implementation (I'm interested but...) and the tests look pretty nice!

Seems that going to multichannel from here wouldn't be too complicated, no?

Also there was a bench for the convolution in the original Paul's benchmarks, we could add it too to have an idea of where we get (I can port it if you don't want to make you hand dirty with JS :)

src/node/convolver.rs
```diff
@@ -0,0 +1,568 @@
...
+    let mut scale = 1. / power;
+
+    // Calibrate to make perceived volume same as unprocessed.
+    scale *= gain_calibration;
+
+    // Scale depends on sample-rate.
+    scale *= gain_calibration_sample_rate / sample_rate;
+
+    // True-stereo compensation.
+    if number_of_channels == 4 {
```
b-ma, line 44: I've saw that's in the spec but it seems weird...

src/node/convolver.rs
orottier: Yes right, we could store the `Arc<dyn FFT>`s directly instead of the planner. Will do that

examples/convolution.rs
orottier: Yeah I have not put very much effort in this, but it showcases the main features so I am happy.
The spec even says:
> Note: If the [buffer](https://www.w3.org/TR/webaudio/#dom-convolvernode-buffer) is set to an new buffer, audio may glitch. If this is undesirable, it is recommended to create a new [ConvolverNode](https://www.w3.org/TR/webaudio/#convolvernode) to replace the old, possibly cross-fading between the two.

src/node/convolver.rs
orottier: Looking at the diagram it makes sense: https://www.w3.org/TR/webaudio/#Convolution-channel-configurations
The output is stereo, so pairs of channels are summed, so compensate with 0.5.
I am more confused by `    let gain_calibration = 0.00125;` which comes out of nowhere. But hey, let's conform to the spec :)

orottier:
> Seems that going to multichannel from here wouldn't be too complicated, no?

Indeed, just a bit of extra bookkeeping

> Also there was a bench for the convolution in the original Paul's benchmarks, we could add it too to have an idea of where we get (I can port it if you don't want to make you hand dirty with JS :)

Good point, I can give it a try!

examples/convolution.rs
b-ma: Yup no big deal we can make something more clean with cross-fades later

src/node/convolver.rs
b-ma: Yup this algorithm is quite strange, don't really get the idea with the hardcoded sample rate too... but as you say "let's conform to the spec :)"

b-ma:
Absolutely not related to the issue but, I'm happy it's out and quite usable ! https://www.npmjs.com/package/node-web-audio-api :)

orottier:
> Absolutely not related to the issue but, I'm happy it's out there and quite usable: https://www.npmjs.com/package/node-web-audio-api :)
> 
> (I will probably move the repo to my team's organization, so maybe I will need to re-invite you as collaborator)

Really cool. Congrats on the milestone!

src/buffer.rs
```diff
@@ -225,6 +225,17 @@ impl AudioBuffer {
         self.channel_data(channel_number).as_slice()
     }
 
+    /// Return a mutable slice of the underlying data of the channel
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if:
+    /// - the given channel number is greater than or equal to the given number of channels.
+    pub fn get_channel_data_mut(&mut self, channel_number: usize) -> &mut [f32] {
```
orottier, line 234: Take care! I slipped in a new public method here. Similar to #214 I ran into an issue where I could not port the example directly.
The method is not part of the spec, but it is idiomatic rust (compare `get_channel_data`/`get_channel_data_mut` to HashMap `get`/`get_mut`).
What do you think?

orottier:
Ready for final comments

src/node/convolver.rs
```diff
@@ -0,0 +1,573 @@
...
+        self.buffer.lock().unwrap().clone()
+    }
+
+    /// Set or update the impulse response buffer
+    ///
+    /// # Panics
+    ///
+    /// Panics when the sample rate of the provided AudioBuffer differs from the audio context
+    /// sample rate.
+    pub fn set_buffer(&self, mut buffer: Option<AudioBuffer>) {
```
b-ma, line 187: For consistency, maybe we should have the same API here as for the `AudioScheduledSourceNode`

https://github.com/orottier/web-audio-api-rs/blob/d003e9938b056f0d840b8f832bd21510e7fb3ddd/src/node/audio_buffer_source.rs#L250

In both case this is `buffer, of type [AudioBuffer](https://webaudio.github.io/web-audio-api/#AudioBuffer), nullable` in the spec. 
For code readability I personally prefer the `pub fn set_buffer(&self, buffer: AudioBuffer)` version, but for spec compliance the `Option` is probably a better choice.

orottier: Good point, but I am not sure what is best.
The ConvolverNode is different because you can set the buffer multiple times. This is not allowed for AudioBufferSourceNode.
Hence I found it necessary to allow the nullable option here. To disable convolution temporarily and add in a new buffer later.
But yeah, this is inconsistent. And probably no-one in their right mind would reuse a convolvernode for different impulse responses. I will make the API consistent with AudioBufferSourceNode

src/buffer.rs
b-ma: Not sure this would solve #214 though as you can't change the buffer length from there, no?

But it feels quite sensible and handy to me

orottier: Yeah that is true of course.
This makes me think, what if you use the JS code from #214 and create a stereo buffer where both channels have unequal length? insanity

b-ma: Yup... JS you know :)

b-ma:
Seems we are good to go :) congrats!

PR #164, opened 2022-07-01
Hide ConcreteBaseAudioContext

orottier:
This is just cosmetic for the docs, a fix for the discussion in #155

### DIFF
--- a/src/context/base.rs
+++ b/src/context/base.rs
@@ -268,7 +268,7 @@ pub trait BaseAudioContext {
         opts: AudioParamDescriptor,
         dest: &AudioContextRegistration,
     ) -> (crate::param::AudioParam, AudioParamId) {
-        let param = self.base().register(move |registration| {
+        let param = self.register(move |registration| {
             let (node, proc) = crate::param::audio_param_pair(opts, registration);
 
             (node, Box::new(proc))
--- a/src/context/base.rs
+++ b/src/context/base.rs
@@ -9,6 +9,7 @@ use crate::media::MediaDecoder;
 use crate::node::{AudioNode, ChannelConfigOptions};
 use crate::param::AudioParamDescriptor;
 use crate::periodic_wave::{PeriodicWave, PeriodicWaveOptions};
+use crate::render::AudioProcessor;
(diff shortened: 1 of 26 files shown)


src/context/base.rs
```diff
@@ -9,21 +9,32 @@ use crate::media::MediaDecoder;
...
 /// An audio context controls both the creation of the nodes it contains and the execution of the
 /// audio processing, or decoding.
-///
-/// Please note that in rust, we need to differentiate between the [`BaseAudioContext`] trait and
-/// the [`ConcreteBaseAudioContext`] concrete implementation.
 #[allow(clippy::module_name_repetitions)]
 pub trait BaseAudioContext {
-    /// retrieves the `ConcreteBaseAudioContext` associated with this `AudioContext`
+    /// Returns the [`BaseAudioContext`] concrete type associated with this `AudioContext`
     fn base(&self) -> &ConcreteBaseAudioContext;
```
b-ma, line 23: maybe this could be `pub(crate)` then?

src/context/base.rs
```diff
@@ -9,21 +9,32 @@ use crate::media::MediaDecoder;
...
 
+    /// Construct a new pair of [`AudioNode`] and [`AudioProcessor`]
+    ///
+    /// The `AudioNode` lives in the user-facing control thread. The Processor is sent to the render thread.
+    fn register<
```
b-ma, line 28: Should we call this one in all `AudioNode`s contructor then? 
i.e. `context.register()` instead of `context.base().register`

b-ma: (maybe I'm wrong thinking too much as OOP...)

b-ma:
Nice too! Definitely going in the right direction. 

Maybe I'm mistaking but it seems to me that we can still access the `ConcreteBaseAudioContext` through `context.base().base()`?

b-ma:
Ho I just saw your message in the issue, which for whatever reason felt into the "Promotions" folder. I think I understand the problem better, forget my comments

src/context/base.rs
orottier: yeah good point, for consistency

PR #167, opened 2022-07-04
Improved AudioScheduledSourceNode

b-ma:
Hey,

A first progress in improving `AudioScheduledSourceNode` performances

Mostly reviewed how output is created:
- fill `output` channel by channel to access memory in an aligned way
- uses more functional paradigms

This improves perf between 1.3 and 1.5 in benchmarks on my mac, which I think is really outside the noise

### before

| id      | name                                                                     | duration (ms) | Speedup vs. realtime  | buffer.duration (s) |
| - | - | - | - | - |
| 2       | Simple source test without resampling (Mono)                             | 147           | 816.3x                | 120 |
| 3       | Simple source test without resampling (Stereo)                           | 193           | 621.8x                | 120 |
| 4       | Simple source test with resampling (Mono)                                | 145           | 827.6x                | 120 |
| 5       | Simple source test with resampling (Stereo)                              | 195           | 615.4x                | 120 |

### after

| id      | name                                                                     | duration (ms) | Speedup vs. realtime  | buffer.duration (s)
| - | - | - | - | - |
| 2       | Simple source test without resampling (Mono)                             | 98            | 1224.5x               | 120
| 3       | Simple source test without resampling (Stereo)                           | 141           | 851.1x                | 120
| 4       | Simple source test with resampling (Mono)                                | 98            | 1224.5x               | 120
| 5       | Simple source test with resampling (Stereo)                              | 140           | 857.1x                | 120

I would be interested if you see the same kind of changes on your side ?

In any case, do not merge yet, I would also like to try something else.

I think we could also apply that strategy to the delay node (which also would allow to remove `set_channels_values_at` fomr `AudioRenderQuantum`)

### edit after 023e767

| id      | name                                                                     | duration (ms) | Speedup vs. realtime  | buffer.duration (s)
| - | - | - | - | - |
| 2       | Simple source test without resampling (Mono)                             | 90            | 1333.3x               | 120
| 3       | Simple source test without resampling (Stereo)                           | 108           | 1111.1x               | 120
| 4       | Simple source test with resampling (Mono)                                | 91            | 1318.7x               | 120
| 5       | Simple source test with resampling (Stereo)                              | 110           | 1090.9x               | 120

### DIFF
--- a/src/node/audio_buffer_source.rs
+++ b/src/node/audio_buffer_source.rs
@@ -180,7 +180,7 @@ impl AudioBufferSourceNode {
                 detune: d_proc,
                 playback_rate: pr_proc,
                 render_state: AudioBufferRendererState::default(),
-                positions: [-1.; RENDER_QUANTUM_SIZE],
+                playback_infos: [None; RENDER_QUANTUM_SIZE],
             };
 
             let node = Self {
@@ -324,10 +324,9 @@ struct AudioBufferSourceRenderer {
     detune: AudioParamId,
     playback_rate: AudioParamId,
     render_state: AudioBufferRendererState,
-    /// Internal buffer used to compute the position to be played in the
-    /// source buffer per sample.
-    /// By convention a position of -1. means output should be zero.
-    positions: [f64; RENDER_QUANTUM_SIZE],
+    /// Internal buffer used to store playback infos to compute the samples
+    /// according to the source buffer. (prev_sample_index, k)
+    playback_infos: [Option<(usize, f32)>; RENDER_QUANTUM_SIZE],
 }
 
 impl AudioProcessor for AudioBufferSourceRenderer {
@@ -448,7 +447,7 @@ impl AudioProcessor for AudioBufferSourceRenderer {
                 || current_time >= stop_time
                 || self.render_state.buffer_time_elapsed >= duration
             {
-                self.positions[index] = -1.;
+                self.playback_infos[index] = None;
                 current_time += dt;
 
                 continue; // nothing more to do for this sample
@@ -503,9 +502,14 @@ impl AudioProcessor for AudioBufferSourceRenderer {
                 && self.render_state.buffer_time < buffer_duration
             {
                 let position = self.render_state.buffer_time * sampling_ratio;
-                self.positions[index] = position;
+                let playhead = position * sample_rate;
+                let playhead_floored = playhead.floor();
+                let prev_index = playhead_floored as usize; // can't be < 0.
+                let k = (playhead - playhead_floored) as f32;
+
+                self.playback_infos[index] = Some((prev_index, k));
             } else {
-                self.positions[index] = -1.;
+                self.playback_infos[index] = None;
             }
 
             let time_incr = dt * computed_playback_rate;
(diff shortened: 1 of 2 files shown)


orottier:
Hey nice, I'm seeing improvements too

```
- 2       | Simple source test without resampling (Mono)                             | 59            | 2033.9x               | 120
- 3       | Simple source test without resampling (Stereo)                           | 67            | 1791.0x               | 120
- 4       | Simple source test without resampling (Stereo and positionnal)           | 196           | 612.2x                | 120
- 5       | Simple source test with resampling (Mono)                                | 54            | 2222.2x               | 120
- 6       | Simple source test with resampling (Stereo)                              | 66            | 1818.2x               | 120
- 7       | Simple source test with resampling (Stereo and positionnal)              | 191           | 628.3x                | 120
```

Compared to current main
```
- 2       | Simple source test without resampling (Mono)                             | 61            | 1967.2x               | 120
- 3       | Simple source test without resampling (Stereo)                           | 82            | 1463.4x               | 120
- 4       | Simple source test without resampling (Stereo and positionnal)           | 203           | 591.1x                | 120
- 5       | Simple source test with resampling (Mono)                                | 57            | 2105.3x               | 120
- 6       | Simple source test with resampling (Stereo)                              | 83            | 1445.8x               | 120
- 7       | Simple source test with resampling (Stereo and positionnal)              | 209           | 574.2x                | 120
```

Exciting news - moreover when we can apply it to other nodes.

I have not looked into the code changes yet, will do tomorrow

src/node/audio_buffer_source.rs
```diff
@@ -506,64 +518,108 @@ impl AudioProcessor for AudioBufferSourceRenderer {
...
+                    // check if end is within the block
+                    let res = self.playback_infos.iter().position(|&el| el.is_none());
+                    end_index = match res {
+                        Some(position) => start_index + position,
+                        None => start_index + RENDER_QUANTUM_SIZE,
+                    };
+                }
+                None => {
+                    start_index = 0;
+                    end_index = 0;
```
orottier, line 554: Are you sure `end_index` must be zero? Setting it to 128 does not fix the build by the way..

src/node/audio_buffer_source.rs
```diff
@@ -506,64 +518,108 @@ impl AudioProcessor for AudioBufferSourceRenderer {
                 && self.render_state.buffer_time < buffer_duration
             {
                 let position = self.render_state.buffer_time * sampling_ratio;
-                self.compute_playback_at_position(position, sample_rate);
+                let playhead = position * sample_rate;
+                let playhead_floored = playhead.floor();
+                let prev_index = playhead_floored as usize; // can't be < 0.
+                let k = (playhead - playhead_floored) as f32;
+
+                self.playback_infos[index] = Some((prev_index, k));
```
orottier, line 526: There's something really strange going on numerically, if I render a 3*128 buffer. In the second chunk that is rendered, playback_info contains the index `183` twice!

b-ma: hum that sounds strange indeed, but it would be interesting to see the value of `k` too because it could just be due to floating point errors I guess. I will investigate

PR #168, opened 2022-07-05
Proof of concept for MediaElement implementation with `creek` backend

orottier:
Creek does the heavy lifting of buffering in a dedicated decoding thread, making the rest realtime safe

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ float_eq = "0.7"
 smallvec = "1.8"
 once_cell = "1.10"
 symphonia = { version = "0.5", default-features = false }
+creek = "0.2.1"
 
 [dev-dependencies]
 rand = "0.8"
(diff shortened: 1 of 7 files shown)


orottier:
This will do for a work in progress, what's left:
- await dependency upgrade from upstream (see linked PR)
- tidy up API (remove &mut calls, revamp Options)
- add docs
- implement full MediaElement API: https://html.spec.whatwg.org/multipage/media.html#media-elements - let's leave that for a separate issue
- the `src` must be a file for now (and cannot be altered), would be cool if we could read from HTTP as well

b-ma:
Hey, looks pretty nice with a clean API

Just wondering what happen (weird edge case I agree) if you put the same `media` in two `media_element_source_node`? In my last implementation it was just playing twice the speed... :)

> the src must be a file for now (and cannot be altered), would be cool if we could read from HTTP as well

Would be really fancy yes!

orottier:
> Just wondering what happen (weird edge case I agree) if you put the same `media` in two `media_element_source_node`? In my last implementation it was just playing twice the speed... :)

I decided to panic in that case. In a later implementation we can actually make it work nicely and just duplicate the audio stream

Issue #170, opened 2022-07-09
OfflineAudioContext and AudioContext are not `Send` and `Sync`


william-gooch:
Since the default `AudioContext` implements `!Send + !Sync`, it's currently impossible to run the `decode_audio_data_sync` function in a separate thread, since the context can't be passed to the new thread.
Even the `OfflineAudioContext` implements `!Sync` so it can't be shared between multiple threads, for example to decode multiple files at the same time.

However, the only place where the context is actually used is for retrieving its sample rate. Perhaps it would be useful to expose a bare version of this function that just takes the sample rate itself, which opens it up to being used across multiple threads, for example to load multiple samples concurrently?

b-ma:
Hey, actually you can just create a new `OfflineAudioContest` in the new thread

Adapting your test without introducing a non spec new method, just requires these small changes then: 

```rs
fn test_concurrent_decoding() {
    let context = AudioContext::new(Default::default());
    let sample_rate = context.sample_rate();

    let handles = (0..5).into_iter()
        .map(|_| {
            std::thread::spawn(move || {
                let offline = OfflineAudioContext::new(1, 0, sample_rate);
                let file = std::fs::File::open("samples/sample.wav").unwrap();
                let audio_buffer = offline.decode_audio_data_sync(file).unwrap();
                audio_buffer
            })
        })
        .collect::<Vec<_>>();

    handles
        .into_iter()
        .for_each(|handle| {
            let audio_buffer = handle.join().unwrap();

            assert_eq!(audio_buffer.sample_rate(), sample_rate);
            assert_eq!(audio_buffer.length(), 142_187);
            assert_eq!(audio_buffer.number_of_channels(), 2);
            assert_float_eq!(audio_buffer.duration(), 3.224, abs_all <= 0.001);
            // ...
        });
}
```

Maybe would be nice to make it an example though, for future persons that would have the same question

orottier:
Hi @william-gooch, thanks for flagging this!
There is no principal reason for the AudioContext and OfflineAudioContext not to be Send and Sync, so I will start working on fixing that. It will be great for ease of use of the library.
For now I think the solution from @b-ma will do fine. The OfflineAudioContext is very lightweight so you can safely instantiate many of them.

orottier:
I have renamed the issue a bit, and I came up with a fix for the `OfflineAudioContext` in #172 

[...]

PR #171, opened 2022-07-09
Add `from_encoded` constructor to `AudioBuffer`.

william-gooch:
I've attempted #170 here, although I'm not entirely sure what the API should look like so I've just made a couple educated guesses.


orottier:
Closing in favour of #172  and #174, thanks for getting the ball rolling!

PR #172, opened 2022-07-12
Make OfflineAudiocontext `Send` and `Sync`

orottier:
Regarding the discussion at #170
This is the first step, with a bit of reasoning, a safety wrapper and a small API change we can make the OfflineAudiocontext `Send` and `Sync`

### DIFF
--- a/examples/benchmarks.rs
+++ b/examples/benchmarks.rs
@@ -47,7 +47,7 @@ fn get_buffer(sources: &[AudioBuffer], sample_rate: f32, number_of_channels: usi
 fn benchmark<'a>(
     stdout: &mut termion::raw::RawTerminal<std::io::Stdout>,
     name: &'a str,
-    context: &mut OfflineAudioContext,
+    context: OfflineAudioContext,
     results: &mut Vec<BenchResult<'a>>,
 ) {
     write!(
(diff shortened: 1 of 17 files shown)


src/context/mod.rs
```diff
@@ -120,6 +120,12 @@ mod tests {
         require_send_sync_static(registration);
     }
 
+    #[test]
+    fn test_offline_audio_context_send_sync() {
```
orottier, line 124: Here's a nice (compile time) regression test for the issue

examples/decode_multithreaded.rs
```diff
@@ -0,0 +1,34 @@
+use float_eq::assert_float_eq;
+use std::sync::Arc;
+use web_audio_api::context::{BaseAudioContext, OfflineAudioContext};
+
+pub fn main() {
```
orottier, line 5: I will add a better example than this, using a threadpool. If you spawn a thread per audio file things may run out of hand

examples/benchmarks.rs
```diff
@@ -97,43 +97,43 @@ fn main() {
     {
         let name = "Baseline (silence)";
 
-        let mut context = OfflineAudioContext::new(2, DURATION * sample_rate as usize, sample_rate);
+        let context = OfflineAudioContext::new(2, DURATION * sample_rate as usize, sample_rate);
```
b-ma, line 100: Really like we can drop the `mut`

src/context/offline.rs
```diff
@@ -70,15 +94,15 @@ impl OfflineAudioContext {
...
+            renderer: SingleUseRenderThread::new(renderer),
         }
     }
 
     /// Given the current connections and scheduled changes, starts rendering audio.
     ///
     /// This function will block the current thread and returns the rendered `AudioBuffer`
     /// synchronously. An async version is currently not implemented.
-    pub fn start_rendering_sync(&mut self) -> AudioBuffer {
+    pub fn start_rendering_sync(self) -> AudioBuffer {
```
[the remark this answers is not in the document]
b-ma: ok seems quite logical indeed

Issue #173, opened 2022-07-13
Sounds stop after a few seconds on Raspberry Pi


b-ma:
Pretty weird things actually happen on Raspberry Pi, after a while (few seconds in general), the render thread goes to 100% and the sound just stops.

I started to investigate:
- benchmarks looks ok (expect for the granular, all benches are faster than real-time with quite decent number), which seems to show that the problem is lower-level and related to the audio thread
- created an idiot test bed, actually just an empty `AudioContext` with no nodes and the problem still occurs
I saw that `RenderThread::render` is called with buffers of 64 samples (sometime really weird numbers but always ratio of 64 at the end), while it should be 256 for stereo and the config given to `cpal` looks good. In any case, I guess there are some intertwined problems here because even if there is an issue with `cpal`, it doesn't really explain why the render thread seems to go into an infinite loop...

I will continue to investigate on that... e.g. try to use `Jack` instead of pure alsa as a backend to see if it makes things better

Any ideas welcome :)

orottier:
Interesting, how does the memory usage progress over time?

b-ma:
It's like normal, and at one point goes straight to 100% (as far as I can tell with htop)

Actually things seems to be due to the buffer size 128, if I increase it (e.g. to 512) things start to get real better. I even managed to run the granular example ! (will run it at lunch time to see how it behaves for a longer period)

I will also try with jack (or maybe pipewire) see if its related to alsa

orottier:
Okay that is good. A lower sample rate might also alleviate the problem (but won't fix it!)

b-ma:
hum, I was a bit optimistic... 512 also crash... I will try all increasing values one by one

> A lower sample rate might also alleviate the problem (but won't fix it!)

Yup, but actually when it runs, it does not take that much cpu (I'm quite impressed on that point), I really think it's more an alsa / cpal issue

b-ma:
1024 worked for 2 hours without any drop!

b-ma:
I think we can close this one as the problem appears to be solved when using at least `AudioContextLatencyCategory::Playback` (1024 samples, i.e. ~20ms of latency at 48kHz) or even larger latency values with `AudioContextLatencyCategory::Custom`.

PR #175, opened 2022-07-15
Fix/latency category

b-ma:
Hey, here are some changes regarding #173, to put more relaxed values for `Balanced` (512) and `Playback` (1024) latency hints. I think we could even multiply these by 2 but no strong opinion neither, it a bit a finger in the wind...

Also:
- refactored a bit `StreamConfigsBuilder::with_latency_hint` to favor power of two buffers sizes 
- renamed `AudioContextLatencyCategory::Specific` to `AudioContextLatencyCategory::Custom` which I find more clear (can revert that easily if you don't agree)

Some changes on the granular example too (with a proto lookahead scheduler!) so that timing is more precise and does not rely anymore on the underlying io buffer size.

### DIFF
--- a/examples/granular.rs
+++ b/examples/granular.rs
@@ -1,3 +1,5 @@
+use rand::rngs::ThreadRng;
+use rand::Rng;
 use std::fs::File;
 use std::{thread, time};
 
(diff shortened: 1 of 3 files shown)


src/io.rs
```diff
@@ -127,53 +127,70 @@ impl StreamConfigsBuilder {
...
     #[allow(clippy::needless_pass_by_value)]
-    fn with_latency_hint(&mut self, v: AudioContextLatencyCategory) {
-        let buffer_size: u32 = u32::try_from(RENDER_QUANTUM_SIZE).unwrap();
-        let default_buffer_size = match self.supported.buffer_size() {
-            SupportedBufferSize::Range { min, .. } => buffer_size.max(*min),
-            SupportedBufferSize::Unknown => buffer_size,
+    fn with_latency_hint(&mut self, latency_hint: AudioContextLatencyCategory) {
+        let render_quantum_size = u32::try_from(RENDER_QUANTUM_SIZE).unwrap();
+
+        // at 44100Hz sample rate (this could be even more relaxed):
```
orottier, line 140: These changes look good to me. Indeed the previous values were quite tight, I think we could even relax the 'playback' variant later.

examples/granular.rs
```diff
@@ -1,3 +1,5 @@
+use rand::rngs::ThreadRng;
```
orottier, line 1: Where can I read more about your granular synthesis project? I'm intrigued

b-ma: That's quite a large topic I guess, but the general idea is quite simple at the end, just create some sound by concatenating small pieces of sound:
- https://en.wikipedia.org/wiki/Granular_synthesis
- http://erickuehnl.com/1995/12/30/granular-synthesis/

A very important aspect to achieve clean results is to have very accurate scheduling, that's why we need some scheduler to work with a logical time that is decoupled from the audio io callbacks and other timing functions. You can find a good explanation of the why and how there: https://web.dev/audio-scheduling/, the article is very JS oriented, but I think the main ideas holds for any language, i.e. you can assume you would have the same inaccuracy and drifting problems with `thread::sleep(...);` (I tested, it does :)

b-ma: If you want to play with an interactive example: https://apps.ismm.ircam.fr/demos/mini-catart (just click and move over the dots, the "Hendrix" file works quite well). It's a slightly different approach, called concatenative synthesis, where the consecutive grains are chosen according to some analysis results (rms and pitch/noisiness in the example)

The other examples I have are currently broken for some reason, but I could update them actually

PR #176, opened 2022-07-15
Improve `DelayNode` performance

b-ma:
Hey, implemented the same strategy as for the `AudioScheduledSourceNode` to access memory in a more aligned way, also added a bench to check the improvement (~1,3x on my machine)

**before**

```
+ Simple source test without resampling (Stereo)                           | 115           | 1043.5x               | 120
+ Stereo source with delay                                                 | 317           | 378.5x                | 120
```

**after**

```
+ Simple source test without resampling (Stereo)                           | 108           | 1111.1x               | 120
+ Stereo source with delay                                                 | 247           | 485.8x                | 120
```

Nice side effect is that it allowed to remove `RenderQuantum::set_channels_values_at`

### DIFF
--- a/src/node/delay.rs
+++ b/src/node/delay.rs
@@ -425,43 +425,44 @@ impl AudioProcessor for DelayReader {
         let ring_size = ring_buffer.len() as i32;
         let ring_index = self.index as i32;
 
-        // compute all infos for rendering
-        for (index, delay) in delay_param.iter().enumerate() {
-            // param is already clamped to max_delay_time internally, so it is
-            // safe to only check lower boundary
-            let clamped_delay = (*delay as f64).max(quantum_duration);
-            let num_samples = clamped_delay * sample_rate;
-            // negative position of the playhead relative to this block start
-            let position = index as f64 - num_samples;
-
-            // find address of the frame in the ring buffer just before `position`
-            let prev_position = position.floor();
-            let (prev_block_index, prev_frame_index) =
-                DelayReader::find_frame_adress_at_position(prev_position, ring_size, ring_index);
-
-            // find address of the frame in the ring buffer just after `position`
-            let next_position = position.ceil();
-            let (next_block_index, next_frame_index) =
-                DelayReader::find_frame_adress_at_position(next_position, ring_size, ring_index);
-
-            // as position is negative k will be what we expect
-            let k = (position - position.floor()) as f32;
-
-            self.playback_infos[index] = PlaybackInfo {
-                prev_block_index,
-                prev_frame_index,
-                next_block_index,
-                next_frame_index,
-                k,
-            };
-        }
-
         // render channels aligned
         for (channel_number, channel) in output.channels_mut().iter_mut().enumerate() {
             channel
                 .iter_mut()
-                .zip(self.playback_infos.iter())
-                .for_each(|(o, infos)| {
+                .enumerate()
+                .zip(delay_param.iter())
+                .zip(self.playback_infos.iter_mut())
+                .for_each(|(((index, o), delay), infos)| {
+                    if channel_number == 0 {
+                        // param is already clamped to max_delay_time internally, so it is
+                        // safe to only check lower boundary
+                        let clamped_delay = (*delay as f64).max(quantum_duration);
+                        let num_samples = clamped_delay * sample_rate;
+                        // negative position of the playhead relative to this block start
+                        let position = index as f64 - num_samples;
+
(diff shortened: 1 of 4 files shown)


src/node/delay.rs
```diff
@@ -267,8 +264,8 @@ struct DelayReader {
     last_written_index: Rc<Cell<Option<usize>>>,
     // local copy of shared `last_written_index` so as to avoid render ordering issues
     last_written_index_checked: Option<usize>,
-    // internal buffer used to compute output per channel at each frame
-    internal_buffer: Vec<f32>,
+    // internal buffer used to compute output delay infos for each sample
+    playback_infos: [(usize, usize, usize, usize, f32); RENDER_QUANTUM_SIZE],
```
orottier, line 268: It may be nice to introduce a struct for this, for readability
```
struct PlaybackInfo {
                prev_block_index: usize,
                prev_frame_index: usize,
                next_block_index: usize,
                next_frame_index: usize,
                k: f32,
}
```

src/node/delay.rs
```diff
@@ -425,31 +422,46 @@ impl AudioProcessor for DelayReader {
...
+                prev_frame_index,
+                next_block_index,
+                next_frame_index,
+                k,
+            );
+        }
 
-            // populate output at index w/ internal_buffer
-            output.set_channels_values_at(index, &self.internal_buffer);
+        // render channels aligned
```
orottier, line 444: Same for the AudioBufferSourceNode, I think this two-step way could be greatly improved (performance and lines of code-wise) by doing all in one step. But's let's improve gradually and merge this in the mean time

orottier:
Benchmark before:
`- 21      | Stereo source with delay                                                 | 152           | 789.5x                | 120`
after:
`- 21      | Stereo source with delay                                                 | 116           | 1034.5x               | 120`

indeed nice

src/node/delay.rs
b-ma: hum lost my comment...
done!

src/node/delay.rs
b-ma: Did it on the DelayNode, seems to improve pref a bit indeed

src/node/delay.rs
```diff
@@ -30,6 +30,15 @@ impl Default for DelayOptions {
     }
 }
 
+#[derive(Copy, Clone)]
```
orottier, line 33: deriving `Default` would do nicely to prevent the large constructor down below

orottier (commented):
Other than the two small fixes suggested, good to go!

src/node/delay.rs
```diff
@@ -267,8 +279,8 @@ struct DelayReader {
     last_written_index: Rc<Cell<Option<usize>>>,
     // local copy of shared `last_written_index` so as to avoid render ordering issues
     last_written_index_checked: Option<usize>,
-    // internal buffer used to compute output per channel at each frame
-    internal_buffer: Vec<f32>,
+    // internal buffer used to compute output delay infos for each sample
+    playback_infos: [PlaybackInfo; RENDER_QUANTUM_SIZE],
```
orottier, line 283: Because you are not reusing any of these values between render quanta, this does not need to be an attribute of the DelayReader, but can just be a local variable in de `process` call. Saves a bit of memory

src/node/delay.rs
b-ma: of course, done!

src/node/delay.rs
b-ma: ok done! 

This is a bit counter intuitive for me, as I have the impression that we reallocate the buffer each time, but I guess that's a JS deformation as we work only with the heap here...

b-ma: (did it on the `AudioBufferSourceNode` as well)

orottier: Yes, that is probably very different in JS.
In Rust, an array `[T]` is always on the stack so there is no 're-allocation'
The only types on the heap in Rust are:
- collections (Vec, String, HashMap, etc - https://doc.rust-lang.org/stable/std/collections/ )
- pointer types (Box, Arc, Rc)

src/node/delay.rs
orottier: FYI if the default only contains zeroes you can do this:
https://github.com/orottier/web-audio-api-rs/commit/fcbce99b56d6eb8c1bb8b812312d51e89f06c532

b-ma: More simple indeed :) thanks !

PR #177, opened 2022-07-16
Draft: async interface (decodeAudioData, startRendering, suspend/resume/close)

orottier:
This is in no way meant to be merged soon, just a conversation starter. I am checking what needs to be done to introduce functions that return a `Promise` in the spec.

### DIFF
--- a/src/context/offline.rs
+++ b/src/context/offline.rs
@@ -100,8 +100,42 @@ impl OfflineAudioContext {
 
     /// Given the current connections and scheduled changes, starts rendering audio.
     ///
+    /// Rendering an OfflineAudioContext is CPU-bound (involves no IO, and no `.await` points) so
+    /// for large renderings you will probably want to run this future on a 'blocking thread', e.g.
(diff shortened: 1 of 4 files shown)


src/context/offline.rs
```diff
@@ -98,10 +98,44 @@ impl OfflineAudioContext {
         }
     }
 
+    /// Given the current connections and scheduled changes, starts rendering audio.
+    ///
+    /// Rendering an OfflineAudioContext is CPU-bound (involves no IO, and no `.await` points) so
```
orottier, line 103: We should probably add a call to https://docs.rs/tokio/latest/tokio/task/fn.yield_now.html for `N` rendered samples. `N` yet to be determined

orottier:
Closing in favour of
- Suspend, resume, close were properly introduced in https://github.com/orottier/web-audio-api-rs/pull/419 and https://github.com/orottier/web-audio-api-rs/pull/412
- new issue #423 for decodeAudioData

PR #180, opened 2022-07-18
Implement actively processing for node without tail time

b-ma:
Some work related to #147, remaining nodes with tail time should be handled once #34 is done

Let me know what you think!

### DIFF
--- a/src/node/delay.rs
+++ b/src/node/delay.rs
@@ -418,6 +418,11 @@ impl AudioProcessor for DelayReader {
 
         let mut playback_infos = [PlaybackInfo::default(); RENDER_QUANTUM_SIZE];
 
+        // A DelayNode in a cycle is actively processing only when the absolute
+        // value of any output sample for the current render quantum is greater
+        // than or equal to 2−126.
+        let mut is_actively_processing = false;
+
         // render channels aligned
         for (channel_number, channel) in output.channels_mut().iter_mut().enumerate() {
             channel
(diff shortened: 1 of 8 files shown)


b-ma:
shows some improvement on granular benchmark, ~1.3x on my side

src/node/delay.rs
```diff
@@ -477,10 +482,20 @@ impl AudioProcessor for DelayReader {
                     let next_sample = ring_buffer[next_block_index].channel_data(channel_number)
                         [next_frame_index];
 
+                    let value = (1. - k) * prev_sample + k * next_sample;
+
+                    if value.abs() >= 2e-126 {
+                        is_actively_processing = true;
+                    }
+
                     *o = (1. - k) * prev_sample + k * next_sample;
```
orottier, line 491: The compiler will fix this for you, but please use the `value` variable

src/node/delay.rs
```diff
@@ -418,6 +418,11 @@ impl AudioProcessor for DelayReader {
 
         let mut playback_infos = [PlaybackInfo::default(); RENDER_QUANTUM_SIZE];
 
+        // A DelayNode in a cycle is actively processing only when the absolute
+        // value of any output sample for the current render quantum is greater
+        // than or equal to 2−126.
+        let mut is_actively_processing = false;
```
orottier, line 424: Such a strange requirement of the spec. When will this ever be the case? And why not do this for all nodes that process inputs? Ah well..

orottier (approved):
Great stuff, very few lines of code!

src/node/delay.rs
orottier: Good points

orottier: Coming back to this, the smallest possible f32 value is 1.17549435E-38f32 which is exactly 2^-126 (which is not the same as 2E-126!)
https://doc.rust-lang.org/std/primitive.f32.html#associatedconstant.MIN_POSITIVE
So the spec just tells us it is fine to clip to zero when the value is too small to be represented by a f32. So we can drop the check

orottier: I went with c3439f99619f5d2229ca041982b5

b-ma: nice catch!

PR #181, opened 2022-07-20
Implement fast track for `AudioBufferSourceNode` when buffer is aligned on output

b-ma:
Pretty happy with this one, and quite huge performance improvements :)

### before
```
- 1       | Baseline (silence)                                                       | 29            | 4137.9x               | 120
- 2       | Simple source test without resampling (Mono)                             | 87            | 1379.3x               | 120
- 3       | Simple source test without resampling (Stereo)                           | 108           | 1111.1x               | 120
- 4       | Simple source test without resampling (Stereo and positionnal)           | 275           | 436.4x                | 120
- 5       | Simple source test with resampling (Mono)                                | 85            | 1411.8x               | 120
- 6       | Simple source test with resampling (Stereo)                              | 105           | 1142.9x               | 120
- 7       | Simple source test with resampling (Stereo and positionnal)              | 287           | 418.1x                | 120
- 8       | Upmix without resampling (Mono -> Stereo)                                | 97            | 1237.1x               | 120
- 9       | Downmix without resampling (Stereo -> Mono)                              | 95            | 1263.2x               | 120
- 10      | Simple mixing (100x same buffer) - be careful w/ volume here!            | 1694          | 17.7x                 | 30
- 11      | Simple mixing (100 different buffers) - be careful w/ volume here!       | 1730          | 17.3x                 | 30
- 12      | Simple mixing with gains                                                 | 399           | 300.8x                | 120
- 13      | Granular synthesis                                                       | 4313          | 1.7x                  | 7.5
- 14      | Synth (Sawtooth with Envelope)                                           | 6569          | 18.3x                 | 120
- 15      | Synth (Sawtooth with gain - no automation)                               | 5480          | 21.9x                 | 120
- 16      | Synth (Sawtooth without gain)                                            | 2753          | 43.6x                 | 120
- 17      | Substractive Synth                                                       | 569           | 210.9x                | 120
- 18      | Stereo panning                                                           | 175           | 685.7x                | 120
- 19      | Stereo panning with automation                                           | 167           | 718.6x                | 120
- 20      | Sawtooth with automation                                                 | 82            | 1463.4x               | 120
- 21      | Stereo source with delay                                                 | 249           | 481.9x                | 120
```

### after
```
- 1       | Baseline (silence)                                                       | 26            | 4615.4x               | 120
- 2       | Simple source test without resampling (Mono)                             | 49            | 2449.0x               | 120
- 3       | Simple source test without resampling (Stereo)                           | 66            | 1818.2x               | 120
- 4       | Simple source test without resampling (Stereo and positionnal)           | 238           | 504.2x                | 120
- 5       | Simple source test with resampling (Mono)                                | 83            | 1445.8x               | 120
- 6       | Simple source test with resampling (Stereo)                              | 108           | 1111.1x               | 120
- 7       | Simple source test with resampling (Stereo and positionnal)              | 276           | 434.8x                | 120
- 8       | Upmix without resampling (Mono -> Stereo)                                | 54            | 2222.2x               | 120
- 9       | Downmix without resampling (Stereo -> Mono)                              | 69            | 1739.1x               | 120
- 10      | Simple mixing (100x same buffer) - be careful w/ volume here!            | 1715          | 17.5x                 | 30
- 11      | Simple mixing (100 different buffers) - be careful w/ volume here!       | 1741          | 17.2x                 | 30
- 12      | Simple mixing with gains                                                 | 409           | 293.4x                | 120
- 13      | Granular synthesis                                                       | 4148          | 1.8x                  | 7.5
- 14      | Synth (Sawtooth with Envelope)                                           | 5634          | 21.3x                 | 120
- 15      | Synth (Sawtooth with gain - no automation)                               | 4826          | 24.9x                 | 120
- 16      | Synth (Sawtooth without gain)                                            | 2501          | 48.0x                 | 120
- 17      | Substractive Synth                                                       | 547           | 219.4x                | 120
- 18      | Stereo panning                                                           | 132           | 909.1x                | 120
- 19      | Stereo panning with automation                                           | 135           | 888.9x                | 120
- 20      | Sawtooth with automation                                                 | 82            | 1463.4x               | 120
- 21      | Stereo source with delay                                                 | 192           | 625.0x                | 120
```

The only thing I didn't manage to do is to assert the correct path (slow or fast track) from the unit test. My idea was to declare some config variable in the unit test and check it in the code to fail if we are in the wrong path, but I actually don't know if its even possible... if you have any idea, it's welcome!

I think I've maybe found a bug with the loop in the slow track, I will check that later

### DIFF
--- a/src/node/audio_buffer_source.rs
+++ b/src/node/audio_buffer_source.rs
@@ -462,6 +462,16 @@ impl AudioProcessor for AudioBufferSourceRenderer {
                 || current_time + block_duration > stop_time
             {
                 let buffer_time = self.render_state.buffer_time;
+                let end_index = if current_time + block_duration > stop_time
+                    || self.render_state.buffer_time + block_duration > duration
+                {
+                    let dt =
+                        (stop_time - current_time).min(duration - self.render_state.buffer_time);
+                    let end_buffer_time = self.render_state.buffer_time + dt;
+                    (end_buffer_time * sample_rate).round() as usize
+                } else {
+                    buffer.length()
+                };
                 let mut apply_offset = 0;
 
                 buffer
@@ -472,16 +482,6 @@ impl AudioProcessor for AudioBufferSourceRenderer {
                         // we need to recompute that for each channel
                         let buffer_channel = buffer_channel.as_slice();
                         let mut start_index = (buffer_time * sample_rate).round() as usize;
-                        let end_index = if current_time + block_duration > stop_time
-                            || self.render_state.buffer_time + block_duration > duration
-                        {
-                            let dt = (stop_time - current_time)
-                                .min(duration - self.render_state.buffer_time);
-                            let end_buffer_time = self.render_state.buffer_time + dt;
-                            (end_buffer_time * sample_rate).round() as usize
-                        } else {
-                            buffer.length()
-                        };
                         let mut offset = 0;
 
                         for (index, o) in output_channel.iter_mut().enumerate() {
--- a/src/node/audio_buffer_source.rs
+++ b/src/node/audio_buffer_source.rs
@@ -661,28 +661,53 @@ mod tests {
         let context = OfflineAudioContext::new(2, RENDER_QUANTUM_SIZE, 44_100.);
 
         let file = std::fs::File::open("samples/sample.wav").unwrap();
-        let audio_buffer = context.decode_audio_data_sync(file).unwrap();
+        let expected = context.decode_audio_data_sync(file).unwrap();
 
-        let src = context.create_buffer_source();
-        src.set_buffer(audio_buffer.clone());
-        src.connect(&context.destination());
-        src.start_at(context.current_time());
-        src.stop_at(context.current_time() + 128.);
+        // 44100 will go through fast track
+        // 48000 will go through slow track
+        [44100, 48000].iter().for_each(|sr| {
+            let decoding_context = OfflineAudioContext::new(2, RENDER_QUANTUM_SIZE, *sr as f32);
 
-        let res = context.start_rendering_sync();
+            let mut filename = "samples/sample-".to_owned();
+            filename.push_str(&sr.to_string());
+            filename.push_str(".wav");
 
-        // check first 128 samples in left and right channels
-        assert_float_eq!(
-            res.channel_data(0).as_slice()[..],
-            audio_buffer.get_channel_data(0)[0..128],
-            abs_all <= 0.
-        );
+            let file = std::fs::File::open("samples/sample.wav").unwrap();
+            let audio_buffer = decoding_context.decode_audio_data_sync(file).unwrap();
 
-        assert_float_eq!(
-            res.channel_data(1).as_slice()[..],
-            audio_buffer.get_channel_data(1)[0..128],
-            abs_all <= 0.
-        );
+            assert_eq!(audio_buffer.sample_rate(), *sr as f32);
+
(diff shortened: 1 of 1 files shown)


b-ma:
> You could introduce an attribute on the AudioBufferSourceOptions: panic_when_unaligned (cfg(test) only), and pass it along to the Renderer - which should than panic in the unaligned code path. But that's a lot of effort and you are sprinkling test code around which make reading harder on the eyes

Hum, seems a bit complicated indeed... let's consider the benchmarks are enough to spot regressions

src/node/audio_buffer_source.rs
```diff
@@ -436,6 +432,113 @@ impl AudioProcessor for AudioBufferSourceRenderer {
...
+                        };
+                        let mut offset = 0;
+
+                        for (index, o) in output_channel.iter_mut().enumerate() {
+                            let mut buffer_index = start_index + index - offset;
+
+                            *o = if buffer_index < end_index {
+                                buffer_channel[buffer_index]
+                            } else {
+                                if loop_ && buffer_index == end_index {
```
orottier, line 493: Looping is not allowed in this code path, right? Do we need this snippet?

src/node/audio_buffer_source.rs
```diff
@@ -436,6 +432,113 @@ impl AudioProcessor for AudioBufferSourceRenderer {
...
+                                if loop_ {
+                                    buffer_channel[buffer_index]
+                                } else {
+                                    0.
+                                }
+                            };
+                        }
+                    });
+
+                if apply_offset != 0 {
```
orottier, line 510: Also, offsets do not apply for this fast path section, no?

src/node/audio_buffer_source.rs
```diff
@@ -436,6 +432,113 @@ impl AudioProcessor for AudioBufferSourceRenderer {
...
+                    self.render_state.buffer_time = ((RENDER_QUANTUM_SIZE - apply_offset) as f64
+                        / sample_rate)
+                        % buffer_duration;
+                } else {
+                    self.render_state.buffer_time += block_duration;
+                }
+            } else {
```
orottier, line 517: With looping and offsets not part of the fast path, I think we only need this following snippet (memcopy).
Of course you still need to pad with zeroes if the end_index < 128 + start_index

src/node/audio_buffer_source.rs
b-ma: No actually, we can loop in the fast path at the only condition that the loop boundaries are the default ones: i.e. `loop_start == 0.` and `loop_end == buffer.duration()`. In this case we are sure that we stay aligned with the output.

This is basically what is done in the "Simple source test without resampling (Mono|Stereo)" benches

b-ma: I agree it's a bit more code, but I think it worth the case. Not a specialist but I think that can be a pretty common situation in games for example

orottier: Ah right, I missed that. I think that is beneficial, yes, let's keep this code path.
My comment about `offset` still stands, right?

b-ma: Yup, I actually tried to simplify but all my attempts introduced weird edge cases and crashed some of the tests...

So, I actually added a comment and changed `apply_offset` to:
```rs
let mut loop_point_index: Option<usize> = None;
```

Hopefully this is more readable, let me know

orottier:
Good stuff!

PR #184, opened 2022-07-20
Fix clamp of AudioParam

b-ma:
Clamping is now done on the computed values after input is added

Really not sure this is the most elegant way, but that's the only one I found to both make the borrow checker happy and to be able to write some tests :)

### DIFF
--- a/src/param.rs
+++ b/src/param.rs
@@ -596,17 +596,15 @@ impl AudioProcessor for AudioParamProcessor {
     ) -> bool {
         let period = 1. / scope.sample_rate as f64;
 
-        let param_intrisic_values_clamped =
-            self.tick(scope.current_time, period, RENDER_QUANTUM_SIZE);
-
         let input = &inputs[0]; // single input mode
-        let param_computed_values = &mut outputs[0];
+        let output = &mut outputs[0];
+        let min = self.min_value;
+        let max = self.max_value;
 
-        param_computed_values
-            .channel_data_mut(0)
-            .copy_from_slice(param_intrisic_values_clamped);
+        let intrisic_values =
+            self.compute_intrisic_values(scope.current_time, period, RENDER_QUANTUM_SIZE);
 
-        param_computed_values.add(input, &AUDIO_PARAM_CHANNEL_CONFIG);
+        AudioParamProcessor::compute_computed_values(intrisic_values, input, output, min, max);
 
         true // has intrinsic value
     }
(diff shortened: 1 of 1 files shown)


src/param.rs
```diff
@@ -615,17 +613,30 @@ impl AudioProcessor for AudioParamProcessor {
...
+        max: f32,
+    ) {
+        output.channel_data_mut(0).copy_from_slice(intrisic_values);
+
+        output.add(input, &AUDIO_PARAM_CHANNEL_CONFIG);
+
+        output
+            .channel_data_mut(0)
+            .iter_mut()
+            .for_each(|s| *s = s.clamp(min, max));
```
b-ma, line 637: should handle "If the sum is NaN, replace the sum with the defaultValue." here

src/param.rs
```diff
@@ -2885,7 +2801,227 @@ mod tests {
         param.set_value_at_time(2., 0.000001);
         param.set_automation_rate(AutomationRate::A);
 
-        let vs = render.tick(0., 1., 10);
+        let vs = render.compute_intrisic_values(0., 1., 10);
         assert_float_eq!(vs, &[2.; 10][..], abs_all <= 0.);
     }
+
+    #[test]
+    fn test_computed_values_clamped() {
```
b-ma, line 2809: All these are a bit redundant, I will try to come up w/ something more simple tomorrow

orottier:
Great stuff. Indeed the API and the tests become a bit clunky now. Maybe just a single method would do, just the old `tick` that performs the intrinsic calc, adding an optional input buffer, and clamping. 
In the test we could just provide `None` or AudioRenderQuantum:: silence for the input buffer argument. 
I could try to that out tomorrow

b-ma:
> Maybe just a single method would do, just the old tick that performs the intrinsic calc, adding an optional input buffer, and clamping. [...] I could try to that out tomorrow

Yup, that seems more simple indeed, ok I let you go on w/ that

Just added a few checks to try to improve perfs, no game changers here, but the tricks are quite simple and seems to improve things a bit.

orottier:
I struggled for quite a while to come up with an elegant solution, which I did not really find in the end.
What changed is that I replaced the free function `AudioParam::compute_computed_values` to just `render.mix_to_output` which looks a bit nicer to me.

b-ma:
> What changed is that I replaced the free function AudioParam::compute_computed_values to just render.mix_to_output which looks a bit nicer to me.

Yup agreed, seeing your solution I don't really understand why I had some problems with the borrow checker in the first place... Still need to learn and understand some Rust guts it seems :)

PR #185, opened 2022-07-22
Replace `std::collections::HashMap` with `rustc_hash::FxHashMap`

b-ma:
Basically everything is in the title. 

Not sure you want to merge this because of the new dependency (i.e. https://docs.rs/rustc-hash/1.1.0/rustc_hash/index.html) and so on. But I just wanted to test something as I saw in a flamegraph of the benchmarks that almost 20% of the time was spent in `graph.nodes.add()` and `graph.nodes.remove()` in `render()`. From what I see, it seems to produce an improvement of ~10% in benches that deal with many nodes, dividing by ~2 the time spent in `graph.nodes.add()` and `graph.nodes.remove()`.

If you have other ideas / plans on this point, let me know if I can help

(I think I have gone through all my ideas according to optimizations now... will try to start on the CompressorNode next week)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ smallvec = "1.8"
 once_cell = "1.10"
 symphonia = { version = "0.5", default-features = false }
 creek = "0.2.1"
+rustc-hash = "1.1.0"
 
 [dev-dependencies]
 rand = "0.8"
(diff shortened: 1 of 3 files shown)


orottier:
Well this is so simple and so effective it would be hard not to merge this.
This is a simple step towards #55 , so feel free to take a stab there if you want to fiddle with `Cell` / ` UnsafeCell`. The main point being that we remove/add nodes during rendering only to satisfy the borrow checker (which is, update a node based on the signal of incoming nodes, so you need both mutable and immutable access to the node map - notoriously hard in Rust)

PR #186, opened 2022-07-24
Generalize over audio backends, add `cubeb` as alternative for `cpal`

orottier:
Just sharing some work in progress, there some work to do still:

- [x] mic input
- [ ] segfault when suspend/close stream
- [x] support all output channel configs
- [ ] validate 4.1 and other output configs
- [ ] support all input channel configs
- [x] latency calculations
- [ ] ...

You can try it out with `cargo run --no-default-features --features cubeb --release --example mono`
For cubeb to run properly, `cmake` must be installed.


README.md
```diff
@@ -85,6 +85,31 @@ It is a work in progress, but eventually we should be able to run the official
 [WebAudioAPI test harness](https://github.com/web-platform-tests/wpt/tree/master/webaudio)
 and track our spec compliance improvements over time.
 
+## Audio backends
+
+By default, the [`cpal`](https://github.com/rustaudio/cpal) library is used for
+cross platform audio I/O. We offer experimental support for the
+[`cubeb`](https://github.com/mozilla/cubeb-rs) backend via the `cubeb` feature
+flag. Please not that `cmake` must be installed locally in order to run `cubeb`.
```
b-ma, line 93: "Please note"

src/io/backend_cpal.rs
```diff
@@ -135,59 +244,14 @@ impl StreamConfigsBuilder {
...
-            }
-        };
+        let buffer_size = super::buffer_size_for_latency_category(
+            latency_hint,
+            self.prefered.sample_rate.0 as f32,
+        ) as u32;
 
         let clamped_buffer_size: u32 = match self.supported.buffer_size() {
             SupportedBufferSize::Unknown => buffer_size,
-            // try to find the best power of two candidate in given range
```
b-ma, line 166: I would have kept this part, it seemed to me that alsa was very happy to crash when the requested `buffer_size` was not a power of two (but this probably needs more rigorous testing...)

b-ma:
As far as I can tell, this looks pretty nice! Having the possibility of no backend is also really a nice to have too

src/io/backend_cpal.rs
orottier: Interesting, but at this point, `buffer_size` is a power of two (1, 4, or 8 times 128).
The only thing now is we clamp it to `(min, max)`. It would be strange if either `min` or `max` is then not supported, no? Because that is what ALSA reports as being available.
If you could at any time have a look at the crash, that would be very beneficial!

README.md
orottier: thanks

src/io/backend_cpal.rs
b-ma: > Interesting, but at this point, buffer_size is a power of two (1, 4, or 8 times 128).

 Yup (or any power of two if custom latency has been given)

> The only thing now is we clamp it to (min, max). It would be strange if either min or max is then not supported, no? Because that is what ALSA reports as being available.

Yup, I agree that appears a bit weird, but if I remember well (which is maybe not the case :) the min / max values reported by alsa I have saw were quite strange

> If you could at any time have a look at the crash, that would be very beneficial!

Yup, I will try to have a closer and more precise look, I was actually hardcoding values for testing so maybe I'm wrong here

orottier:
I'm still seeing segfaults, deadlocks and other obscure stuff when interacting with the `cubeb` stream object (suspend, close, get latency, ..) but for the rest it works really nice.
I think we can merge this as it also contains some improvements on code style and on the cpal backend.
I will open an issue with improvements for the cubeb code for further improvements

PR #189, opened 2022-07-27
Implement dynamic compressor

b-ma:
Here is a first version that seems to work...

Actually, this does not completely follow the letter of the spec (which is really weird I think, over-specifying some internals in linear scale which prevent working with dB...), but hopefully most of the spirit is there. I have mostly followed this paper https://www.eecs.qmul.ac.uk/~josh/documents/2012/GiannoulisMassbergReiss-dynamicrangecompression-JAES2012.pdf, and made some plots of the internal values to make sure it behaves as expected (at least as I understand it... :)

Also, I can ear it does something but I don't really know if it sounds as it is supposed to... I'm not used enough to this effect to be able to tell anything about the quality. I think we will need to create some kind of interactive interface and ask to a trained sound engineer to test it to have useful feedback here. 

I will try to create a prototype in JS, which will allow to compare browsers implementations and the node.js version from the same interface.

### DIFF
--- a/examples/compressor.rs
+++ b/examples/compressor.rs
@@ -5,13 +5,13 @@ use web_audio_api::node::{AudioNode, AudioScheduledSourceNode, DynamicsCompresso
 fn main() {
     env_logger::init();
 
-    println!("> gradually increase the amount of compression applied on the sample");
 
     let context = AudioContext::default();
 
     let file = File::open("samples/sample.wav").unwrap();
     let buffer = context.decode_audio_data_sync(file).unwrap();
 
+    println!("> no compression");
     let src = context.create_buffer_source();
     src.connect(&context.destination());
     src.set_buffer(buffer.clone());
@@ -20,12 +20,19 @@ fn main() {
     // enjoy listening
     std::thread::sleep(std::time::Duration::from_secs(4));
 
-    for i in 0..7 {
+    println!("> compression (hard knee)");
+    println!("+ attack: {:?}ms", 30);
+    println!("+ release: {:?}ms", 100);
+    println!("+ ratio: {:?}", 12);
+    println!(">");
+
+    for i in 0..6 {
+        println!("+ threshold at {:?}", -10. * i as f32);
         let compressor = DynamicsCompressorNode::new(&context, Default::default());
         compressor.connect(&context.destination());
         compressor.threshold().set_value(-10. * i as f32);
         compressor.knee().set_value(0.); // hard knee
-        compressor.attack().set_value(0.05); // hard knee
+        compressor.attack().set_value(0.03); // hard knee
         compressor.release().set_value(0.1); // hard knee
 
         let src = context.create_buffer_source();
--- a/examples/compressor.rs
+++ b/examples/compressor.rs
@@ -5,7 +5,6 @@ use web_audio_api::node::{AudioNode, AudioScheduledSourceNode, DynamicsCompresso
 fn main() {
     env_logger::init();
 
-
     let context = AudioContext::default();
 
     let file = File::open("samples/sample.wav").unwrap();
--- a/examples/compressor.rs
+++ b/examples/compressor.rs
@@ -1,0 +1,39 @@
+use std::fs::File;
+use web_audio_api::context::{AudioContext, BaseAudioContext};
+use web_audio_api::node::{AudioNode, AudioScheduledSourceNode, DynamicsCompressorNode};
+
+fn main() {
+    env_logger::init();
+
+    println!("> gradually increase the amount of compression applied on the sample");
+
+    let context = AudioContext::default();
+
+    let file = File::open("samples/sample.wav").unwrap();
+    let buffer = context.decode_audio_data_sync(file).unwrap();
+
+    let src = context.create_buffer_source();
+    src.connect(&context.destination());
+    src.set_buffer(buffer.clone());
+    src.start();
+
+    // enjoy listening
+    std::thread::sleep(std::time::Duration::from_secs(4));
+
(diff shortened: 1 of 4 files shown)


orottier:
Exciting! Indeed hard to validate correctness, but I replicated the https://mdn.github.io/webaudio-examples/compressor-example/ example and with my amateur ears it sounds quite similar.

Also https://github.com/web-platform-tests/wpt/tree/master/webaudio/the-audio-api/the-dynamicscompressornode-interface does not provide any meaningful tests - which makes sense because Paul told us the Chrome and Firefox implementations differ anyway

examples/compressor.rs
```diff
@@ -0,0 +1,45 @@
+use std::fs::File;
+use web_audio_api::context::{AudioContext, BaseAudioContext};
+use web_audio_api::node::{AudioNode, AudioScheduledSourceNode, DynamicsCompressorNode};
+
+fn main() {
+    env_logger::init();
+
+    let context = AudioContext::default();
+
+    let file = File::open("samples/sample.wav").unwrap();
```
orottier, line 10: This may not be the most interesting file for a compressor..

src/node/dynamics_compressor.rs
```diff
@@ -0,0 +1,512 @@
...
+        let mut reduction_gains = [0.; 128]; // lin
+        let mut detector_values = [0.; 128]; // lin
+
+        for i in 0..RENDER_QUANTUM_SIZE {
+            // pick highest value for this index across all input channels
+            // @tbc - this seems to be what is done in chrome
+            let mut max = f32::MIN;
+
+            for channel in input.channels().iter() {
+                let sample = channel.as_slice()[i];
```
orottier, line 302: Should you not take the `.abs()` already here? Otherwise you would only consider positive values

src/node/dynamics_compressor.rs
```diff
@@ -0,0 +1,512 @@
...
+    ) -> bool {
+        // single input/output node
+        let input = inputs[0].clone();
+        let output = &mut outputs[0];
+        let sample_rate = scope.sample_rate;
+
+        let ring_size = self.ring_buffer.capacity();
+        // ensure ring buffer is filled with silence
+        if self.ring_buffer.len() < ring_size {
+            let mut silence = input.clone();
```
orottier, line 245: You can call `input.silence()` to create a new silent buffer

src/node/dynamics_compressor.rs
```diff
@@ -0,0 +1,512 @@
...
+impl AudioProcessor for DynamicsCompressorRenderer {
+    fn process(
+        &mut self,
+        inputs: &[AudioRenderQuantum],
+        outputs: &mut [AudioRenderQuantum],
+        params: AudioParamValues,
+        scope: &RenderScope,
+    ) -> bool {
+        // single input/output node
+        let input = inputs[0].clone();
```
orottier, line 238: This clone might not be necessary

src/node/dynamics_compressor.rs
```diff
@@ -0,0 +1,512 @@
...
+        // update reduction shared w/ main thread
+        self.reduction.store(reduction_gain);
+
+        // store input in delay line
+        self.ring_buffer[self.ring_index] = input;
+
+        // apply
+        let read_index = (self.ring_index + 1) % ring_size;
+        let delayed = &self.ring_buffer[read_index];
+
```
orottier, line 358: I think at this point, you could say `*output = delayed.clone()` and then modify these values in place with the `reduction_gains`, no? Would save a few lines

orottier (commented):
All in all this looks really great. Very happy that we fix this omission in our lib

examples/compressor.rs
orottier: Maybe `think-stereo-48000.wav` is already a bit more interesting

src/node/dynamics_compressor.rs
b-ma: This gives the following error:
```
238 |         let input = inputs[0];
    |                     ^^^^^^^^^
    |                     |
    |                     cannot move out of here
    |                     move occurs because `inputs[_]` has type `quantum::AudioRenderQuantum`, which does not implement the `Copy` trait
    |                     help: consider borrowing here: `&inputs[0]`
```

I imagined I needed to clone to grab ownership, so that I can put it in the ring buffer?

src/node/dynamics_compressor.rs
b-ma: Actually no, the `silence()` method is only on the `AudioRenderQuantumChannel`

But I agree having to write these two lines is a bit cumbersome. Maybe `make_silent()` could return `self` so that we could write in one line:

```rs
let silence = channel.clone().make_silent();
```

Which I think looks like a good tradeoff?

src/node/dynamics_compressor.rs
b-ma: Sure, good catch!

examples/compressor.rs
b-ma: I was afraid this file was already hardcore compressed but we can actually ear the difference. I update that

src/node/dynamics_compressor.rs
b-ma: Yup true, but I'm not really sure I'm really convinced with this block sized delay, it's way more simple and efficient but it also introduce differences according to the sample rate. 
Maybe we should apply sub quantum logic (without going into sub-sample interpolation) to stick closer to the 6ms delay whatever the sample rate?

orottier: I'm fine with incremental improvements. Merging this!

Issue #154, opened 2022-04-21
Cancel `AudioParam::set_target_at_time` when close enough to target value


b-ma:
from #20 

> `set_target_at_time` can run forever as it as no "real" end time or value, a possible performance improvement would be to internally cancel the event when `intrisic_value` is close enough to the target (note that Chrome implements such strategy)

b-ma:
```rs
// at the end of the block, avoid doing this for each sample
let diff = target - intrisic_value;
if diff.is_subnormal() {
    intrisic_value = target;
    // insert `set_value` event in timeline at next block
}
```
could do the trick

b-ma:
Fixed in #190

PR #190, opened 2022-07-30
End `set_target_at_time` event when diff to target is subnormal

b-ma:
Related to #154

Not completely sure this is valid for every chip architecture, any idea?

### DIFF
--- a/src/param.rs
+++ b/src/param.rs
@@ -13,6 +13,12 @@ use crate::{AtomicF32, RENDER_QUANTUM_SIZE};
 use crossbeam_channel::{Receiver, Sender};
 use lazy_static::lazy_static;
 
+
+/// For SetTargetAtTime event, that theoreticaly cannot end, if the diff between
+/// the current value and the target is below this threshold, the value is set
+/// to target value and the event is considered ended.
+const SNAP_TO_TARGET: f32 = 1e-10;
+
 // arguments sanity check functions for automation methods
 #[track_caller]
 fn assert_non_negative(value: f64) {
@@ -875,10 +881,11 @@ impl AudioParamProcessor {
                     }
                 }
             }
-            // // [spec] Similarly a NotSupportedError exception MUST be thrown if any
-            // // automation method is called at a time which is contained in [𝑇,𝑇+𝐷), 𝑇
-            // // being the time of the curve and 𝐷 its duration.
-            // // @note - Cancel methods are not automation methods
+
+            // [spec] Similarly a NotSupportedError exception MUST be thrown if any
+            // automation method is called at a time which is contained in [𝑇,𝑇+𝐷), 𝑇
+            // being the time of the curve and 𝐷 its duration.
+            // @note - Cancel methods are not automation methods
             if event.event_type == AudioParamEventType::SetValueAtTime
                 || event.event_type == AudioParamEventType::SetValue
                 || event.event_type == AudioParamEventType::LinearRampToValueAtTime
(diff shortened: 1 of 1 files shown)


src/param.rs
```diff
@@ -1316,7 +1316,36 @@ impl AudioParamProcessor {
                                     diff,
                                     next_block_time,
                                 );
-                                self.intrisic_value = value;
+
+                                let diff = end_value - value;
+
+                                // abort event if diff is subnormal
+                                if diff.is_subnormal() || diff == 0. {
```
b-ma, line 1323: I added this `diff == 0.` check so that it would work with CPU architecture that may not support subnormal (...but does it even exists?). Would be interesting to see the behaviour on your ARM machine

orottier:
Hey, this is definitely the right idea, but don't you think that `1.175e-38` (subnormal cutoff) is a cutoff that is way too small?
We could choose to clip at `0.0001` maybe already? Or ideally, something that is dynamic based on the default/min/max values.
Chrome takes an even bigger cutoff I see now: https://chromium.googlesource.com/chromium/blink/+/refs/heads/main/Source/modules/webaudio/AudioParam.cpp#39

orottier:
Mathematically, the most pure would be to take a multiple of the [timeConstant](https://webaudio.github.io/web-audio-api/#dom-audioparam-settargetattime-timeconstant) value and clip after that much time has passed.
After a single `timeConstant` has passed, the value has reached `1−1/𝑒 (around 63.2%)` of difference required to reach the target. So after 5 times `timesConstant` this will be `1 - 1/e^7` which is 99.9% of the target diff reached. Good enough? And accomodates both for params that have a very large range or very small range

b-ma:
Actually Chrome is (or was maybe? can't find it anymore in the source) doing something similar with the time constant. What I'm really not sure is if it behaves well in every condition, 2 examples:
- if we schedule some `set_target_at_time` with a very very small change on a `GainNode` (e.g. just moving a volume fader by a very small amount), can we be sure this time constant trick alone prevents us to fall into subnormal values?
- if we go from 10100Hz to 100Hz on an Oscillator then, if I understand well, the final clamped step will be 10Hz (10000 * 0.001), which is quite a large step around 100Hz (~1 semitone)

On the other side, having a bigger arbitrary limit could be a solution, but it poses the problem that this limit does not represent the same thing for every param (or even within the same param range if it is a kind of log scale such as frequency), which feels weird to me. 

At least with the subnormal solution the only assumption/promise we make is "no need to continue computing that because we can't" (and we don't want to trash the graph with subnormals), regardless of the param boundaries, of the event details, or of what the param actually represents. It's very probably a bit more expensive because we will compute things that are not really audible for sure but, at least for now, it feels like the more simple and safe solution to me.

src/param.rs
orottier: It seems that my computer does support subnormal numbers. I wrote a small test file but I cannot guarantee that the compiler is doing something magical to avoid sending the f32 calculations to the hardware directly.
I am just so confused about subnormal numbers. They cannot be represented by f32, no? So how can we have a Vec<f32> and have code like this
```
  let diff = end_value - *v;

  if diff.is_subnormal() {
      *v = end_value;
  }
```
if end_value = 1.0 and *v is the next f32 value after 1.0, the diff will be `f32::EPSILON` which is greater than `f32::MIN_VALUE` meaning the check will be useless for almost all cases?

orottier:
> if we go from 10100Hz to 100Hz on an Oscillator then, if I understand well, the final clamped step will be 10Hz (10000 * 0.001), which is quite a large step around 100Hz (~1 semitone)

Yeah good point. If we would use the `timeConstant` method it should indeed be a lot more than 7 periods before snapping to target.

src/param.rs
b-ma: > It seems that my computer does support subnormal numbers.

ok good to know, so probably it's not a good idea to rely on such check to do something else than flushing to zero

> I am just so confused about subnormal numbers

Héhé yeah me too :) but I'm sure I have seen some of them them on my Mac, in the new test the diff was actually "some non zero thing" and `subnormal()` was true, so we must be careful to flush subnormal to zero or perfs could be spoiled on some platforms. From what I understand (...quite badly I have to confess) with fast math compile flag or functions this problem would disappear but it seems to be a nightly feature for now (e.g. https://doc.rust-lang.org/std/intrinsics/fn.fmul_fast.html)

>  we define SNAP_TO_TARGET = 1E-10 or something and avoid all the extra code around subnormal stuff

Ok let's go like that then!

b-ma:
Ok done! We actually still need to check and flush subnormals if target is 0, but seems more logical. I finally think checking the diff was kind of a non sens 

Also found a new bug according to the start_time of SetTarget and SetValueCurve events that is fixed now

orottier:
Great work

Issue #188, opened 2022-07-26
Setup proper benchmark suite


orottier:
- [x] Setup `benches` folder to run - https://doc.rust-lang.org/1.12.1/book/benchmark-tests.html
- [ ] Split the `example/benchmarks.rs` file to separate benchmarks
- [ ] Use criterion to run benchmarks locally
- [x] Run benches in CI
- [x] In pull requests, report benchmark results (compared to current main)
- [ ] Add all benchmarks to the CI benchmarks
- [ ] fix bot not always able to comment on the PR

b-ma:
Seems the failing comment is due to the fact that as I'm opening the PR the used token is the one of my fork that does not have write permissions. 

It seems that one way to fix this would be to use

``` 
on:
  pull_request_target:
    branches: [ main ]
```

But then it seems this creates security issues as the code is built on your behalf... (cf. https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target)

Apparently the right way to handle that is to combine several workflows with different access rights as explained there: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ (didn't read carefully, I will have a more in depth read tomorrow)

PR #192, opened 2022-08-02
Improved perf on `StereoPannerNode`

b-ma:
Hey,

Reviewed the algorithm to access memory in a more aligned way (same idea as for delay node and audio buffer source node)

**before**

```
- 18      | Stereo panning                                                           | 110           | 1090.9x               | 120
- 19      | Stereo panning with automation                                           | 117           | 1025.6x               | 120
```

**after**

```
- 18      | Stereo panning                                                           | 83            | 1445.8x               | 120
- 19      | Stereo panning with automation                                           | 81            | 1481.5x               | 120
```

Also:
- simplified a bit the code
- improved the tests
- improved example


src/node/stereo_panner.rs
```diff
@@ -45,13 +34,56 @@ impl Default for StereoPannerOptions {
...
 }
 
+/// Assert that the channel count is valid for the StereoPannerNode
+/// see <https://webaudio.github.io/web-audio-api/#audionode-channelcount-constraints>
+///
+/// # Panics
+///
+/// This function panics if given count is greater than 2
+///
+fn assert_valid_channel_count(count: usize) {
```
orottier, line 44:
```
#[track_caller]
```

just like the checks in src/lib.rs will give a nicer error message

src/node/stereo_panner.rs
```diff
@@ -148,9 +207,10 @@ impl StereoPannerNode {
 
 /// `StereoPannerRenderer` represents the rendering part of `StereoPannerNode`
 struct StereoPannerRenderer {
-    /// The position of the input in the output’s stereo image.
+    /// Position of the input in the output’s stereo image.
     /// -1 represents full left, +1 represents full right.
     pan: AudioParamId,
+    buffer: [[f32; 2]; 128],
```
orottier, line 213: You are using an temp buffer because rust won't let you get two `channel_data_mut` at the same time? Let's build a helper to fix that

orottier (approved):
Good stuff, I will try to come up with a `stereo_channels_mut` method on AudioRenderQuantum to avoid the temporary buffer (but not now, need to go!)

src/node/stereo_panner.rs
b-ma: No actually, that was kind of how it was made in the first place, i.e.

```
out_data[0][sample_idx] = left;
out_data[1][sample_idx] = right;
```

But in terms of perf this is less good because (at least as I understand it) your writing head has to switch constantly between two memory regions, while with this intermediary buffer, everything is well aligned (that's a bit counterintuitive but that was the idea in the `AudioBufferSourceNode` and `DelayNode ` that were using the now disappeared `RenderQuantum::set_channels_values_at` (https://github.com/orottier/web-audio-api-rs/commit/9a3993054fdf621b3acc494df3c7170a82c3a718#diff-717673885f123988d3c7e106e512029ce25a5aed857fac7e31e4cc04352ce946)

Happy to have another try if you think it worth it but I'm pretty sure the perfs will go down, I really tested a lot of strategies

src/node/stereo_panner.rs
b-ma: Yup! I forgot this one

src/node/stereo_panner.rs
b-ma: Actually, the more I think about it the more I think it's related to this discussion between planar and interleaved layouts. I slowly start to understand why interleaved is better / more simple then... (given that you deal with fixed number of channels)

orottier:
Have a look at https://github.com/orottier/web-audio-api-rs/commit/695d35c2771f3727a7cae2833015cd27aec1cc49
It avoids bound checks (like we have on `main` now), and avoids the double buffer (in your PR).
Could you check perf on your machine? Mine has a slight improvement!

src/node/stereo_panner.rs
orottier: Definitely something to reconsider yes. I really appreciate the amount of work and thoughts you are putting in this stuff!

b-ma:
On my side it seems quite constantly a bit below, e.g.:

```
- 18      | Stereo panning                                                           | 87            | 1379.3x               | 120
- 19      | Stereo panning with automation                                           | 84            | 1428.6x               | 120
```

But nothing terrible neither and the code is de facto more clean, so take the one you prefer!

Be aware that I just removed the term "equal power" from the doc, that's not true I think (at least for mono sources) there is a gain when in the middle that is quite clear in the example. I rechecked and the term actually does not appear in the spec neither, it is an issue in the mdn doc.

b-ma:
Could you send your full benchmarks output after merging these changes? I would be interested to see where we get on your machine after this pack of updates on perfs

orottier:
```
- 1       | Baseline (silence)                                                       | 14            | 8571.4x               | 120
- 2       | Simple source test without resampling (Mono)                             | 22            | 5454.5x               | 120
- 3       | Simple source test without resampling (Stereo)                           | 24            | 5000.0x               | 120
- 4       | Simple source test without resampling (Stereo and positionnal)           | 108           | 1111.1x               | 120
- 5       | Simple source test with resampling (Mono)                                | 42            | 2857.1x               | 120
- 6       | Simple source test with resampling (Stereo)                              | 55            | 2181.8x               | 120
- 7       | Simple source test with resampling (Stereo and positionnal)              | 139           | 863.3x                | 120
- 8       | Upmix without resampling (Mono -> Stereo)                                | 21            | 5714.3x               | 120
- 9       | Downmix without resampling (Stereo -> Mono)                              | 19            | 6315.8x               | 120
- 10      | Simple mixing (100x same buffer) - be careful w/ volume here!            | 972           | 30.9x                 | 30
- 11      | Simple mixing (100 different buffers) - be careful w/ volume here!       | 1018          | 29.5x                 | 30
- 12      | Simple mixing with gains                                                 | 225           | 533.3x                | 120
- 13      | Granular synthesis                                                       | 2170          | 3.5x                  | 7.5
- 14      | Synth (Sawtooth with Envelope)                                           | 2122          | 56.6x                 | 120
- 15      | Synth (Sawtooth with gain - no automation)                               | 1786          | 67.2x                 | 120
- 16      | Synth (Sawtooth without gain)                                            | 1086          | 110.5x                | 120
- 17      | Substractive Synth                                                       | 287           | 418.1x                | 120
- 18      | Stereo panning                                                           | 41            | 2926.8x               | 120
- 19      | Stereo panning with automation                                           | 42            | 2857.1x               | 120
- 20      | Sawtooth with automation                                                 | 50            | 2400.0x               | 120
- 21      | Stereo source with delay                                                 | 83            | 1445.8x               | 120
```

Here you go, this is `main` and this branch combined (will merge in a minute).
My benchmarks are quite noisy, hence #188 but I hope this gives some idea of the improvements

b-ma:
Nice! 

At first sight it looks consistent with the improvements on my side, I will have a closer look tomorrow. But we slowly reach some interesting numbers compared to the ones in the paper :)

Other perf ideas I think we could consider from this point:
- The strategy we discussed in #117, I will try to sketch something next week if I find some time. 
- The problem of add / remove in the graph (I add a very rapid try on that which failed miserably :)
- I think we could also re-consider the discussion about the param length being 1 or 128, because now that we have something such as https://github.com/orottier/web-audio-api-rs/blob/183768249204c94bc0c40722186ba6898eed4731/src/param.rs#L958 we could do interesting things even with a-rate params now, e.g. for gain

```rs
if gain.len() == 1 {
  if gain[0] = 1. {
    *output = input.clone();
    return false;
  } else if gain == 0. {
    *output = input.silence();
    return false;
  }
}
```

orottier:
Thanks for the overview, I added two tickets for your points: #195  #196 

> The problem of add / remove in the graph (I add a very rapid try on that which failed miserably :)
I will have a try!

FYI I was working on a HRTF implementation (using the `hrtf` crate) but I run into many problems. I may open a draft PR any time soon because I'm stuck and might need some help ;)

b-ma:
> FYI I was working on a HRTF implementation (using the hrtf crate) but I run into many problems. I may open a draft PR any time soon because I'm stuck and might need some help ;)

Arg one the most simple problem around... not sure if I can be of any help there but I can at least try :)

PR #197, opened 2022-08-05
Fixes #193

### LINKED ISSUE #193: Wrong behaviour when inserting `LinearRampToValueAtTime` and `ExponentialRampToValueAtTime` events
This check is wrong in some situations:

https://github.com/orottier/web-audio-api-rs/blob/53b1dc0b643d65e1695cff30ce0137e95538990e/src/param.rs#L918

For example if we do something like:

```rs
param.linear_ramp_to_value_at_time(1., context.current_time() + 1.);
std::thread::sleep(std::time::Duration::from_secs(4));
// first event has ended thus the timeline is empty and `last_event` is defined
param.linear_ramp_to_value_at_time(2., context.current_time() + 1.);
```
The check will insert a new `set_value` event on the second `linear_ramp_to_value_at_time` call, while it should not because `last_event` is defined. Therefore, the `start_time` of the second `linear_ramp` event will be wrong.

The weird funny fact is that the current behaviour works without clicks, while the specified behaviour _should_ click without a `set_value_at_time` before the second `linear_ramp`... :)

Relevant spec links:
- https://webaudio.github.io/web-audio-api/#dom-audioparam-linearramptovalueattime
- https://webaudio.github.io/web-audio-api/#dom-audioparam-exponentialramptovalueattime

b-ma:
Nothing more than what the title says!

### DIFF
--- a/src/param.rs
+++ b/src/param.rs
@@ -917,14 +917,31 @@ impl AudioParamProcessor {
             // with intrisic value and calling time.
             // cf. https://www.w3.org/TR/webaudio/#dom-audioparam-linearramptovalueattime
             // cf. https://www.w3.org/TR/webaudio/#dom-audioparam-exponentialramptovalueattime
-            //
+            if self.event_timeline.is_empty()
+                && self.last_event.is_none()
+                && (event.event_type == AudioParamEventType::LinearRampToValueAtTime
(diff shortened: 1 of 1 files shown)


orottier:
Good one

Issue #79, opened 2021-12-30
Support zero delay in DelayNode


orottier:
The current implementation does not allow for zero delay. The minimum delay is one render quantum because we have split up the DelayNode up front in a DelayWriter and a DelayReader

As discussed in #70 :

The idea would be to create a dummy connection between DelayReader and DelayWriter in the DelayNode::new constructor, this connection would do almost nothing (as basically the writer do not use output and the reader do not use input) expect guarantee the order of the processing if the delay is not in a loop. In the graph process, if the node is found in a cycle, the graph could just delete this connection and somehow flag the Reader as "in_cycle" so that the later would know it must clamp its minimum delay to the quantum duration. (I guess that's not that far from what is described in the spec, and I don't even think this would need to be reversible, e.g. once in a cycle the node behaves like that forever even if the cycle is broken later, which is really a weird edge-case I can't even imagine where users should anyway know what they are doing)

orottier:
Fixed with #198

PR #198, opened 2022-08-05
Enable sub quantum delay

b-ma:
Hey, here is a first draft on #79

Don't merge this yet, I'm really not sure it's finished but It would be nice to have your feedback on what I have done so far. As this is related to the graph, I'm a bit nervous because I'm not sure I fully understand the `visit/order` stuff :)

- I had to remove a test which is failing now but that should be reviewed because it was checking against clamped delay values. I'm not sure why it breaks, need to dig more on that.
- Also I'd like to add more tests to be sure things behaves as expected

Except from the one I mentioned the all other tests seems to be happy

### DIFF
--- a/src/node/constant_source.rs
+++ b/src/node/constant_source.rs
@@ -162,11 +162,6 @@ impl AudioProcessor for ConstantSourceRenderer {
             return true;
         }
 
-        if stop_time < scope.current_time {
-            output.make_silent();
-            return false;
-        }
-
         output.force_mono();
 
         let offset_values = params.get(&self.offset);
@@ -187,7 +182,7 @@ impl AudioProcessor for ConstantSourceRenderer {
         }
 
         // tail_time false when output has ended this quantum
-        stop_time > next_block_time
+        stop_time >= next_block_time
     }
 }
 
--- a/src/node/constant_source.rs
+++ b/src/node/constant_source.rs
@@ -186,7 +186,8 @@ impl AudioProcessor for ConstantSourceRenderer {
             current_time += dt;
         }
 
-        true
+        // tail_time false when output has ended this quantum
+        stop_time > next_block_time
     }
 }
 
--- a/src/node/delay.rs
+++ b/src/node/delay.rs
@@ -8,6 +8,7 @@ use super::{AudioNode, ChannelConfig, ChannelConfigOptions, ChannelInterpretatio
 use std::cell::{Cell, RefCell, RefMut};
 use std::rc::Rc;
 use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
 
 /// Options for constructing a [`DelayNode`]
 // dictionary DelayOptions : AudioNodeOptions {
@@ -215,7 +216,7 @@ impl DelayNode {
         let last_written_index_clone = last_written_index.clone();
 
         // shared bool that defined if the node has been found in a cycle
-        let in_cycle = Rc::new(AtomicBool::new(false));
+        let in_cycle = Arc::new(AtomicBool::new(false));
         let in_cycle_clone = in_cycle.clone();
 
         let node = context.register(move |writer_registration| {
@@ -253,7 +254,6 @@ impl DelayNode {
                 ring_buffer: shared_ring_buffer,
                 index: 0,
                 last_written_index,
-                in_cycle,
             };
 
             (node, Box::new(writer_render))
@@ -264,6 +264,7 @@ impl DelayNode {
         // connect Writer to Reader to guarantee order of processing and enable
         // sub-quantum delay. If found in cycle this connection will be deleted
         // by the graph and the minimum delay clamped to one render quantum
+        context.base().mark_cycle_breaker(writer_id, in_cycle);
         context.base().connect(writer_id, reader_id, 0, 0);
 
         node
@@ -279,7 +280,6 @@ struct DelayWriter {
     ring_buffer: Rc<RefCell<Vec<AudioRenderQuantum>>>,
     index: usize,
     last_written_index: Rc<Cell<Option<usize>>>,
-    in_cycle: Rc<AtomicBool>,
 }
 
 struct DelayReader {
@@ -289,7 +289,7 @@ struct DelayReader {
     last_written_index: Rc<Cell<Option<usize>>>,
     // local copy of shared `last_written_index` so as to avoid render ordering issues
     last_written_index_checked: Option<usize>,
-    in_cycle: Rc<AtomicBool>,
+    in_cycle: Arc<AtomicBool>,
 }
 
 // SAFETY:
@@ -340,11 +340,6 @@ impl RingBufferChecker for DelayWriter {
 }
 
 impl AudioProcessor for DelayWriter {
-    fn can_break_cycle(&self) -> bool {
-        self.in_cycle.store(true, Ordering::SeqCst);
-        true
-    }
-
     fn process(
         &mut self,
         inputs: &[AudioRenderQuantum],
--- a/src/node/delay.rs
+++ b/src/node/delay.rs
@@ -7,6 +7,7 @@ use super::{AudioNode, ChannelConfig, ChannelConfigOptions, ChannelInterpretatio
 
 use std::cell::{Cell, RefCell, RefMut};
 use std::rc::Rc;
+use std::sync::atomic::{AtomicBool, Ordering};
 
 /// Options for constructing a [`DelayNode`]
 // dictionary DelayOptions : AudioNodeOptions {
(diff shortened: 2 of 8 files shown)


src/render/graph.rs
```diff
@@ -180,13 +183,32 @@ impl Graph {
...
+                let node = self.nodes.get(node_id).unwrap();
+                node.processor.can_break_cycle()
+            });
+
+            match cycle_breaker_node {
+                Some(&node_id) => {
+                    // store node id to clear the node outgoing edges
+                    cycle_breakers.push(node_id);
+                    // remove nodes from mark temp after pos
+                    marked_temp.truncate(pos);
```
b-ma, line 201: I'm not completely sure of this

orottier: Yeah, it's quite hard to reason about it. My gut feeling is that this implementation is to optimistic. It would be safer to clear out `marked_temp` and also remove these nodes from `marked`. Then start visiting again from the first item that was in `marked_temp`

orottier (commented):
Great to see we are nearing a solution for this issue.
Very happy to see the extensive tests for it. It will help us iterate on the solution while checking all works well.
I dropped two notes, but I do not have a solution ready at hand. Would you like me to take the work from here and spend some time on it, or will you have another go?

src/render/processor.rs
```diff
@@ -26,6 +26,10 @@ pub struct RenderScope {
 ///
 /// Check the `examples/worklet.rs` file for example usage of this trait.
 pub trait AudioProcessor: Send {
+    fn can_break_cycle(&self) -> bool {
```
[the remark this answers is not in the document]
orottier: We cannot really use downcasting for this problem, e.g. https://stackoverflow.com/a/33687996 without adding an `as_any` method to all processors, which is undesirable too.

Another solution would be to explicitly store `is_cycle_breaker` in the `Node`, and setting that value with a call  
`ConcreteBaseAudioContext::markCycleBreaker(&self, node_id: &AudioNodeId)` that passes the info to the render thread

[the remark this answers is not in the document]
b-ma: Yup I agree that's not very clean, especially as it only concerns one (weird) node, that was just the only I found :)

src/render/graph.rs
b-ma: Maybe a nice thing to have would be a unit test with really weird edge case: some cycle inside a cycle I guess, to check this behaves as expected (whatever the solution)?

b-ma:
> Would you like me to take the work from here and spend some time on it, or will you have another go?

Actually if you have ideas to iterate, please go on yes! I can try to contribute the test I spoke about but then I'm short with ideas about how to continue

orottier:
Okay I will have a try at all three things (`can_break_cycle`, tests, and correct sorting algorithm). I have some ideas

b-ma:
> Okay I will have a try at all three things (can_break_cycle, tests, and correct sorting algorithm). I have some ideas

ok cool!

b-ma:
[7ef0597](https://github.com/orottier/web-audio-api-rs/pull/198/commits/7ef05970c8ad9c2181c365b07b6216be9689ba26) Nice approach, way cleaner!

b-ma:
The graph test suite looks very nice!

orottier:
Okay this is conceptually finished but I'm still looking for a more elegant graph ordering after the cycles are broken

b-ma:
I think if we have a test like:

```
            +---------+    +---------+   
            v         |    v         |
+---+     +---+     +----------+   +---+
| 1 | --> | 2 | --> | 3: delay |   | 4 |
+---+     +---+     +----------+   +---+
            |            |           |
            v            +------------
          +---+
          | 5 |
          +---+
```

where `4` should be muted while `3` should break the cycle and that it works consistently over several iterations. My impression is that we have nice test case to know if the optimistic solution is enough (i.e. truncate only is ok, which I actually I can't think why it would break but maybe I'm wrong) or if we need to be pessimistic (reorder the graph, considering that creating feedback delays is not something people do so often that is it a big issue neither :)

To make it even more perverse and see how the recursive stuff behaves, we could also add connections between `4` and `2`:

```
           +--------------------------+
           |                          |
           | +--------+    +--------+ |  
           v v        |    v        | |
+---+     +---+     +----------+   +---+
| 1 | --> | 2 | --> | 3: delay |   | 4 |
+---+     +---+     +----------+   +---+
            |            |           |
            v            +------------
          +---+
          | 5 |
          +---+
```
or reverse

```
            +---------+     +--------+   
            v         |     v        | 
+---+     +---+     +----------+   +---+
| 1 | --> | 2 | --> | 3: delay |   | 4 |
+---+     +---+     +----------+   +---+
           | |        |             | |
           | |        +-------------+ |
           | +------------------------+           
           v            
          +---+
          | 5 |
          +---+
```

 which I guess should lead to the exact same result (`4` muted, `3` breaks cycle)?

orottier:
I'm adding your examples, stay tuned!
Unfortunately I already concluded the optimistic graph sorting is incorrect at [b1498a2](https://github.com/orottier/web-audio-api-rs/pull/198/commits/b1498a24d5da84a78e9143ee64d5595f186ed93a)
This still has the `truncate and continue` style, but the cycle integration test consistently fails
I tried to fix the optimistic method but could not get it to work yet, I think you need to unroll the marked, marked_temp and ordered items in a very specific way. Maybe I will succeed tomorror

orottier:
L2 access seems to be improved with [b85de76](https://github.com/orottier/web-audio-api-rs/pull/198/commits/b85de766535dfeb607b694e3a97c736973fc4a9d)
In that vein I opened #208 to explore further optimizations

src/node/delay.rs
```diff
@@ -432,7 +435,9 @@ impl AudioProcessor for DelayReader {
         let quantum_duration = RENDER_QUANTUM_SIZE as f64 * dt;
 
         let delay_param = params.get(&self.delay_time);
-        let in_cycle = self.in_cycle.load(Ordering::SeqCst);
+
+        let latest_frame_written = self.latest_frame_written.load(Ordering::SeqCst);
+        let in_cycle = latest_frame_written != scope.current_frame;
```
b-ma, line 440: I'm not that sure this is right when in cycle actually:
- if not in cycle the order of rendering is guaranteed, therefore we can trust this check
- but when in cycle, the order of rendering is not guaranteed (at least it wasn't, did it changed somehow?), so the check could fail sometime?

b-ma:
> This still has the truncate and continue style, but the cycle integration test consistently fails

oh ok, I didn't see the sub comment in the commit message, so pessimistic we must be indeed!

> L2 access seems to be improved with [b85de76](https://github.com/orottier/web-audio-api-rs/pull/198/commits/b85de766535dfeb607b694e3a97c736973fc4a9d)
> In that vein I opened https://github.com/orottier/web-audio-api-rs/issues/208 to explore further optimizations

Nice one! (perf is really a trap :)

src/node/delay.rs
orottier: Hmmmm right, it is quite nuanced actually and my code needs to be improved
- When not in a cycle, there is a edge from writer to reader so writer always renders first
- When in a cycle, that edge is dropped. The nature of the cycle means output of the reader feeds into the writer, hence, the reader always renders first
- However, when the user breaks the cycle manually by dropping some other connections, the order is no longer guaranteed and thus random

The last case is a bit of an edge case of course, but I could make the in_cycle property sticky (one you are in a cycle, treat as if you will always stay in that cycle) so we prevent erratic behaviour in rendering of sub-quantum delays.

Another issue I realize is when the Writer is dropped, it will stop updating the `latest_frame_written`. This causes the Reader to set `in_cycle` true, disallowing sub-quantum delays. If the delay is set to 100 samples this behaviour is not right becase the final 28 frames om the last Writer call should still render directly in the first 28 frames of the next Reader output.

Pfft, not simple stuff, I will iterate again

src/node/constant_source.rs
```diff
@@ -186,7 +186,8 @@ impl AudioProcessor for ConstantSourceRenderer {
             current_time += dt;
         }
 
-        true
+        // tail_time false when output has ended this quantum
+        stop_time > next_block_time
```
b-ma, line 190: Remembered that we had to output one channel of silence before returning false, but can't find it in the spec actually... I think the check line 165 makes no sense anymore then?

b-ma: handled in e088c95

orottier:
Thanks for the update, I think we have done enought in this PR.
I added two more test cases, one remark:

> Something like that should result in both 4 muted and 3 breaks cycle: (D)

Because 2 is also part of the cycle with 4, it is muted as well

Issue #196, opened 2022-08-04
Optimizations for single-valued AudioParam buffers


orottier:
I think we could also re-open the discussion about the param length being 1 or 128, because now that we have something such as https://github.com/orottier/web-audio-api-rs/blob/183768249204c94bc0c40722186ba6898eed4731/src/param.rs#L958 we could do interesting things even with a-rate params now, e.g. for gain

```rs
if gain.len() == 1 {
  if gain[0] = 1. {
    *output = input.clone();
    return false;
  } else if gain[0] == 0. {
    *output = input.silence();
    return false;
  }
}
```

_Originally posted by @b-ma in https://github.com/orottier/web-audio-api-rs/issues/192#issuecomment-1205538120_

b-ma:
I had a try on that (that ended up to be quite dirty): that's very interesting and at first sight it seems we can grab a lot in term of perfs in many nodes, but 1. it therefore has impact on many nodes (i.e. all the ones that have a-rate parameters) which could lead to a very blotted PR if I continue like that, and 2. we lack some dedicated tests to make sure things behave well. 

I think I've found a solution that might allow to put the main ideas out while having a very minimal impacts on the nodes, so we can just modify them one by one later, but I need to restart from a fresh and clean basis.

orottier:
Okay let me know when to take a look, a WIP PR that does not compile is fine too

PR #199, opened 2022-08-08
Use `RefCell` when iterating and rendering `Graph`

orottier:
From https://github.com/orottier/web-audio-api-rs/issues/55#issuecomment-1207347931

### DIFF
--- a/src/render/graph.rs
+++ b/src/render/graph.rs
@@ -288,57 +288,54 @@ impl Graph {
         // keep track of end-of-lifecyle nodes
         let mut nodes_dropped = false;
 
-        // split (mut) borrows
-        let ordered = &self.ordered;
+        // for borrow-checker reasons, move mutable borrow of nodes out of self
         let nodes = &mut self.nodes;
 
         // process every node, in topological sorted order
-        ordered.iter().for_each(|index| {
-            let can_free = {
-                // remove node from graph, re-insert later (for borrowck reasons)
-                let mut node = nodes.get(index).unwrap().borrow_mut();
+        self.ordered.iter().for_each(|index| {
+            // remove node from graph, re-insert later (for borrowck reasons)
+            let mut node = nodes.get(index).unwrap().borrow_mut();
 
-                // make sure all input buffers have the correct number of channels, this might not be
-                // the case if the node has no inputs connected or the channel count has just changed
-                let interpretation = node.channel_config.interpretation();
-                let count = node.channel_config.count();
+            // make sure all input buffers have the correct number of channels, this might not be
+            // the case if the node has no inputs connected or the channel count has just changed
+            let interpretation = node.channel_config.interpretation();
+            let count = node.channel_config.count();
+            node.inputs
+                .iter_mut()
+                .for_each(|i| i.mix(count, interpretation));
+
+            // let the current node process
+            let params = AudioParamValues::from(&*nodes);
+            let tail_time = node.process(params, scope);
+
+            // iterate all outgoing edges, lookup these nodes and add to their input
+            node.outgoing_edges
+                .iter()
+                // audio params are connected to the 'hidden' usize::MAX output, ignore them here
+                .filter(|edge| edge.other_index != usize::MAX)
+                .for_each(|edge| {
+                    let mut output_node = nodes.get(&edge.other_id).unwrap().borrow_mut();
+                    output_node.has_inputs_connected = true;
+                    let signal = &node.outputs[edge.self_index];
+                    let channel_config = &output_node.channel_config.clone();
+
+                    output_node.inputs[edge.other_index].add(signal, channel_config);
+                });
+
+            let can_free = node.can_free(tail_time);
+
+            // Node is not dropped.
+            if !can_free {
+                // Reset input buffers as they will be summed up in the next render quantum.
                 node.inputs
                     .iter_mut()
-                    .for_each(|i| i.mix(count, interpretation));
+                    .for_each(AudioRenderQuantum::make_silent);
 
-                // let the current node process
-                let params = AudioParamValues::from(&*nodes);
-                let tail_time = node.process(params, scope);
+                // Reset input state
+                node.has_inputs_connected = false;
+            }
 
-                // iterate all outgoing edges, lookup these nodes and add to their input
-                node.outgoing_edges
-                    .iter()
-                    // audio params are connected to the 'hidden' usize::MAX output, ignore them here
-                    .filter(|edge| edge.other_index != usize::MAX)
-                    .for_each(|edge| {
-                        let mut output_node = nodes.get(&edge.other_id).unwrap().borrow_mut();
-                        output_node.has_inputs_connected = true;
-                        let signal = &node.outputs[edge.self_index];
-                        let channel_config = &output_node.channel_config.clone();
-
-                        output_node.inputs[edge.other_index].add(signal, channel_config);
-                    });
-
-                let can_free = node.can_free(tail_time);
-
-                // Node is not dropped.
-                if !can_free {
-                    // Reset input buffers as they will be summed up in the next render quantum.
-                    node.inputs
-                        .iter_mut()
-                        .for_each(AudioRenderQuantum::make_silent);
-
-                    // Reset input state
-                    node.has_inputs_connected = false;
-                }
-
-                can_free
-            };
+            drop(node); // release borrow of self.nodes
 
             // Check if we can decommission this node (end of life)
             if can_free {
--- a/src/render/graph.rs
+++ b/src/render/graph.rs
@@ -6,6 +6,7 @@ use crate::render::RenderScope;
 
 use rustc_hash::FxHashMap;
 use smallvec::{smallvec, SmallVec};
+use std::cell::RefCell;
 
 /// Connection between two audio nodes
 struct OutgoingEdge {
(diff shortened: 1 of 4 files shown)


src/render/graph.rs
```diff
@@ -352,7 +370,7 @@ impl Graph {
         }
 
         // Return the output buffer of destination node
-        &self.nodes.get(&NodeIndex(0)).unwrap().outputs[0]
+        self.nodes.get(&NodeIndex(0)).unwrap().borrow().outputs[0].clone()
```
orottier, line 373: I removed your `unsafe` use here just by cloning out the buffer.
In case of the online AudioContext rendering, we need an owned version anyway.
For OfflineAudioContext I think the benefit of no-unsafe is worth the `Rc` increment from `clone`

orottier:
> The downside is that I didn't manage to get rid of unsafe code in 2 places. It very localized and seems to be the same problem each time (i.e. returning a reference to the buffer in Graph::render() and AudioParamValues::get()) so maybe you would have an idea to handle that?

Yes this is the deal with `RefCell`. You will not be able to hide the fact that you are using it, so they will always appear in your function signatures, unless you resort to unsafe code

> It very localized
This is a fallacy of `unsafe`. Even though the unsafe code block is tiny, the implication can reach very far, in seemingly unrelated, safe code as well. You will never know what the compiler assumes and optimizes

I will try to document a `SAFETY` block for the unsafe code in `AudioParamValues`. I think we will be able to get away with it. Returning a `&[f32]` from the `get` function makes a lot of sense. And we expose that to library users, so I rather not return a `Ref` object.

All in all this is a very elegant, minimal solution to our problem. Kudo's!

src/render/processor.rs
```diff
@@ -58,22 +59,24 @@ pub trait AudioProcessor: Send {
...
 
 impl<'a> AudioParamValues<'a> {
-    pub(crate) fn from(nodes: &'a FxHashMap<NodeIndex, Node>) -> Self {
+    pub(crate) fn from(nodes: &'a FxHashMap<NodeIndex, RefCell<Node>>) -> Self {
         Self { nodes }
     }
 
     pub(crate) fn get_raw(&self, index: &AudioParamId) -> &AudioRenderQuantum {
-        self.nodes.get(&index.into()).unwrap().get_buffer()
+        unsafe { (*self.nodes.get(&index.into()).unwrap().as_ptr()).get_buffer() }
```
orottier, line 71: Sorry, I tried, but this `unsafe` snippet is fundamentally unsound.
If I modify the function signatures to make the returned references of static lifetime, it still compiles:
```rust
-    pub(crate) fn get_raw(&self, index: &AudioParamId) -> &AudioRenderQuantum {
+    pub(crate) fn get_raw(&self, index: &AudioParamId) -> &'static AudioRenderQuantum {
         unsafe { (*self.nodes.get(&index.into()).unwrap().as_ptr()).get_buffer() }
     }

     /// Get the computed values for the given [`crate::param::AudioParam`]
     ///
     /// For both A & K-rate params, it will provide a slice of length [`crate::RENDER_QUANTUM_SIZE`]
-    pub fn get(&self, index: &AudioParamId) -> &[f32] {
+    pub fn get(&self, index: &AudioParamId) -> &'static [f32] {
         // let buffer = self.nodes.get(&index.into()).unwrap().borrow().get_buffer();
         // &buffer.channel_data(0)[..]
         &self.get_raw(index).channel_data(0)[..]
```
Which is obvious. By moving to a ptr and then back, you are erasing the lifetime information.
This is problematic. A user provided `AudioProcessor` might take the `&'static [f32]` from the `AudioParamValues.get` method, and store it for a longer time than the buffer exists (the graph may deallocate it already in the next `render` call)

Let me try to make it sound by dropping unsafe, but the function signature will suffer. Let's see if it is acceptable

b-ma: Hum, I'm not sure I understand well, these lifetime stuff are still a bit like dark magic for me... (slowly understanding things but, slowly :)

I just wonder naively, why can't we use the same trick you did in `Graph::render()` and just rely on a clone somehow?

orottier:
Benchmarks with `unsafe` removed:
```
- 1       | Baseline (silence)                                                       | 11            | 10909.1x              | 120
- 2       | Simple source test without resampling (Mono)                             | 16            | 7500.0x               | 120
- 3       | Simple source test without resampling (Stereo)                           | 22            | 5454.5x               | 120
- 4       | Simple source test without resampling (Stereo and positionnal)           | 97            | 1237.1x               | 120
- 5       | Simple source test with resampling (Mono)                                | 40            | 3000.0x               | 120
- 6       | Simple source test with resampling (Stereo)                              | 53            | 2264.2x               | 120
- 7       | Simple source test with resampling (Stereo and positionnal)              | 128           | 937.5x                | 120
- 8       | Upmix without resampling (Mono -> Stereo)                                | 21            | 5714.3x               | 120
- 9       | Downmix without resampling (Stereo -> Mono)                              | 15            | 8000.0x               | 120
- 10      | Simple mixing (100x same buffer) - be careful w/ volume here!            | 933           | 32.2x                 | 30
- 11      | Simple mixing (100 different buffers) - be careful w/ volume here!       | 977           | 30.7x                 | 30
- 12      | Simple mixing with gains                                                 | 203           | 591.1x                | 120
- 13      | Granular synthesis                                                       | 2132          | 3.5x                  | 7.5
- 14      | Synth (Sawtooth with Envelope)                                           | 1788          | 67.1x                 | 120
- 15      | Synth (Sawtooth with gain - no automation)                               | 1452          | 82.6x                 | 120
- 16      | Synth (Sawtooth without gain)                                            | 886           | 135.4x                | 120
- 17      | Substractive Synth                                                       | 280           | 428.6x                | 120
- 18      | Stereo panning                                                           | 40            | 3000.0x               | 120
- 19      | Stereo panning with automation                                           | 42            | 2857.1x               | 120
- 20      | Sawtooth with automation                                                 | 47            | 2553.2x               | 120
- 21      | Stereo source with delay                                                 | 75            | 1600.0x               | 120
```

b-ma:
The safe version seems to be a bit less performant (approx. in between before and the unsafe version for some case, other ones seems untouched, that's weird...) on my machine, but if it's the price for safety maybe it worth it, you are the only one who really understands the consequences here :) In any case this is already quite nice improvements I think.

I guess #188 will start to be more and more important in the todo list, I have the impression that the more we gain in terms of perfs, the more the noise we have in the current benches appears to be important

src/render/processor.rs
```diff
@@ -67,16 +78,20 @@ impl<'a> AudioParamValues<'a> {
...
 
-    pub(crate) fn get_raw(&self, index: &AudioParamId) -> &AudioRenderQuantum {
-        unsafe { (*self.nodes.get(&index.into()).unwrap().as_ptr()).get_buffer() }
+    pub(crate) fn get_raw(&self, index: &AudioParamId) -> AudioRenderQuantum {
+        self.nodes
+            .get(&index.into())
+            .unwrap()
+            .borrow()
+            .get_buffer()
+            .clone()
```
b-ma, line 87: Not related to this issue but maybe we will have to review the API here, I started to sketch something for #196 and this is actually quite a blocking point

src/render/processor.rs
orottier: > these lifetime stuff are still a bit like dark magic for me

which is fine, unless you use `unsafe` to mess with lifetimes ;)

> I just wonder naively, why can't we use the same trick you did in Graph::render() and just rely on a clone somehow?

Definitely, but the function signatures will change.
One simple way would be to have the `get` method return an AudioRenderQuantumChannel
```
impl<'a> AudioParamValues<'a>

    pub fn get(&self, index: &AudioParamId) -> impl std::ops::Deref<Target = [f32; 128]> + '_ {
        self.nodes
            .get(&index.into())
            .unwrap()
            .borrow()
            .get_buffer()
            .channel_data(0)
            .clone()
    }
}
```

But now, it will deref always to a `[f32; 128]` fixed length array. In the current impl we deref to `[f32]`, a slice with no fixed length. This leaves room for your work in #196

src/render/processor.rs
orottier: As mentioned above, we can make this work given the current method signatures.
Maybe for now ignore the `get_raw` calls for the AudioListener, we will fix that later.
To make `pub fn get(&self, index: &AudioParamId)` work, maybe add a `bool` to `AudioRenderQuantumChannel`: `is_single_valued` to denote that only the first f32 should be used. Then you should be able to get the idea right, and later we will make it nicer/more performant. 
I hope this makes sense to you. Feel free to push some work in progress any time and we will look together

orottier:
> The safe version seems to be a bit less performant
Yes, a little it seems. My machine is still too noisy to measure actually. I will start with the proper benchmark suite then.
For now, I will probably merge this latest version - maybe touch up the code a bit to make it more readable.

b-ma:
> For now, I will probably merge this latest version

nice!

src/render/graph.rs
```diff
@@ -308,34 +314,43 @@ impl Graph {
...
                 node.has_inputs_connected = false;
+            }
 
-                // Re-insert node in graph
-                nodes.insert(*index, node);
+            drop(node); // release borrow of self.nodes
+
+            // Check if we can decommission this node (end of life)
+            if can_free {
+                // Node is dropped, we should perform a new topological sort of the audio graph
```
b-ma, line 342: Just one last thing maybe, this comment is confusing because we actually don't trigger a new sort of the graph here

src/render/processor.rs
b-ma: > add a bool to AudioRenderQuantumChannel: is_single_valued to denote that only the first f32 should be used

ok that was one solution I had in mind, I will continue this way then

PR #200, opened 2022-08-09
Enable caching of cargo files in CI

orottier:


orottier:
Before: https://github.com/orottier/web-audio-api-rs/actions/runs/2827155380/attempts/1  3m38s
After: https://github.com/orottier/web-audio-api-rs/actions/runs/2827155380/attempts/2 1m34s

b-ma:
Nice!

PR #202, opened 2022-08-10
Implemented tail time and actively processing on DynamicCompressorNode

b-ma:
Related to #34 and #147

I'm not 100% sure of that (seems almost to easy...), but I can't find neither why it would not do the job, so...


orottier:
Very simple indeed. I see no issue actually

Issue #194, opened 2022-08-04
How to deal with subnormal floats?


orottier:
> FYI this branch breaks SIMD-optimizations.

Damned !

> I'm wondering if this check (and the one we just added to the AudioParam) is worth it. We are slowing down all architectures to save some performance on exotic ones?

I'm afraid they are not really exotic platforms... Intel x86_64 just like mine... ARMv8 feels more exotic to me :)

I will remove that from the GainNode for now, maybe it's best to open an issue to make sure we don't forget the possible problem and wait until we both better understand it and (hopefully) have a better solution to propose

I wouldn't compare this with the check in the AudioParam that just happen once in a while in very specific conditions though

https://stackoverflow.com/questions/9314534/why-does-changing-0-1f-to-0-slow-down-performance-by-10x/9314926#9314926, there is also a comment about audio application there

_Originally posted by @b-ma in https://github.com/orottier/web-audio-api-rs/pull/191#discussion_r937352505_

orottier:
I'm hoping we can do something on the compiler level. A hint is provided here: https://gist.github.com/GabrielMajeri/545042ee4f956d5b2141105eb6a505a9
But that is limited to some SIMD operations

b-ma:
Yep, I saw this one too. But reading other threads it seems it may not be a really good idea to do that within a library as the whole final application would inherit the behaviour which may not be desirable...

From what I understand (but I'm really not sure there, this has to be confirmed) with fast math operations, this would be avoided too but these are still nightly, e.g. https://doc.rust-lang.org/std/intrinsics/fn.fmul_fast.html.

orottier:
https://github.com/Sin-tel/no_denormals looks very useful

PR #203, opened 2022-08-10
Refactor IIRFilterNode

b-ma:
Some work on the IIRFilter, mainly:

- simplified the code
- removed the remaining test against the snapshot
- added tests against the `BiquadFilter`,  both for output and get_frequency_response() - there is just one small discrepancy for get_frequency_response and the phases of the notch filter, I didn't really dig why but doesn't seem like a huge issue neither
- handle tail time
 
I guess, if we don't consider the ConvolverNode, that we get near to have #34 and #147 finished now!


b-ma:
Huhu, yup that's my "before holiday's sprint" :) It will be more quiet after this week

> I will have a good look soon

No problem take your time, I know I'm just flooding these days

src/node/iir_filter.rs
```diff
@@ -116,96 +198,66 @@ impl IIRFilterNode {
...
-            " InvalidAccessError: All paramaters should be the same length"
-        );
-
-        // Ensures that given frequencies are in the correct range
-        let min = 0.;
-        let max = self.context().sample_rate() / 2.;
-        for f in frequency_hz.iter_mut() {
-            *f = f.clamp(min, max);
+            let (mag, phase) = response.to_polar();
+            mag_response[i] = mag as f32;
```
orottier, line 239: For performance (not really important here) you could zip them along in the loop, this prevents bounds checks on every loop item:
```
for (i, &f, m, p) in frequency_hz.iter().enumerate().zip(mag_response.iter_mut()).zip(phase_response.iter_mut())
```

src/node/iir_filter.rs
```diff
@@ -278,268 +306,398 @@ impl AudioProcessor for IirFilterRenderer {
...
+                return false;
+            }
+        }
 
-        true // todo tail time - issue #34
-    }
-}
+        // eventually resize state according to input number of channels
+        // if in tail time we should continue with previous number of channels
+        if !input.is_silent() {
```
orottier, line 330: Nitpick, maybe swap the if/else as
```
if input.is_silent {
 // silent case
} else {
 // non-silent
}
```

Parsing the else was a bit hard to me, double negation

orottier (approved):
Great stuff, nice tests.
As you can see I'm trying to get a benchmark versus main running, will follow soon.
I left some very small nitpicks

orottier:
Unfortunately, our own benchmark suite shows the same regression in performance as the github-bot posted.
I have not dug into it, but two things come to mind:
- the `all_channels_equal` trick
- `let states = vec![Vec::<f64>::with_capacity(MAX_CHANNELS); coeffs_len];` is probably better for cache performance when we transpose it `Vec<Vec<coefs>, channels>`, and even better, make it a single Vec, not Vec of Vecs

b-ma:
Hey, before you spend too much time on it, I think there is no regression (I really didn't change much the actual computing code) except there is now `output.set_number_of_channels(num_channels);` while it was missing before so it was only mono. If you force `set_channel_number` to 1 in the new version, the footprint should be mostly the same I think

I will try to have a look at your reviews tomorrow

b-ma:
Hum actually, I just made a rapid test and the channel count does not explain everything:

### main

```
IIR filter                                                               | 79            | 1519.0x               | 120
```

### refactor

> with default channel_count (stereo)

```
IIR filter                                                               | 160           | 750.0x                | 120
```

> with set_channel_count(1)

```
IIR filter                                                               | 106           | 1132.1x               | 120
```

So, I added or changed something to is not good indeed...

> the all_channels_equal trick

hum, we didn't merge that yet, I made some tests but it wasn't very satisfying (cf. https://github.com/orottier/web-audio-api-rs/issues/117#issuecomment-1206302154)

> let states = vec![Vec::<f64>::with_capacity(MAX_CHANNELS); coeffs_len]; is probably better for cache performance when we transpose it Vec<Vec<coefs>, channels>, and even better, make it a single Vec, not Vec of Vecs

yup probably but that's something I didn't really change neither

I will try to dig into that, sounds weird

b-ma:
Actually there seem to be two things that seems to explain (the test are noisy that's a bit difficult to be affirmative):

- the check for subnormal seems to have a cost, but I'm not sure it's a good idea to remove it (or we remove it for now and put that as a task for #194)

- before, there was:
```
let output = self.norm_coeffs[0].0.mul_add(input, self.states[0][channel_number]);
``` 
that I changed (just for readability) to:
```
let b0 = self.norm_coeffs[0].0;
let last_state = self.states[0][channel_number];

let output = b0.mul_add(input, last_state);
```
I'm a bit surprised here, but maybe the conjunction of these two things allowed for some sort of compiler optimizations that we loose, I don't know

On the states layout, before it was `states: Vec<[f64; MAX_CHANNELS]>` that I changed to `states: Vec<Vec<f64>>` which may also be a bad idea. But as you said, we can probably do better than each of these ones.

orottier:
Thanks for digging in. I am inclined to remove the subnormal check and leave that for later in the dedicated issue.
I will make some changes and merge if performance is going in the right direction

b-ma:
> Thanks for digging in. I am inclined to remove the subnormal check and leave that for later in the dedicated issue.
> I will make some changes and merge if performance is going in the right direction

ok sound goods, for the benches keep in mind that we comparison is biased with main as in main the processing is only mono while it is stereo here (you need to set `iir.set_channel_count(1)` to compare things more properly)

orottier:
Yup, when we force mono processing for me there is no performance difference anymore with main..
Going to merge this then, but will roll back the `help compiler optimize inner loop` commit. It makes no difference locally and at CI

PR #204, opened 2022-08-10
Use SmallVec for Node inputs and outputs

orottier:
This is just a silly PR to check out the new benchmark feature.

### DIFF
--- a/src/render/graph.rs
+++ b/src/render/graph.rs
@@ -23,9 +23,9 @@ pub struct Node {
     /// Renderer: converts inputs to outputs
     processor: Box<dyn AudioProcessor>,
     /// Reusable input buffers
-    inputs: Vec<AudioRenderQuantum>,
+    inputs: SmallVec<[AudioRenderQuantum; 2]>,
     /// Reusable output buffers, consumed by subsequent Nodes in this graph
-    outputs: Vec<AudioRenderQuantum>,
+    outputs: SmallVec<[AudioRenderQuantum; 2]>,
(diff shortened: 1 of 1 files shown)


orottier:
Well, that's quite surprising actually..
In any case, not merging

FYI @b-ma this now runs for all pull requests - but I still need to put all our benchmarks in place. 
Let's see if this is of benefit and corresponds to what we see locally using the `benchmark` example file

orottier:
Nice, results seem pretty stable when running on different VMs.

b-ma:
Nice stuff! And having a github bot posting messages is really cool :) 
These L1 et L2 things could turned to be very interesting in long term (reminding me Paul's keynote)

Some questions:
- do the `+5.270977%` is the difference compared to `main`? 
- is it the gain in performance or the increased number of accesses (which would then be a loss in performances I guess)? In short should we prefer "+" or "-" here :) I didn't find it in the doc

Also, as it seems it could be hard to it locally (I tried to install valgrind but it does not work...), do you think it would be possible later to also have an alternative version using `criterion.rs` (hopefully without much code overhead) that we could run locally with a flag or something? (they seems to say that it could be a good idea https://github.com/bheisler/iai#comparison-with-criterion-rs)

orottier:
Indeed as you noticed, it compares to `main`.
Lower is better, but sometimes you trade numbers so decreasing L2 access while increasing L1 might be beneficial.
It would be nice to use criterion as well, that is definitely possible with some work and clever usage of compiler flags. I will update #188

PR #205, opened 2022-08-11
Varying param size

b-ma:
Hey, here is a first pass concerning #196, 

I did the minimum on the nodes to make things work. I only tested simple/obvious modifications on the Gain and the StereoPanner to have a taste of which kind of perf improvement we can expect when things are properly handled.

#### before

```
- 1       | Baseline (silence)                                                       | 23            | 5217.4x               | 120
- 2       | Simple source test without resampling (Mono)                             | 34            | 3529.4x               | 120
- 3       | Simple source test without resampling (Stereo)                           | 54            | 2222.2x               | 120
- 4       | Simple source test without resampling (Stereo and positionnal)           | 168           | 714.3x                | 120
- 5       | Simple source test with resampling (Mono)                                | 81            | 1481.5x               | 120
- 6       | Simple source test with resampling (Stereo)                              | 111           | 1081.1x               | 120
- 7       | Simple source test with resampling (Stereo and positionnal)              | 218           | 550.5x                | 120
- 8       | Upmix without resampling (Mono -> Stereo)                                | 36            | 3333.3x               | 120
- 9       | Downmix without resampling (Stereo -> Mono)                              | 38            | 3157.9x               | 120
- 10      | Simple mixing (100x same buffer) - be careful w/ volume here!            | 1741          | 17.2x                 | 30
- 11      | Simple mixing (100 different buffers) - be careful w/ volume here!       | 1725          | 17.4x                 | 30
- 12      | Simple mixing with gains                                                 | 324           | 370.4x                | 120
- 13      | Granular synthesis                                                       | 2443          | 3.1x                  | 7.5
- 14      | Synth (Sawtooth with Envelope)                                           | 2905          | 41.3x                 | 120
- 15      | Synth (Sawtooth with gain - no automation)                               | 2277          | 52.7x                 | 120
- 16      | Synth (Sawtooth without gain)                                            | 1382          | 86.8x                 | 120
- 17      | Substractive Synth                                                       | 419           | 286.4x                | 120
- 18      | Stereo panning                                                           | 82            | 1463.4x               | 120
- 19      | Stereo panning with automation                                           | 81            | 1481.5x               | 120
- 20      | Sawtooth with automation                                                 | 78            | 1538.5x               | 120
- 21      | Stereo source with delay                                                 | 204           | 588.2x                | 120
```

#### after

```
- 1       | Baseline (silence)                                                       | 27            | 4444.4x               | 120
- 2       | Simple source test without resampling (Mono)                             | 32            | 3750.0x               | 120
- 3       | Simple source test without resampling (Stereo)                           | 51            | 2352.9x               | 120
- 4       | Simple source test without resampling (Stereo and positionnal)           | 129           | 930.2x                | 120
- 5       | Simple source test with resampling (Mono)                                | 76            | 1578.9x               | 120
- 6       | Simple source test with resampling (Stereo)                              | 104           | 1153.8x               | 120
- 7       | Simple source test with resampling (Stereo and positionnal)              | 184           | 652.2x                | 120
- 8       | Upmix without resampling (Mono -> Stereo)                                | 36            | 3333.3x               | 120
- 9       | Downmix without resampling (Stereo -> Mono)                              | 34            | 3529.4x               | 120
- 10      | Simple mixing (100x same buffer) - be careful w/ volume here!            | 1606          | 18.7x                 | 30
- 11      | Simple mixing (100 different buffers) - be careful w/ volume here!       | 1607          | 18.7x                 | 30
- 12      | Simple mixing with gains                                                 | 347           | 345.8x                | 120
- 13      | Granular synthesis                                                       | 2104          | 3.6x                  | 7.5
- 14      | Synth (Sawtooth with Envelope)                                           | 2681          | 44.8x                 | 120
- 15      | Synth (Sawtooth with gain - no automation)                               | 1747          | 68.7x                 | 120
- 16      | Synth (Sawtooth without gain)                                            | 1035          | 115.9x                | 120
- 17      | Substractive Synth                                                       | 453           | 264.9x                | 120
- 18      | Stereo panning                                                           | 56            | 2142.9x               | 120
- 19      | Stereo panning with automation                                           | 66            | 1818.2x               | 120
- 20      | Sawtooth with automation                                                 | 81            | 1481.5x               | 120
- 21      | Stereo source with delay                                                 | 215           | 558.1x                | 120
````

The perf increase for stereo panner is quite interesting, for the gain this is a bit less clear. ~In particular I'm a bit surprised by "Simple mixing with gains" but maybe I did something wrong... I didn't really dig into it.~_(edit: that's actually logical, the gains are not 1. as I remembered)_

A small list recapitulating what I did on the nodes:

- Analyser - no param
- AudioBufferSource - only constrained k-rate parameters
- BiquadFilter - 4 a-rate params: simple naive fix
- ChannelMerger - no param
- ChannelSplitter - no param
- ConstantSource - 1 a-rate param: simple naive fix
- Delay - 1 a-rate param: simple naive fix
- Destination - no param
- Dynamics compressor: only constrained k-rate params
- Gain - 1 a-rate param: simple naive fix + obvious optimizations
- IIRFilter - no param
- MediaElementAudioSource - no param
- MediaStreamAudioDestination - no param
- MediaStreamAudioSource - no param
- Oscillator - 2 a-rate params: simple naive fix
- Panner - a-rate param but treated as k-rate for now, nothing to do
- StereoPanner - 1 a-rate param: simple naive fix + obvious optimizations
- WaveShaper - no params

Let me know what you think!

### CODE REVIEW

**Comment by b-ma** [comment]
Damned the bot failed to post the message! Here are the results:

```
bench_ctor
  Instructions:              869259 (-0.467742%)
  L1 Accesses:              1760072 (-0.271127%)
  L2 Accesses:                 7820 (-0.445576%)
  RAM Accesses:               10316 (-0.096843%)
  Estimated Cycles:         2160232 (-0.245204%)

bench_sine
  Instructions:             9006985 (-4.433743%)
  L1 Accesses:             13476035 (-5.795511%)
  L2 Accesses:                29369 (-6.420469%)
  RAM Accesses:               12468 (+0.032092%)
  Estimated Cycles:        059260 (-5.631454%)

bench_sine_gain
  Instructions:             9517086 (-7.979000%)
  L1 Accesses:             14242739 (-9.003834%)
  L2 Accesses:                31651 (-23.60736%)
  RAM Accesses:               12614 (-0.0853%)
  Estimated Cycles:        14842484 (-8.945955%)

bench_sine_gain_delay
  Instructions:              233726 (-1.440638%)
  L1 Accesses:               337985 (-0.934070%)
  L2 Accesses:                80384 (-36.06058%)
  RAM Accesses:               13852 (-0.0079%)
  Estimated Cycles:          346787 (-1.548167%)
```

Seems to answer my questions of this morning: yes it's compared against main and yes "-" something seems to be what we want

File: src/node/stereo_panner.rs [hunk 1]
```diff
@@ -239,39 +239,73 @@ impl AudioProcessor for StereoPannerRenderer {
...
+                                *r = input_right * gain_right;
+                            } else {
+                                let x = pan;
+                                let [gain_left, gain_right] = get_stereo_gains(x);
+
+                                *l = input_left * gain_left;
+                                *r = input_left.mul_add(gain_right, input_right);
+                            }
+                        });
+                }
```
b-ma, line 308: a bit too much duplicated code here, that was just a rapid test, could be refactored

src/render/processor.rs
```diff
@@ -62,7 +62,13 @@ impl Deref for DerefAudioRenderQuantumChannel<'_> {
     type Target = [f32];
 
     fn deref(&self) -> &Self::Target {
-        self.0.get_buffer().channel_data(0).as_slice()
+        let buffer = self.0.get_buffer();
```
b-ma, line 65: not sure this is the correct way

src/node/biquad_filter.rs
```diff
@@ -609,12 +609,13 @@ impl AudioProcessor for BiquadFilterRenderer {
...
         let mut current_coefs = Coefficients::default();
 
+        // @todo - optimization only compute coefs once if all params have length == 1
         coefs_list
             .iter_mut()
-            .zip(frequency.iter())
-            .zip(detune.iter())
-            .zip(q.iter())
-            .zip(gain.iter())
+            .zip(frequency.iter().cycle().take(RENDER_QUANTUM_SIZE))
```
orottier, line 615: `take` is not necessary after `cycle` because `zip` will stop after the shortest iterator (which is `coefs_list`)
This goes for all usage of `take` in this PR!

src/node/oscillator.rs
```diff
@@ -377,49 +377,52 @@ impl AudioProcessor for OscillatorRenderer {
...
+                if current_time < start_time || current_time >= stop_time {
+                    *o = 0.;
+                    current_time += dt;
+
+                    return;
+                }
 
-                continue;
-            }
+                // @todo: we could avoid recompute that if both param lengths are 0
```
orottier, line 392: > if both param lengths are 0

you mean 1, right?

src/render/processor.rs
orottier: Looks, good. A bit nicer is maybe:
```
let len = if buffer.is_single_valued() { 1 } else { RENDER_QUANTUM_SIZE };
&self.0.get_buffer().channel_data(0)[..len]
```

orottier (commented):
Nice implementation, promising results.
I will think a bit more about where to leave this boolean `single_valued`, it could be inside the channel, inside the buffer, or in a wrapper for the param values.

src/param.rs
```diff
@@ -626,24 +625,42 @@ impl AudioParamProcessor {
     }
 
     fn mix_to_output(&mut self, input: &AudioRenderQuantum, output: &mut AudioRenderQuantum) {
-        #[cfg(test)]
-        assert_eq!(self.buffer.len(), RENDER_QUANTUM_SIZE);
```
orottier, line 630: I still like this test assertion, maybe bring it back with `len == 1 or RENDER_QUANTUM_SIZE`
It might guard us in the future for wrong boundary checks for filling the buffer

src/param.rs
```diff
@@ -626,24 +625,42 @@ impl AudioParamProcessor {
...
+            output.set_is_single_valued(true);
 
-                if o.is_nan() {
-                    *o = self.default_value;
-                }
+            let output_channel = output.channel_data_mut(0);
+            output_channel[0] = value.clamp(self.min_value, self.max_value);
+        } else {
+            // the param buffer is constant but it should be modulated by an
+            // incoming signal, so we need to resize it to RENDER_QUANTUM_SIZE
```
orottier, line 641: Maybe in this branch it is now simpler to set `output = input` and then use the loop to add in the `buffer`, using the `cycle` trick again. You don't need the resize in that case.

You could also optimize for the case that buffer is a single value of `0.` maybe, then you can skip the whole loop

src/param.rs
```diff
@@ -970,28 +988,24 @@ impl AudioParamProcessor {
...
 
-        let next_block_time = dt.mul_add(count as f64, block_time);
         let is_a_rate = self.is_a_rate.load(Ordering::SeqCst);
         let is_k_rate = !is_a_rate;
+        let is_timeline_empty = self.event_timeline.is_empty();
 
-        if is_k_rate {
-            self.buffer.resize(count, self.intrisic_value);
+        if is_k_rate || is_timeline_empty {
+            self.buffer.resize(1, self.intrisic_value);
```
orottier, line 999: `self.buffer.push(self.intrisic_value)` is a bit more intuitive maybe

src/param.rs
```diff
@@ -970,28 +988,24 @@ impl AudioParamProcessor {
...
+            // nothing to compute in timeline
+            if is_timeline_empty {
+                return;
+            }
         }
 
+        let next_block_time = dt.mul_add(count as f64, block_time);
+
         loop {
             let some_event = self.event_timeline.peek();
```
orottier, line 1011: If the next event is at `> next_block_time` and we are not currently in a `setTargetAtTime`, we can also return early with a single valued buffer, no?

orottier:
Feel free to leave some of my comments for later, I think we could merge this in the meantime and add a ticket with further improvements

src/render/processor.rs
b-ma: done!

src/node/stereo_panner.rs
b-ma: I didn't change that, it's no big deal neither and can be done later I think

src/node/biquad_filter.rs
b-ma: yup you are right, was a bit too much defensive, I removed them all

src/node/oscillator.rs
b-ma: yup :)

src/param.rs
b-ma: done

src/param.rs
b-ma: yeah right more simple indeed

> You could also optimize for the case that buffer is a single value of 0. maybe, then you can skip the whole loop

Hum not really, we still need to clamp because we can't trust the incoming signal boundaries. Another case is when `input.is_silent()` then we can `output.copy_from_slice(buffer). In all cases we will need the loop to check `nan`s and clamp, don't know if it worth it. I added a comment in the code, we can see that later I think

src/param.rs
b-ma: yup done

src/param.rs
b-ma: Hum, actually that's a bit more tricky, because for example for `LinearRamp` or `ExponentialRamp` the actual event time is the end time, their `start_time` being `last_event.time` so this check alone would break them.

I think we could maybe do better, but I need to write all the possible cases down, they don't fit in my head :)

b-ma:
> The ArrayVec change seems to have a negative impact, so strange
Yup not completely negative but a bit weird indeed, maybe we can just undo this `VecArray` thing, as you feel it's best.

I would be curious actually to test how would behave `buffer` as a simple [f32, RENDER_QUANTUM_SIZE] and just maintain the index manually. Will try that at some point to see if it improve performances or if it is transparent compared to the current implementation

PR #207, opened 2022-08-12
Fix CI post message

b-ma:
I did that based on https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

I couldn't test the post of the message itself, because the second job was not triggered for some reason on my repo (maybe it must be in main branch I don't really know...), but the artifact has been correctly created as you can see there https://github.com/b-ma/web-audio-api-rs/pull/1/checks

As the second script is very mostly a copy/paste from the github article I guess it should work properly (or a least don't be far).

### DIFF
--- a/.github/workflows/post-pr-result.yml
+++ b/.github/workflows/post-pr-result.yml
@@ -1,0 +1,59 @@
+name: Comment on the pull request
+
+# read-write repo token
+# access to secrets
+#
+# untrusted code should not be compiled or executed in this script
+# see https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
+on:
+  workflow_run:
+    workflows: ["Rust"]
+    types:
+      - completed
+
+jobs:
+  upload:
+    runs-on: ubuntu-latest
+    if: >
+      ${{ github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      - name: 'Download artifact'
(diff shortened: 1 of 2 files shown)


b-ma:
The artefact has been correctly uploaded here too https://github.com/orottier/web-audio-api-rs/actions/runs/2846981432

orottier:
Okay I skimmed the article and this looks alright, let's see!

orottier:
Now we need a way to not run this only on pull requests, not an update of main :)
It tried to comment on issue 0: https://github.com/orottier/web-audio-api-rs/actions/runs/2847620236

b-ma:
> Now we need a way to not run this only on pull requests, not an update of main :)

Damn I was thinking this https://github.com/orottier/web-audio-api-rs/blob/ab9cf1145dad18f7e2b9523bb6d6f50418ed0e01/.github/workflows/post-pr-result.yml#L18 would do the trick...

b-ma:
Seems there is a `branches-ignore` filter (cf. https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-branches), so maybe adding in `post-pr-result.yml` would do the trick:

```
on:
  workflow_run:
    branches-ignore:    
      - 'main'
    workflows: ["Rust"]
    types:
      - completed
```

orottier:
Thanks, will try tomorrow. Posting the message from your delay-cycle branch worked, great success!

PR #211, opened 2022-08-23
Perf `BiquadFilter`

b-ma:
Just a few changes to handle more cleanly the `param.len()`, I also removed the subnormal check (~50% of the perf improvements) to be consistent with the IIR Filter and added some simple dedicated benches

### before

```
Biquad filter                                                            | 141           | 851.1x                | 120
```

### after

```
Biquad filter                                                            | 95            | 1263.2x               | 120
```

### DIFF
--- a/benches/my_benchmark.rs
+++ b/benches/my_benchmark.rs
@@ -98,8 +98,8 @@ pub fn bench_buffer_src_biquad() {
     let buffer = ctx.decode_audio_data_sync(file).unwrap();
 
     // Create an biquad filter node (defaults to low pass)
-    let biquad = context.create_biquad_filter();
-    biquad.connect(&context.destination());
+    let biquad = ctx.create_biquad_filter();
+    biquad.connect(&ctx.destination());
     biquad.frequency().set_value(200.);
 
     // Play buffer and pipe to filter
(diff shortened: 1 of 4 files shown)


benches/my_benchmark.rs
```diff
@@ -92,6 +92,25 @@ pub fn bench_buffer_src_iir() {
...
+    src.connect(&biquad);
+    src.set_buffer(buffer);
+    src.start();
+
+    assert_eq!(ctx.start_rendering_sync().length(), 48000);
+}
+
 iai::main!(
     bench_ctor,
     bench_sine,
```
orottier, line 116: You should add `bench_buffer_src_biquad` also here :)

orottier: As you will notice, since the test is not on main, it will still show up in the PR comment but since it has only ran on this branch, you cannot compare performance.
I'm thinking of a fix to copy the PR branch benchmarks file to the main branch and then run them to see the desired results

orottier (commented):
Very cool!

orottier:
Sorry for hijacking your PR to experiment with the benchmarks in CI!

PR #213, opened 2022-08-24
Dedicated path for `GainNode` if param is length 1

b-ma:
Really nothing fancy, but interestingly improves the "mixing with gains" bench

#### before

```
Simple mixing with gains                                                 | 347           | 345.8x                | 120
```

#### after

```
Simple mixing with gains                                                 | 282           | 425.5x                | 120
```

### DIFF
--- a/src/node/gain.rs
+++ b/src/node/gain.rs
@@ -113,12 +113,20 @@ impl AudioProcessor for GainRenderer {
 
         *output = input.clone();
 
-        output.modify_channels(|channel| {
-            channel
-                .iter_mut()
-                .zip(gain.iter().cycle())
-                .for_each(|(o, g)| *o *= g);
-        });
+        if gain.len() == 1 {
+            let g = gain[0];
+
+            output.channels_mut().iter_mut().for_each(|channel| {
+                channel.iter_mut().for_each(|o| *o *= g);
+            });
+        } else {
+            output.channels_mut().iter_mut().for_each(|channel| {
+                channel
+                    .iter_mut()
+                    .zip(gain.iter().cycle())
+                    .for_each(|(o, g)| *o *= g);
+            });
+        }
 
         false
     }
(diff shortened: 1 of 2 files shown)


src/node/gain.rs
```diff
@@ -113,12 +113,20 @@ impl AudioProcessor for GainRenderer {
 
         *output = input.clone();
 
-        output.modify_channels(|channel| {
-            channel
-                .iter_mut()
-                .zip(gain.iter().cycle())
-                .for_each(|(o, g)| *o *= g);
-        });
+        if gain.len() == 1 {
```
orottier, line 116: Want to go the extra mile and handle the `gain[0] < 0.` and `gain[0] == 1.` case? Maybe even with an treshold (so 0.9999999 gain will leave the buffer as is)

b-ma: Actually, the cases for `gain[0] == 0` (silence) and `gain[0] == 1` (identity) are already handled just before (line 104).

I'm not sure we can assume to do something such as `gain[0] < 0.` because the boundaries of the gain are 
```rs
min_value: f32::MIN,
max_value: f32::MAX,
```
meaning a gain of -1 is possible and would just reverse the waveform... don't know why someone would do that, but maybe there is some use cases for that.

For the threshold idea, as it's ok to have a gain of 10 (if your input signal is very low for example), we would need to check a diff then. But indeed maybe that would be interesting to have something to take into account `assert_float_eq!(db_to_lin(0.), 1., abs <= 1e-8);` (i.e. some volume fader controlled in dB)

b-ma: Saying for example that a diff of -120dB (1e-6) compared to 0 or 1 is a priori far from audible so we can clamp to use the very fast track (i.e. `silence()` or `clone()`)

orottier: Yep, nice catch

PR #223, opened 2022-09-14
Feature/count mode

orottier:
For #155

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,8 +67,6 @@ deviations could not be avoided:
 
 Our main limitations include:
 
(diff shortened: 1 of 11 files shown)


b-ma:
Hey looks good, looks like a good tradeoff between spec, consistency and readability to me.

I just wonder if shouldn't do the same for the `ChannelConfig` for consistency then?
https://github.com/orottier/web-audio-api-rs/blob/06b3136dc01681d13d0bf9600f5d7267dd47debd/src/node/mod.rs#L148

PR #224, opened 2022-09-19
HRTF panning

orottier:
#44 
I used `hrtf` crate and an ircam HRTF file. Convert azimuth/elevation to required `sample_vec` for the `HrtfContext`
It took me a lot of time to get the coordinate system right and feed it into the `hrtf` crate properly.
The result is subtle, but audible for me. I created a spectogram of the result and it shows more lower frequencies when the panner object is behind the listener.

HRTF file downloaded from http://recherche.ircam.fr/equipes/salles/listen/download.html
after examining http://recherche.ircam.fr/equipes/salles/listen/sounds.html I found this one works best for me.

TODO:
- [x] tail time
- [x] update panning model type while running
- [ ] choose the best HRTF file - check license
- [x] embed HRTF file in library


orottier:
It's not entirely clear to me if there are license or attribution requirements for the Ircam files, or we can just ship them inside our lib

b-ma:
Hey, didn't have the time to test or look at the code yet but looks promising, congrats!

Just a few notes on the HRTF file itself (I just asked to a colleague from the spat team):
- the database you use is outdated, there is a more recent one called Bili that has been released since (can't find the link right now)
- integrating the file directly into the library may also be touchy because the files are copyrighted. I will need to ask to the head of the team for more information and know if we can do that (considering it's an open-source project etc.)

orottier:
Cool, thanks for digging in. I won't merge before we get the final answer on this

b-ma:
(didn't receive any answer yet about the HRTF file... I will retry next week)

b-ma:
Hey, finally some (quite good) news on that point. I managed to have a talk with the head of the spat team who is ok that we use one of their file. Few points:
- He confirmed the file you used is quite old and pointed me to a more recent database, (unfortunately I don't have my notebook right now, so I'll give you the link tomorrow...). The database works with the SoFa protocol so we might have to do some work to just retrieve what we need (I could probably help here using this https://github.com/Ircam-RnD/serveSofaHrir), also there are each time 2 versions of the HRTF: one for implementation based on convolution and one based on some kind of (probably much more efficient) bank of IIR filters (but it doesn't seem the lib you used implements that strategy and while I can vaguely figure out the principles, I wouldn't try implementing that myself...)
- We should also add some Acknowledgement in the README saying where the impulses comes from, and also that any problems comes from our implementation and not from the HRTF files

b-ma:
actually this could be interesting to check: https://github.com/Ircam-RnD/binauralModeled

orottier:
@b-ma shall I merge this version? Or do you have the more recent databases at hand?

b-ma:
> Or do you have the more recent databases at hand?

Arg no, I can send you the link tomorrow

> shall I merge this version?

Let me just check w/ the spat team that they are ok with the acknowledgements

orottier:
Alright!

Issue #219, opened 2022-09-05
Implement `EventHandler` attributes


orottier:
Search for `of type EventHandler` in https://webaudio.github.io/web-audio-api/ 

Simplest implementation is probably an `Option<Box<dyn Fn(Event)>>`

b-ma:
I've started to think about it and to try some things, nothing really working and I'm doing things outside the lib to not mess around too much (I guess once we have the right approach it should quite straightforward to port into the lib).

I will try to make a working prototype these next days, or at least a failing one that would properly represent showing the different questions/problems I end with

orottier:
Great.
Yeah, the `removeEventListener` is tricky. Let's see how far we can come without overcomplicating stuff

b-ma:
Regarding your work on AudioRenderCapacity, I think it's maybe better to go with the simple `oneneded`, etc API, probably building the more advanced one on top of that would be feasible after, what do you think?

There is one thing I had in mind but I'm not sure it's the right approach, so your enlightenments would be welcome :) 
My idea was somehow to delegate all this to the context, so that we have only one thread that waits for incoming events from control thread with only one communication channel. But my impression is that it requires the have some book-keeping of the AudioNode ids in the control thread too. 

as pseudo code, would mean something like:
```
node.onended(func) {
    context.register_callback(node.id, type, func)
    // context store callback in some nested map: NodeId => EventType => Func
}

// on render thread
renderer.dispatch_event(type)
-> sends a { nodeId, type } message for the control / callback thread
```
Also that would mean that when the renderer are dropped, this would trigger some cleaning in the callback map

Not sure I'm clear, but do you think that's a good way to go?

orottier:
It is probably easier to use many channels instead of one.
So when a `AudioBufferSourceNode` is created, set up a `onended_channel`. Put the `Sender` inside the renderer. The `Receiver` alongside an `Arc<Mutex<Callback>>` can be added to a new (singleton) struct `EventLoop` 
In a dedicated thread, the `EventLoop` can poll all receivers and run the callback when appropriate.
This requires very little bookkeeping so should be relatively simple, no?
Dropping a renderer could send a `None` value via the channel to let the `EventLoop` drop the callback

orottier:
You can use https://docs.rs/crossbeam-channel/latest/crossbeam_channel/macro.select.html to poll all receivers concurrently in the EventLoop thread

orottier:
[...]

PR #226, opened 2022-10-05
Add AudioRenderCapacity functionality

### LINKED ISSUE #218: Implement AudioRenderCapacity
https://webaudio.github.io/web-audio-api/#audiorendercapacity

orottier:
Fixes #218

### DIFF
--- a/src/capacity.rs
+++ b/src/capacity.rs
@@ -1,0 +1,164 @@
+use crossbeam_channel::{Receiver, Sender};
+use std::sync::{Arc, Mutex};
+
+#[derive(Copy, Clone)]
+pub(crate) struct LoadValueData {
+    pub render_timestamp: f64,
+    pub load_value: f64,
+}
+
+/// Options for constructing an `AudioRenderCapacity`
+pub struct AudioRenderCapacityOptions {
+    /// An update interval (in seconds) for dispaching [`AudioRenderCapacityEvent`]s
+    pub update_interval: f64,
+}
+
+impl Default for AudioRenderCapacityOptions {
+    fn default() -> Self {
+        Self {
+            update_interval: 1.,
+        }
+    }
+}
+
+/// Performance metrics of the rendering thread
+#[derive(Clone, Debug)]
+pub struct AudioRenderCapacityEvent {
+    timestamp: f64,
+    average_load: f64,
+    peak_load: f64,
+    underrun_ratio: f64,
+}
+
+impl AudioRenderCapacityEvent {
+    fn new(timestamp: f64, average_load: f64, peak_load: f64, underrun_ratio: f64) -> Self {
+        Self {
+            timestamp,
+            average_load: (average_load * 100.).round() / 100.,
+            peak_load: (peak_load * 100.).round() / 100.,
+            underrun_ratio: (underrun_ratio * 100.).ceil() / 100.,
+        }
+    }
+
+    /// The start time of the data collection period in terms of the associated AudioContext's currentTime
+    pub fn timestamp(&self) -> f64 {
+        self.timestamp
+    }
+
+    /// An average of collected load values over the given update interval
+    pub fn average_load(&self) -> f64 {
+        self.average_load
+    }
+
+    /// A maximum value from collected load values over the given update interval.
+    pub fn peak_load(&self) -> f64 {
+        self.peak_load
+    }
+
+    /// A ratio between the number of buffer underruns and the total number of system-level audio callbacks over the given update interval.
+    pub fn underrun_ratio(&self) -> f64 {
+        self.underrun_ratio
+    }
+}
+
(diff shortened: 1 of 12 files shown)


src/capacity.rs
```diff
@@ -0,0 +1,164 @@
...
+        let signal = LoadValueData {
+            render_timestamp: f64::NAN,
+            load_value: f64::NAN,
+        };
+        let _ = self.sender.send(signal);
+    }
+
+    /// An EventHandler for [`AudioRenderCapacityEvent`].
+    #[allow(clippy::missing_panics_doc)]
+    pub fn onupdate<F: FnMut(AudioRenderCapacityEvent) + Send + 'static>(&mut self, callback: F) {
```
orottier, line 161: This may collide with your work on #219 so let's make sure it aligns later

b-ma:
Hey, this looks really nice, I will have a closer look tomorrow morning!

src/capacity.rs
```diff
@@ -0,0 +1,164 @@
...
+        let mut timestamp: f64 = 0.;
+        let mut load_sum: f64 = 0.;
+        let mut counter = 0;
+        let mut peak_load: f64 = 0.;
+        let mut underrun_sum = 0;
+
+        let mut next_checkpoint = options.update_interval;
+        std::thread::spawn(move || {
+            for item in receiver {
+                let LoadValueData {
```
b-ma, line 110: Not sure of this naming, a bit too general I think, maybe something more verbose but clear? something like `AudioRenderCapacityLoadData`?

src/context/online.rs
```diff
@@ -220,4 +220,12 @@ impl AudioContext {
         let opts = node::MediaElementAudioSourceOptions { media_element };
         node::MediaElementAudioSourceNode::new(self, opts)
     }
+
+    /// Returns an [`AudioRenderCapacity`] instance associated with an AudioContext.
+    #[must_use]
+    pub fn render_capacity(&self) -> AudioRenderCapacity {
+        let (sender, receiver) = crossbeam_channel::bounded(1);
```
b-ma, line 227: Could we somehow reuse the sender and receiver (just clone) so that we have only one sender in audio thread? This would allow to:
- send only 1 message between threads, even if `context.render_capacity()` has been called several time
- avoid the `Vec` and the loop in render thread

src/context/online.rs
```diff
@@ -220,4 +220,12 @@ impl AudioContext {
...
+        self.base().add_audio_capacity_listener(sender.clone());
+        AudioRenderCapacity::new(sender, receiver)
```
b-ma, line 229: Not sure this is the right thing to do (and somehow related to the previous message), from what I understand from the spec, this should rather be some sort of singleton no?
- 1 call, create and return the instance
- subsequent calls, return a clone() ?

src/capacity.rs
```diff
@@ -0,0 +1,164 @@
...
+
+                counter += 1;
+                load_sum += load_value;
+                peak_load = peak_load.max(load_value);
+                if load_value > 1. {
+                    underrun_sum += 1;
+                }
+
+                if render_timestamp >= next_checkpoint {
+                    let event = AudioRenderCapacityEvent::new(
```
b-ma, line 128: I understand there is probably no need to limit the precision here (problems of fingerprinting are not the same), but maybe we could just add a comment referring to (https://webaudio.github.io/web-audio-api/#dom-audiorendercapacityevent-averageload)?

src/render/thread.rs
```diff
@@ -133,8 +139,38 @@ impl RenderThread {
 
     // This code is not dead: false positive from clippy
     // due to the use of #[cfg(not(test))]
```
b-ma, line 141: Is this comment still relevant?

src/capacity.rs
b-ma: Actually, I didn't went further as for now, I'm a bit stuck in big refactoring issues with no mental space left :)... but I think the API I ended up was pretty much the same, really cool to have this example

b-ma:
I'm just a bit concerned with the singleton thing, but except that issue, that's a really nice and interesting work!

orottier:
Interesting point about the singleton. I may have misunderstood the spec.
The `AudioContext.renderCapacity` value is indeed an attribute, not a function, so it would not make sense to return different instances per call.
This does make me wonder what happens if you call `start` multiple times, does it just take the latest value of the update interval? I will check that out in the browser implementations

src/render/thread.rs
orottier: Removed, thanks

src/capacity.rs
orottier: done

src/capacity.rs
orottier: Done

b-ma:
Looks all good to me!

Issue #216, opened 2022-09-02
Let user specify output device for online AudioContext


orottier:
Official spec will land soon:
https://github.com/WebAudio/web-audio-api/issues/2400#issuecomment-1101671575

It should be quite straightforward for us to port to the cpal and cubeb backend

orottier:
Spec has landed. I will whip up a simple implementation first where you can specify the output device during AudioContext construction.
The feature to change the output device while the context is running will be very tricky. `cpal` may not offer the desired functionality - https://github.com/RustAudio/cpal/issues/373 . Cubeb does have some callbacks for device switching. It probably warrants a dedicated ticket to handle disconnection of the current device.
Additional complexity is that the sample rate may change in this case which requires extra bookkeeping throughout the render thread

PR #227, opened 2022-10-25
Specify output device at AudioContext creation

orottier:
This is the first of three PRs for #216 

1. Specify output device at AudioContext creation
2. Change output device for running AudioContext
3. Create AudioContext without output

### DIFF
--- a/examples/sink_id.rs
+++ b/examples/sink_id.rs
@@ -12,8 +12,10 @@ fn main() {
     let sink_id = std::io::stdin().lines().next().unwrap().unwrap();
 
     // Create an audio context (default: stereo);
-    let mut options = AudioContextOptions::default();
-    options.sink_id = Some(sink_id);
+    let options = AudioContextOptions {
+        sink_id: Some(sink_id),
+        ..AudioContextOptions::default()
+    };
 
     let context = AudioContext::new(options);
     println!("Playing beep for sink {:?}", context.sink_id());
--- a/examples/sink_id.rs
+++ b/examples/sink_id.rs
@@ -16,6 +16,7 @@ fn main() {
     options.sink_id = Some(sink_id);
 
     let context = AudioContext::new(options);
+    println!("Playing beep for sink {:?}", context.sink_id());
 
     // Create an oscillator node with sine (default) type
     let osc = context.create_oscillator();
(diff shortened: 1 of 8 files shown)


examples/sink_id.rs
```diff
@@ -0,0 +1,29 @@
+use web_audio_api::context::{AudioContext, AudioContextOptions, BaseAudioContext};
```
orottier, line 1: This example works best with `--features cubeb`
Cpal does not allow you to choose speaker output when a headset is plugged in

b-ma:
Hey, sorry for the long silence, I will have a shot in this tomorrow morning

src/io/backend_cpal.rs
```diff
@@ -89,7 +90,23 @@ impl AudioBackend for CpalBackend {
     where
         Self: Sized,
     {
-        let mut builder = StreamConfigsBuilder::new();
+        let host = cpal::default_host();
+        log::info!("Host: {:?}", host.id());
```
b-ma, line 94: Maybe this should be in the "None" branch, as this is not used in the other branch, the log might be confusing if sink_id is set

src/io/backend_cubeb.rs
```diff
@@ -195,48 +201,55 @@ impl AudioBackend for CubebBackend {
...
         let buffer_size = buffer_size_req.max(min_latency);
+        let device_id = options.sink_id.as_ref().and_then(|d| {
+            Self::enumerate_devices()
+                .into_iter()
+                .find(|e| e.device_id() == d)
+                .map(|e| *e.device().downcast::<DeviceId>().unwrap())
+        });
 
         let stream = match number_of_channels {
             // so sorry, but I need to constify the non-const `number_of_channels`
```
b-ma, line 212: I know you don't like these, but a macro would be appropriate here I think :), as it could generate the code according to MAX_CHANNELS

b-ma:
Except my two small comments, looks all good to me and the example is working well !

src/io/backend_cpal.rs
orottier: Good point.
It is a bit sloppy, the other branch eventually also calls into the default host as well, so the statement is not wrong per se.

src/io/backend_cubeb.rs
orottier: You are probably right. But I don't have the expertise to come up with the macro declaration.
Reading https://stackoverflow.com/questions/33751796/is-there-a-way-to-count-with-macros this cannot be done easily with `macro_rules!`, and we would have to resort to procedural macros which would be really overkill.
I would love to be proven wrong though! Any ideas?

Issue #187, opened 2022-07-25
`cubeb` audio backend improvements


orottier:
- [ ] Fix occasional segfaults when interacting with the `stream` (suspend, resume, get latency)
- [ ] Support other channel configs than STEREO
- [ ] Setup a single stream for both input and output handling (currently two streams are used, leading to time drifts)

orottier:
A solution for the segfaults could be to run the rust implementation of the audiounit backend: https://github.com/mozilla/cubeb-rs/pull/75#discussion_r946494221

Issue #183, opened 2022-07-20
Resample when the requested sample rate is not available on the audio output device


orottier:
You can specify the requested sample rate via `AudioContextOptions.sample_rate`
If this sample rate is not available at the audio output device, we fall back to a value that is supported.
However, the spec mandates that in that case, the audio graph should still render at the user requested sample rate, and only at the DestinationNode one should resample to the rate that is required for the output device:

https://webaudio.github.io/web-audio-api/#AudioContext-constructors
> If contextOptions.[sampleRate](https://webaudio.github.io/web-audio-api/#dom-audiocontextoptions-samplerate) is specified, set the [sampleRate](https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-samplerate) of this [AudioContext](https://webaudio.github.io/web-audio-api/#AudioContext) to this value. Otherwise, use the sample rate of the default output device. If the selected sample rate differs from the sample rate of the output device, this [AudioContext](https://webaudio.github.io/web-audio-api/#AudioContext) MUST resample the audio output to match the sample rate of the output device.
> Note: If resampling is required, the latency of the AudioContext may be affected, possibly by a large amount.

PR #232, opened 2022-10-31
Feature/set sink

orottier:
This is the second of three PRs for https://github.com/orottier/web-audio-api-rs/issues/216

1. #227 
2. Change output device for running AudioContext
3. Create AudioContext without output

You will notice a severe lack of tests for this complicated feature. That is because the functionality is only exposed in online `AudioContext` which need an actual audio backend. For now, because that will be changed in PR number 3.

You can test it with `cargo run --release --example sink_id` but notice that `cpal` only lists your default audio output, so you won't be able to switch from headphones to builtin speakers (on OSX). I added a virtual passthrough device to use for testing. Sad though. It is not really testable with backend `cubeb` because it suffers from segfaults when closing a stream. Known issue #187  :(

### DIFF
--- a/src/context/online.rs
+++ b/src/context/online.rs
@@ -1,6 +1,6 @@
 //! The `AudioContext` type and constructor options
 use crate::context::{AudioContextState, BaseAudioContext, ConcreteBaseAudioContext};
-use crate::io::{self, AudioBackend, ControlThreadInit};
+use crate::io::{self, AudioBackend, ControlThreadInit, RenderThreadInit};
 use crate::media::{MediaElement, MediaStream};
 use crate::node::{self, ChannelConfigOptions};
 use crate::AudioRenderCapacity;
@@ -57,6 +57,8 @@ pub struct AudioContext {
     backend: Box<dyn AudioBackend>,
     /// Provider for rendering performance metrics
     render_capacity: AudioRenderCapacity,
+    /// Initializer for the render thread (when restart is required)
+    render_thread_init: RenderThreadInit,
 }
 
 impl BaseAudioContext for AudioContext {
@@ -94,7 +96,9 @@ impl AudioContext {
     #[allow(clippy::needless_pass_by_value)]
     #[must_use]
     pub fn new(options: AudioContextOptions) -> Self {
-        let (backend, control_thread_init) = io::build_output(options);
+        let (control_thread_init, render_thread_init) = io::thread_init();
+        let backend = io::build_output(options, render_thread_init.clone());
+
         let ControlThreadInit {
             frames_played,
             ctrl_msg_send,
@@ -118,6 +122,7 @@ impl AudioContext {
             base,
             backend,
             render_capacity,
+            render_thread_init,
         }
     }
 
@@ -128,7 +133,7 @@ impl AudioContext {
     // it to the audio subsystem, so this value is zero. (see Gecko)
     #[allow(clippy::unused_self)]
     #[must_use]
-    pub const fn base_latency(&self) -> f64 {
+    pub fn base_latency(&self) -> f64 {
         0.
     }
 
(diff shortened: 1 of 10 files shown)


src/io/backend_cpal.rs
```diff
@@ -1,25 +1,23 @@
 //! Audio IO management API
```
orottier, line 1: I thoroughly rewrote this file, so I'm closing #51

src/context/online.rs
```diff
@@ -140,15 +165,73 @@ impl AudioContext {
...
+    pub fn sink_id(&self) -> Option<String> {
+        self.backend.lock().unwrap().sink_id().map(str::to_string)
+    }
+
+    /// Update the current audio output device.
+    ///
+    /// This function operates synchronously and might block the current thread. An async version
+    /// is currently not implemented.
+    #[allow(clippy::needless_collect, clippy::missing_panics_doc)]
+    pub fn set_sink_id_sync(&self, sink_id: String) -> Result<(), Box<dyn Error>> {
```
orottier, line 186: This is the meat of the PR.

1. check if sink is already active
2. validate sink
3. acquire necessary locks (avoid issues with concurrent calls)
4. shut down current render thread and recover the audio graph
5. boot up new thread, and ship the audio graph
6. release locks

src/context/online.rs
```diff
@@ -140,15 +165,73 @@ impl AudioContext {
...
+
+        // acquire the audio graph from the current render thread, shutting it down
+        let (graph_send, graph_recv) = crossbeam_channel::bounded(1);
+        let message = ControlMessage::Shutdown { sender: graph_send };
+        ctrl_msg_send.send(message).unwrap();
+        let graph = graph_recv.recv().unwrap();
+
+        // hotswap the backend
+        let options = AudioContextOptions {
+            sample_rate: Some(self.sample_rate()),
```
orottier, line 214: If this sample rate is not available on the new sink, bad things will happen. That is why #183 exists

src/context/online.rs
```diff
@@ -140,15 +165,73 @@ impl AudioContext {
     /// the time at which the first sample in the buffer is actually processed
     /// by the audio output device.
     #[must_use]
+    #[allow(clippy::missing_panics_doc)]
     pub fn output_latency(&self) -> f64 {
-        self.backend.output_latency()
+        self.backend.lock().unwrap().output_latency()
```
b-ma, line 170: not sure of the thread layout here, but can't this lock() be a problem regarding the audio thread?

b-ma: that's something we might want to call periodically during rendering I guess

src/context/online.rs
```diff
@@ -140,15 +165,73 @@ impl AudioContext {
...
+            latency_hint: AudioContextLatencyCategory::default(), // todo reuse existing setting
+            sink_id: Some(sink_id),
+        };
+        *backend_guard = io::build_output(options, self.render_thread_init.clone());
+
+        // send the audio graph to the new render thread
+        let message = ControlMessage::Startup { graph };
+        ctrl_msg_send.send(message).unwrap();
+
+        self.base().set_state(AudioContextState::Running);
```
b-ma, line 224: if the context was suspended maybe we want to keep it suspended?

Actually, I didn't read the thing carefully, but the spec seems to say:

>If wasRunning is true:
>
> Set the [[[rendering thread state]]](https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-rendering-thread-state-slot) on the [AudioContext](https://webaudio.github.io/web-audio-api/#AudioContext) to "suspended".

https://webaudio.github.io/web-audio-api/#ref-for-dom-audiocontext-setsinkid%E2%91%A0

But as we already differ on the initial state, I don't really know... 
Just keeping the state as it was could be a good tradeoff?

b-ma:
That's quite a thing, I'm really not sure I understand everything (I'm pretty I don't actually :), so maybe my comments are not very relevant.

Actually it makes me think it would be quite nice to have some graph of who lives where in terms of threads, I'm always a bit lost in this part of the code.

In any case, the example seems to work on my side switching between some virtual device and real output

src/context/online.rs
orottier: I realize now the `backend` naming is confusing.
The `backend` object live in the control thread, so locking is no issue.
A better name would be `backend_manager`, because it passes instructions from the control thread (start, pause, close) and collects info from the actual backend in the render thread (latency, sink_id).
I will update the PR by renaming this struct

src/context/online.rs
orottier: Yes, very good point. I will update the PR

orottier:
> Actually it makes me think it would be quite nice to have some graph of who lives where in terms of threads, I'm always a bit lost in this part of the code.

Yes, I created a new issue for this: #235

PR #234, opened 2022-11-03
Render audio graph without emitting to the speakers 

orottier:
This is the third PR for #216 
There are some leftover To-dos and your review comments that I will address soon. Thanks for having a look!


src/context/online.rs
```diff
@@ -44,10 +44,15 @@ pub struct AudioContextOptions {
...
     /// tradeoffs between audio output latency and power consumption
     pub latency_hint: AudioContextLatencyCategory,
-    /// Sample rate of the audio Context and audio output hardware
+
+    /// Sample rate of the audio context and audio output hardware
     pub sample_rate: Option<f32>,
-    /// The audio output device - use `None` for the default device
-    pub sink_id: Option<String>,
+
+    /// The audio output device
```
orottier, line 51: This API is terrible and I know it. The spec is not much better. I will explore other options

orottier:
Hi @b-ma I would love your opinion on my latest API-compliance decision:

```
commit 3e33da3636f7712474a798ac1ab5a00dcea8a886 

    Drop double `Option` from AudioContextOptions.sink_id

    Before: `Option<Option<String>>`
    /// - use `None` for the default device
    /// - use `Some(None)` to process the audio graph without playing through an audio output device.
    /// - use `Some(Some(sinkId))` to use the specified audio sink id, obtained with `enumerate_devices`

    After: `Option<String>`
    /// - use `None` to process the audio graph without playing through an audio output device.
    /// - use `Some("")` for the default audio output device
    /// - use `Some(sinkId)` to use the specified audio sink id, obtained with `enumerate_devices`

    We now require the user to specify `Some("")` for the default audio
    device. In the docs we suggest to use the `::default()` implementation
    so you are not required to think about it for most usecases.
```

b-ma:
Hey! yup I try to check that today or tomorrow

b-ma:
Hey, after reading the spec (which I agree is quite hard to understand) and read again (a bit fast) [https://github.com/WebAudio/web-audio-api/issues/2400](https://github.com/WebAudio/web-audio-api/issues/2400), I would personally go for the very simple:

```
`String`
    /// - use `""` for the default audio output device
    /// - use `"none"` to process the audio graph without playing through an audio output device.
    /// - use `sinkId` to use the specified audio sink id, obtained with `enumerate_devices`
```

This is not very elegant, but much more simple in my opinion... and as the `sinkId`s are generated by the `enumerate_devices` code, we can guarantee there won't be any conflicts (but I must also admit I don't really get why these `AudioSyncInfo` and `AudioSyncOptions` types are defined in the spec)

Maybe we should mark the whole thing as unstable until this is implemented in Chrome or Firefox so that we can compare with what they did, to see if we didn't miss something?

src/context/online.rs
```diff
@@ -9,6 +11,20 @@ use crate::AudioRenderCapacity;
 use std::error::Error;
 use std::sync::Mutex;
 
+/// Check if the provided sink_id is available for playback
+///
+/// It should be `None` of `Some(sinkId)` where `sinkId` is returned from `enumerate_devices`
```
b-ma, line 16: It should be `None` ~of~ or `Some(sinkId)` ?

orottier:
Thanks for the suggestion. Way simpler indeed.
I will leave it this way for the 0.x versions

PR #237, opened 2022-11-09
Feature/events

b-ma:
Hey, here is a first draft concerning the Events (cf. #219). It's really not finished nor clean yet but the minimal example works and I think it's better to have your feedback before continuing.

Few remarks:
- I ended up using raw `u64` to store the ids, but it could be cleaner if we just merged `AudioNodeId` (control thread) and `NodeIndex` (render thread) into the same struct (this distinction is a bit weird and confusing in my opinion, but maybe I missed something).
- I'm really not sure about the namings 
- I think there is a problem with cleaning the nodes on the render thread: when debugging using raw `println!` the event was added on each block until the program exit, which should not be the case (that's why I added a flag in the `AudioBufferSourceRenderer` to make sure to send the event only once, but it should not be necessary I think)
- I wanted to check for the events after all rendering was done, similar to `RenderThread::handle_control_messages()`, but with the way the node are cleaned I ended up doing this in the middle the render loop
- Only works on online contexts (don't know if it make sens for offline context, but didn't check in the spec...), and the build probably crash w/ cubeb for now..
- Probably lots of small weird things... :)

One thing I'm quite happy with, is that I think it should be quite straightforward to use the traits for the contexts' events too.

Let me know what you think!


src/events.rs
```diff
@@ -0,0 +1,61 @@
...
+// maybe not the right, but needs to review according to type EventHandler in capacity.rs
+pub(crate) struct EventHandlerInfos {
+    // could be Option w/ None meaning that its a context event (cf. onSinkChange, onStateChange, etc.)
+    pub node_id: u64,
+    pub event_type: EventType,
+    pub callback: Box<&'static (dyn Fn() + Send + Sync + 'static)>,
+}
+
+// for control thread entities
+pub(crate) trait EventListener {
```
orottier, line 17: I don't think this trait needs to exist. It can just be a method on the ConcreteBase

src/events.rs
```diff
@@ -0,0 +1,61 @@
...
+pub(crate) struct EventEmitterMessage {
+    // could be Option w/ None meaning that its a context event
+    pub node_id: u64, // we use raw u64 to
+    // could be Option w/ None meaning the node is dropped on the render thread
+    // and listeners can be cleared
+    pub event_type: EventType,
+}
+
+// for render thread entities
+// @todo - maybe should be a (bounded?) queue
```
orottier, line 48: A bounded queue means we either drop messages, or the render thread could run into a blocking operation.
Neither would be acceptable.
An unbounded queue allocates though, thus violating our 'render thread should not allocate' principle. We could look for another data structure at a later time

src/node/audio_buffer_source.rs
```diff
@@ -402,18 +432,33 @@ impl AudioProcessor for AudioBufferSourceRenderer {
         // 2. the duration has been reached.
         if scope.current_time >= stop_time || self.render_state.buffer_time_elapsed >= duration {
             output.make_silent(); // also converts to mono
+
+            // @note: we need this check because this is called a until the program
+            // ends, such as if the node was never removed from the graph
```
orottier, line 437: An AudioProcessor is never dropped from the graph as long as the AudioNode object is still alive in the control thread. You could manually `drop` it from the example and see if that changes anything. The reason we keep it is because the control thread could still connect it to other nodes, and maybe even restart it (not allowed for buffer source node I think..)
In any case the check is needed because the cleanup process is not instant. It peels of 1 layer per render quantum. So if there were any nodes feeding into this one, they are cleaned first, then the next render call this node is dropped.

src/render/graph.rs
```diff
@@ -45,6 +47,10 @@ impl Node {
             .process(&self.inputs[..], &mut self.outputs[..], params, scope)
     }
 
+    fn event_emitter(&mut self) -> Option<&mut dyn EventEmitter>{
```
orottier, line 50: I think it would be better to put the event emitter in the `RenderScope` object so you can directly call it from the `process` call of the AudioProcessor. We of course hide it from the public API

src/events.rs
```diff
@@ -0,0 +1,61 @@
...
+pub trait EventEmitter {
```
orottier, line 49: I don't think this trait needs to exist. It can just be a method inside the render thread

src/render/graph.rs
```diff
@@ -407,6 +417,25 @@ impl Graph {
...
+                let event_sender = event_sender.unwrap();
+
+                if let Some(event_emitter) = node.event_emitter() {
+                    if let Some(event) = event_emitter.get_stagging_event() {
+                        let message = EventEmitterMessage {
+                            node_id: index.0, // this is inconsistent maybe we could use AudioNodeId everywhere
+                            event_type: event,
+                        };
+
+                        let _r = event_sender.send(message);
```
orottier, line 434: For a bounded channel, you must use `try_send` to avoid blocking

src/render/processor.rs
```diff
@@ -54,6 +55,10 @@ pub trait AudioProcessor: Send {
         params: AudioParamValues,
         scope: &RenderScope,
     ) -> bool;
+
+    fn event_emitter(&mut self) -> Option<&mut dyn EventEmitter> {
```
orottier, line 59: This is now part of the public API, which I think we should avoid

orottier (commented):
Hey, impressive work. I think this is going in the right direction but I also feel like it is implemented in a way that is bit overly complex. I left quite a few comments. How do you want to continue? You can continue working on it but I am also fine iterating on it myself and then discuss with you again.

Regarding the AudioNodeId - I think it would make a lot of sense to keep them shared between the control and render thread. Very good suggestion, I never put too much thought in it

src/context/concrete_base.rs
```diff
@@ -185,6 +190,27 @@ impl ConcreteBaseAudioContext {
             LISTENER_PARAM_IDS.end,
         );
 
+        // (?) only for online context
+        if event_channel.is_some() {
+            // init event loop
+            let event_channel = event_channel.unwrap();
+            let event_handlers = base.inner.event_handlers.clone();
+            //
+            std::thread::spawn(move || loop {
```
orottier, line 199: Instead of `loop` and `recv` you can use 
`for message in event_channel.iter() { `

src/context/concrete_base.rs
```diff
@@ -185,6 +190,27 @@ impl ConcreteBaseAudioContext {
...
+                // (?) this thread is dedicated to event so we can block
+                if let Ok(message) = event_channel.recv() {
+                    let handlers = event_handlers.lock().unwrap();
+                    // find EventHandlerInfos that match messsage
+                    if let Some(infos) = handlers.iter().find(|item| {
+                        item.node_id == message.node_id && item.event_type == message.event_type
+                    }) {
+                        (infos.callback)();
```
orottier, line 207: Should we drop the handler at this point? Or can it fire multiple times?

src/context/mod.rs
```diff
@@ -26,8 +26,8 @@ const LISTENER_PARAM_IDS: Range<u64> = 2..11;
 /// Unique identifier for audio nodes.
 ///
 /// Used for internal bookkeeping.
-#[derive(Debug)]
-pub(crate) struct AudioNodeId(u64);
+#[derive(Debug, Copy, Clone)]
+pub(crate) struct AudioNodeId(pub u64);
```
orottier, line 30: I deliberately marked the u64 as private for type safety. This way users could never construct a malicious NodeId that could crash the render thread. I think if we share the AudioNodeId between control and render thread you do not need to mark it public

examples/event_emitter.rs
```diff
@@ -0,0 +1,28 @@
...
+    let audio_context = AudioContext::default();
+
+    let file = File::open("samples/sample.wav").unwrap();
+    let buffer = audio_context.decode_audio_data_sync(file).unwrap();
+
+    let src = audio_context.create_buffer_source();
+    src.connect(&audio_context.destination());
+    src.set_buffer(buffer);
+
+    // @todo - should receive an event
```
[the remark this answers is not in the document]
b-ma: No sure what you mean, my comment was about that it still unfinished because the callback signature should be 
```rs
|e: Event| {
   println!("{:?}", event.type);
}
```
cf. https://dom.spec.whatwg.org/#interface-event

src/events.rs
```diff
@@ -0,0 +1,61 @@
...
+pub enum EventType {
+    Ended
+}
+
+// maybe not the right, but needs to review according to type EventHandler in capacity.rs
+pub(crate) struct EventHandlerInfos {
+    // could be Option w/ None meaning that its a context event (cf. onSinkChange, onStateChange, etc.)
+    pub node_id: u64,
+    pub event_type: EventType,
+    pub callback: Box<&'static (dyn Fn() + Send + Sync + 'static)>,
```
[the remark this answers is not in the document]
b-ma: I have to say I started with `<'a>` lifetime and it started polluting everything and I found this that fixed the problem (still black magic to me :)

src/events.rs
b-ma: Yup, would be good to do make a list of all possible event, just to see if it is even possible that a node (or context) triggers several events in the same render quantum

src/render/graph.rs
b-ma: Yeah that's a possibility indeed, I was in the model of

- thread.render()
- thread.check_event();
- thread.handle_incoming_messages();

but maybe I'm just overcomplicate things here

src/render/processor.rs
b-ma: yup, I made few things public just to make it work but of course this is not good, with a common `AudioNodeId`I think this particular issue would disappear

src/context/concrete_base.rs
b-ma: I think for context events (e.g. `onSinkChange`) we should not drop the handlers here because they might be triggered several times, my idea was to remove them when the node is dropped on the render thread.

Also the e.g. `AudioBufferSourceNode.onended` signature should probably be `Option<callback>` so that we can remove the handler manually in user code by passing `None`

src/context/mod.rs
b-ma: Yup, see above

b-ma:
> How do you want to continue? You can continue working on it but I am also fine iterating on it myself and then discuss with you again.

Yup please continue, seems you have a clearer picture than me on what should be done from now (I can contribute too on simple aspects if needed, i.e. not the ones that rely on `'static` stuff for examples :)

orottier:
Unifying the AudioNodeId and the render side NodeId brought me a lot of joy. I think the code is a lot cleaner and easier to follow now.

> I deliberately marked the u64 as private for type safety. This way users could never construct a malicious NodeId that could crash the render thread. I think if we share the AudioNodeId between control and render thread you do not need to mark it public

I put some more thought in this and since the `AudioNodeId` is only `pub(crate)` I think there is no real issue here and we can keep everything simple.

The PR is in good shape now I think. Do you want to add more event types and handlers? And maybe iterate more on the design (e.g. callbacks can take an event payload)? I would also feel fine with merging this already - perhaps only add some documentation

examples/event_emitter.rs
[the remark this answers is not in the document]
orottier: Ah I understand now. Indeed the todo remains

orottier:
Hey @b-ma, I have implemented onended for the Oscillator and ConstantSource.

I think we are ready to merge this, no?

Before releasing a v1.0, I think we should think about:
- removing an event listener
- event listeners with a payload, e.g. https://webaudio.github.io/web-audio-api/#OfflineAudioCompletionEvent
- unify interface with the RenderCapacity.on_update

b-ma:
Hey, yup sorry again (I'm a bit out of the game these times...)

But yup looks good to me!

PR #238, opened 2022-11-13
Fix convolver normalize type

b-ma:
Small fix regarding https://github.com/ircam-ismm/node-web-audio-api/issues/13


b-ma:
(damn created the branch from the event one, maybe was not the best idea), restarting...

PR #239, opened 2022-11-13
Fix convolver normalize

b-ma:
Small fix regarding https://github.com/ircam-ismm/node-web-audio-api/issues/13


orottier:
Thanks!

Issue #217, opened 2022-09-02
Support custom render quantum size


orottier:
On hold for now:
https://github.com/WebAudio/web-audio-api/pull/2469

But it would be nice to explore solutions for us.
A solution could be to make the render structs generic over the buffer size: `RenderThread<const N = RENDER_QUANTUM_SIZE>`
We should check if this is a breaking API change for users that implement the AudioProcessor trait

Issue #140, opened 2022-04-07
Publish v1 release candidate for public review


orottier:
Possibly announce on the rust audio forum and elsewhere for last minute feedback

orottier:
Released today 🎊

b-ma:
\o/

PR #242, opened 2022-11-24
Prepare public API for variable render quantum size

orottier:
I think we have settled on most of the public API surface. However in preparation of #217  I think one more modification must be made: remove the constant `RENDER_QUANTUM_SIZE`. 

Relatest to #76 and #140

Might have performance impact. I will add more benchmarks

### DIFF
--- a/src/render/quantum.rs
+++ b/src/render/quantum.rs
@@ -169,7 +169,7 @@ impl std::ops::Drop for AudioRenderQuantumChannel {
 /// Render thread audio buffer, consisting of multiple channel buffers
 ///
 /// This is a  fixed length audio asset of `render_quantum_size` sample frames for block rendering,
-/// basically a matrix of `channels * AudioRenderQuantumChannel` cf.
+/// basically a list of [`AudioRenderQuantumChannel`]s cf.
 /// <https://webaudio.github.io/web-audio-api/#render-quantum>
 ///
 /// The `render_quantum_size` is equal to 128 by default, but in future versions it may be equal to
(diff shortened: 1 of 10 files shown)


src/render/quantum.rs
```diff
@@ -75,9 +75,12 @@ impl AllocInner {
 
 /// Render thread channel buffer
```
orottier, line 76: `AudioRenderQuantumChannel` now has no public methods, except the `Deref(Mut) to [f32]`

src/render/quantum.rs
```diff
@@ -135,10 +138,10 @@ impl AudioRenderQuantumChannel {
 use std::ops::{Deref, DerefMut};
 
 impl Deref for AudioRenderQuantumChannel {
```
orottier, line 140: This here is the motivation of the PR. We stop hardcoding the length of the render quantum to 128, but leave it unspecified (we return a [slice](https://doc.rust-lang.org/std/primitive.slice.html), instead of an [array](https://doc.rust-lang.org/std/primitive.array.html)).
This incurs a small performance hit (a percent, looking at the benchmarks) because more bound checks need to be performed at runtime.
However, this opens the way for the "dynamic render quantum size"  feature without breaking the current API. With this PR merged, the `RENDER_QUANTUM_SIZE` (or, the value 128) is no longer present in the API surface

b-ma:
Ok I think I see the idea, looks good to me! 
(I think I even prefer how we can use the channels this, seems more natural in terms of API)

PR #244, opened 2022-11-30
AudioContext.onsinkchange + some fixes on event handling

orottier:
- implement onsinkchange, add test, use in sink_id example
- refactor event thread code into an EventLoop
- drop event handlers of type Ended when they have run
- use Fn for 1-time event handlers, FnMut for handlers that can run multiple times
- set state to Suspended for AudioContext while we change the sink id

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -1,4 +1,6 @@
 use crate::context::AudioNodeId;
+use crossbeam_channel::Receiver;
+use std::sync::{Arc, Mutex};
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub(crate) enum EventType {
@@ -19,4 +21,37 @@ pub(crate) struct TriggerEventMessage {
     // could be Option w/ None meaning the node is dropped on the render thread
     // and listeners can be cleared
     pub event_type: EventType,
+}
+
+#[derive(Clone, Default)]
+pub(crate) struct EventLoop {
+    callbacks: Arc<Mutex<Vec<EventHandler>>>,
+}
+
+impl EventLoop {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn run(&self, event_channel: Receiver<TriggerEventMessage>) {
+        let self_clone = self.clone();
+
+        std::thread::spawn(move || loop {
+            // this thread is dedicated to event handling so we can block
+            for message in event_channel.iter() {
+                let mut handlers = self_clone.callbacks.lock().unwrap();
+                // find EventHandlerInfos that matches messsage and execute callback
+                handlers
+                    .iter_mut()
+                    .filter(|item| {
+                        item.node_id == message.node_id && item.event_type == message.event_type
+                    })
+                    .for_each(|handler| (handler.callback)())
+            }
+        });
+    }
+
+    pub fn add_handler(&self, handler: EventHandler) {
+        self.callbacks.lock().unwrap().push(handler)
+    }
 }
--- a/src/events.rs
+++ b/src/events.rs
@@ -9,7 +9,7 @@ pub(crate) struct EventHandler {
     // could be optional meaning that its a context event (cf. onSinkChange, onStateChange, etc.)
     pub node_id: AudioNodeId,
     pub event_type: EventType,
-    pub callback: Box<dyn Fn() + Send + Sync + 'static>,
+    pub callback: Box<dyn FnMut() + Send + 'static>,
 }
 
 #[derive(Debug)]
--- a/src/events.rs
+++ b/src/events.rs
@@ -7,11 +7,25 @@ pub(crate) enum EventType {
     Ended,
 }
 
+pub(crate) enum Callback {
+    Once(Box<dyn FnOnce() + Send + 'static>),
+    Multiple(Box<dyn FnMut() + Send + 'static>),
+}
+
+impl Callback {
+    fn run(self) {
+        match self {
+            Self::Once(f) => (f)(),
+            Self::Multiple(mut f) => (f)(),
+        }
+    }
+}
+
 pub(crate) struct EventHandler {
     // could be optional meaning that its a context event (cf. onSinkChange, onStateChange, etc.)
     pub node_id: AudioNodeId,
     pub event_type: EventType,
-    pub callback: Box<dyn FnMut() + Send + 'static>,
+    pub callback: Callback,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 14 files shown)


b-ma:
Looks really nice!

Just dropping some notes here:

Parsing the spec, we have:
- [BaseAudioContext.onstatechange](https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-onstatechange) - not yet implemented I think, but should be quite simple now
- [BaseAudioContext.onsinkchange](https://webaudio.github.io/web-audio-api/#dom-audiocontext-onsinkchange) - implemented here
- [AudioRenderCapacity.onupdate](https://webaudio.github.io/web-audio-api/#dom-audiorendercapacity-onupdate) - implemented
- [OfflineAudioContext.oncomplete](https://webaudio.github.io/web-audio-api/#dom-offlineaudiocontext-oncomplete) - not implemented, but as you said probably not very relevant until we have the `async` version, in any case should not be complicated
- [AudioScheduledSourceNode.onended](https://webaudio.github.io/web-audio-api/#dom-audioscheduledsourcenode-onended) - implemented
- [ScriptProcessorNode.onaudioprocess](https://webaudio.github.io/web-audio-api/#dom-scriptprocessornode-onaudioprocess) - _deprecated_
- [AudioWorklet.onprocessorerror](https://webaudio.github.io/web-audio-api/#dom-audioworkletnode-onprocessorerror) - this is probably the more tricky / needs discussion one now

So it seems we are in a very good track!

Except for the AudioWorklet which propagate an Error, all of them propagate an [Event](https://dom.spec.whatwg.org/#event) which are quite complex in the JS spec but probably not really relevant for us except according to their types I think (even the `Event` of `AudioRenderCapacity` derives from this type cf. https://webaudio.github.io/web-audio-api/#audiorendercapacityevent). So maybe we could model this with an `enum`?

For the ability to remove the `EventHandler`, the only solution I can see would be to change, e.g.:
```rs
pub fn onsinkchange<F: FnMut() + Send + 'static>(&self, callback: F)
```
to
```rs
pub fn onsinkchange<F: FnMut() + Send + 'static>(&self, callback: Option<F>)
```

If callback is `None` the callback is removed from the list, that would be quite consistent with what we would do in JS I think (I can test that to be really sure, but I'm quite confident here)

```js
audioContext.onsinkchange = (e) => doSomething();
// later
audioContext.onsinkchange = null; // would remove the handle
```

And the current behavior of calling the `onsinkchange` with a different `callback` replacing the previously registered one is also the expected behavior in JS as it is an attribute, eg.:

```
audioContext.onsinkchange = (e) => doSomething();
audioContext.onsinkchange = (e) => doOtherSomething(); // replace the previous callback
```

What do you think?

In any case all these change really seems to go in the right direction

src/events.rs
```diff
@@ -1,22 +1,71 @@
...
 }
 
 pub(crate) struct EventHandler {
-    // could be optional meaning that its a context event (cf. onSinkChange, onStateChange, etc.)
-    pub node_id: AudioNodeId,
-    pub event_type: EventType,
-    pub callback: Box<dyn Fn() + Send + Sync + 'static>,
+    pub event: Event,
+    pub callback: Callback,
+}
```
b-ma, line 28: Not sure the naming is really consistent with the spec, as far as I understand it, in the spec `EventHandler` is what is called `Callback` here. Probably not really important as it is `pub(crate)` but could be confusing

src/render/processor.rs
```diff
@@ -22,17 +22,13 @@ pub struct RenderScope {
     pub sample_rate: f32,
 
     pub(crate) node_id: Cell<AudioNodeId>,
-    pub(crate) event_sender: Option<Sender<TriggerEventMessage>>,
+    pub(crate) event_sender: Option<Sender<Event>>,
 }
 
 impl RenderScope {
-    pub(crate) fn send_event(&self, event_type: EventType) {
+    pub(crate) fn send_ended_event(&self) {
```
b-ma, line 29: Feels a bit less generic to me than the previous version, but as it is the only event nodes can currently send this is probably very ok

orottier: Yes you are right, we are losing a bit of flexibility here, but it simplifies the code.
Whenever new types of events are added to the spec we can change it back again

src/events.rs
orottier: Valid point, I will rename them

orottier:
> For the ability to remove the EventHandler, the only solution I can see would be to change, e.g.:

Yeah, I think this strikes a nice balance between simplicity, rust compatibility and coherence with how it works in JS.
So basically we decide only one callback can be active at all times. We will not introduce an `addEventListener`-like system.

I will merge this version for now and start working on `pub fn onsinkchange<F: FnMut() + Send + 'static>(&self, callback: Option<F>)` in a new PR.

> Except for the AudioWorklet which propagate an Error, all of them propagate an [Event](https://dom.spec.whatwg.org/#event) which are quite complex in the JS spec but probably not really relevant for us except according to their types I think (even the Event of AudioRenderCapacity derives from this type cf. https://webaudio.github.io/web-audio-api/#audiorendercapacityevent). So maybe we could model this with an enum?

I need to think about this a bit more. An enum will definitely work but I think it is a bit annoying for end users. They would need to `match` the enum all the time and ignore the variants that are not expected from that event type. I think I can work out some magic with downcasting, Any-types, etc and make it user-friendly. Let's see!

src/events.rs
orottier: I just realized we already use the name EventHandler:

```
pub(crate) struct EventHandler {
    pub event: Event,
    pub callback: Callback,
}
```

It is the callback combined with the type of event we are expecting for the callback. So I'm leaving it this way.

orottier:
Regarding your comment https://github.com/orottier/web-audio-api-rs/pull/244#issuecomment-1336205542
> [AudioWorklet.onprocessorerror](https://webaudio.github.io/web-audio-api/#dom-audioworkletnode-onprocessorerror) - this is probably the more tricky / needs discussion one now

We could implement this by putting `AudioProcessor::process` calls in a [panic handler](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html). But I'm not sure about the performance impact. It would be a good idea anyway because otherwise a rogue audio processor could crash the full render thread.

orottier:
> * [AudioWorklet.onprocessorerror](https://webaudio.github.io/web-audio-api/#dom-audioworkletnode-onprocessorerror) - this is probably the more tricky / needs discussion one now

I have an idea regarding https://doc.rust-lang.org/std/panic/fn.catch_unwind.html but we will have to measure the performance impact. Wrapping `process` calls in a panic handler is a good idea anyway, to prevent the render thread from crashing due to a single bad node

PR #245, opened 2023-01-08
Ability to cancel event handler

orottier:
FYI @b-ma  I wanted to share some quick work in progress.
There is a drawback for the API we came up with:
`pub fn onupdate<F: FnMut(AudioRenderCapacityEvent) + Send + 'static>(&self, value: Option<F>) {`
This makes it cumbersome to pass `None`, the compiler will complain:
```
43 |     cap.onupdate(None);
   |         ^^^^^^^^ cannot infer type of the type parameter `F` declared on the associated function `onupdate`
   |
help: consider specifying the generic argument
   |
43 |     cap.onupdate::<F>(None);
```
Which the user can fix by specifying:
`cap.onupdate(None::<fn(_)>);`

We should either make this very clear in the documentation, or come up with an alternative, such as:
```
pub fn onupdate_set<F: FnMut(AudioRenderCapacityEvent) + Send + 'static>(&self, value: F);
pub fn onupdate_clear<F: FnMut(AudioRenderCapacityEvent) + Send + 'static>(&self);
```

### DIFF
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -46,7 +46,7 @@ pub use panner::*;
 mod stereo_panner;
 pub use stereo_panner::*;
 mod waveshaper;
-use crate::events::Callback;
+use crate::events::EventHandler;
 pub use waveshaper::*;
 
 pub(crate) const TABLE_LENGTH_USIZE: usize = 8192;
@@ -359,13 +359,13 @@ pub trait AudioScheduledSourceNode: AudioNode {
     /// determined by stop() is reached. For an [`AudioBufferSourceNode`], the event is also
     /// dispatched because the duration has been reached or if the entire buffer has been played.
     ///
-    /// Calling this function multiple times will accumulate all event handlers. It is currently
-    /// not possible to remove an event handler.
+    /// Only a single event handler is active at any time. Calling this method multiple times will
+    /// override the previous event handler.
     fn onended<F: FnOnce() + Send + 'static>(&self, callback: F) {
         let callback = |_| callback();
         self.context().register_event_handler(
             crate::events::Event::Ended(self.registration().id()),
-            Callback::Once(Box::new(callback)),
+            EventHandler::Once(Box::new(callback)),
         );
     }
 }
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -362,6 +362,7 @@ pub trait AudioScheduledSourceNode: AudioNode {
     /// Calling this function multiple times will accumulate all event handlers. It is currently
     /// not possible to remove an event handler.
     fn onended<F: FnOnce() + Send + 'static>(&self, callback: F) {
+        let callback = |_| callback();
         self.context().register_event_handler(
             crate::events::Event::Ended(self.registration().id()),
             Callback::Once(Box::new(callback)),
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -4,7 +4,7 @@ use std::sync::atomic::{AtomicU32, AtomicUsize, Ordering};
 use std::sync::Arc;
 
 use crate::context::{AudioContextRegistration, ConcreteBaseAudioContext};
-use crate::events::Event;
+use crate::events::EventType;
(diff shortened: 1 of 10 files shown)


src/node/mod.rs
```diff
@@ -359,12 +359,17 @@ pub trait AudioScheduledSourceNode: AudioNode {
...
     /// dispatched because the duration has been reached or if the entire buffer has been played.
     ///
-    /// Calling this function multiple times will accumulate all event handlers. It is currently
-    /// not possible to remove an event handler.
-    fn onended<F: FnOnce() + Send + 'static>(&self, callback: F) {
-        self.context().register_event_handler(
+    /// Only a single event handler is active at any time. Calling this method multiple times will
+    /// override the previous event handler.
+    fn onended<F: FnOnce() + Send + 'static>(&self, value: Option<F>) {
+        let callback = value.map(|f| {
```
orottier, line 365: Notice I'm using a silly trick here and [here](https://github.com/orottier/web-audio-api-rs/pull/245/files#diff-d20ac8af5de1b3b89086e74e5f5881a19f0d73a14819c79546277cfbb9f1ecc9R176)
The user can supply a `FnOnce()` here which takes no arguments. Which is good because the `onended` event carries no payload.
For the AudioRenderCapacity, the user supplies `FnMut(AudioRenderCapacityEvent)` because there is a payload here.
Since all callbacks live in the `event_handlers` map of the `EventLoop`, I'm converting them both to take a payload and erase these differences here

b-ma:
Damn, that's a bit tricky indeed, I think the first solution (i.e. `cap.onupdate(None::<fn(_)>);`) is a bit weird / complicated in terms of API

What about:
```rs
set_onupdate(..);
clear_onupdate(...);
```
at least it would be kind of consistent with how we handled JS attributes in other places (e.g. `set_value`, `set_buffer`)?

b-ma:
Hey, just had a closer look and I think all that event stuff is in a really good shape now! 

I guess the last thing I would wait from an JS API point of view, is that all `EventHandler` receive an `Event` with (at least) a `type` attribute as argument, i.e.:

```rs
// just use raw string as JS so we don't have to expose any internal types
osc.set_onended(|e| e.type_ == "ended" );
context.set_onsinkchange(|e| e.type_ == "sinkchange");
capacity.set_onupdate(|e| e.type_ == "update"));
```
I agree this is maybe a bit idiomatic and redondant, but in JS it's kind of a contract I guess that anytime you register a callback you just get an event that has a `type` as argument, e.g.

```js
$someDomElement.addEventListener('click', e => console.log(e.type == 'click'));
````

Not a very strong opinion neither (because it's a bit idiomatic and redondant :) but if it's not too problematic to implement with the render capacity event, I guess it would be a nice to have. For the other one it seems quite simple, or am I missing something?

```rs
pub(crate) enum EventPayload {
    Ended { type_: String },
    SinkChange { type_: String },
    // ...
}
```
---

On a more internal perspective, I think we could maybe rename `EventPayload` to `Event` and `Event` to `EventDispatch`:

```rs
pub(crate) enum Event { 
    None,
    RenderCapacity(AudioRenderCapacityEvent),
}

pub(crate) struct EventDispatch {
    type_: EventType,
    event: Event,
}
```

Then we would be very aligned on the spec I think, with the `EventHandler` that is executed with an `Event` as argument (it's really no big deal neither as it is `pub(crate)` but maybe that would help a bit for reading and following the code).

And to be very (very) picky, maybe rename `EventType::SinkChanged` to `EventType::SinkChange`

orottier:
Thanks for having a look!
I will try your suggestion and give all handlers an `Event` argument. I will maybe look a bit silly but on the other hand it is more consistent, all handler will then take an argument. Let's see how it looks!

> On a more internal perspective, I think we could maybe rename EventPayload to Event and Event to EventDispatch
sure

> And a last one to be very (very) picky, maybe rename EventType::SinkChanged to EventType::SinkChange
yup

orottier:
Merging this to get a clean PR for the requested changes

PR #246, opened 2023-01-10
Switch to using easyfft instead of realfft

WalterSmuts:
This simplifies the logic, removing a bunch of boilerplate and performance should remain the same or improve.


orottier:
Hello, and thanks for the effort. I have been following your progress in [the realfft repo](https://github.com/HEnquist/realfft/pull/23) and I think your crate offers some nice ergonomical and safety benefits.
That said I don't have time to audit your crate, plus the code you have touched on our end is very light on tests, so I won't merge this anytime soon. I'm definitely keeping the PR open to have a good look later!

WalterSmuts:
Sure, feel free to take all the time you want. There are a bunch of tests that caught a bunch of silly mistakes, so there's definitely test coverage in this code. But there's no reason to rush. Can't say there's much to gain other than code being simplified. Perhaps removing some of the boilerplate will lead to other simplifications to be made in the future. As it stands however there shouldn't be much of a change in behaviour though so definitely not very time sensitive.

orottier:
Apologies. I have decided to stick to the well-known realfft library

PR #247, opened 2023-01-11
Implement `onprocessorerror` event system

orottier:
- [x] check performance impact (less than 1% regression for instruction count)
- [x] rebase onto #248 
- [ ] reason about UnwindSafe
- [x] add tests
- [ ] decide if the requirement about boxed closures is acceptable


src/render/graph.rs
```diff
@@ -377,10 +378,23 @@ impl Graph {
                 .iter_mut()
                 .for_each(|i| i.mix(count, interpretation));
 
-            // let the current node process
+            // let the current node process (catch any panics that may occur)
             let params = AudioParamValues::from(&*nodes);
             scope.node_id.set(*index);
-            let tail_time = node.process(params, scope);
+            let (success, tail_time) = {
```
orottier, line 384: Caveat emptor

src/node/mod.rs
```diff
@@ -321,6 +320,30 @@ pub trait AudioNode {
...
         self.channel_config().set_count(v)
     }
+
+    /// Register callback to run when an unhandled exception occurs in the audio processor.
+    ///
+    /// Note that once a unhandled exception is thrown, the processor will output silence throughout its lifetime.
+    ///
+    /// Only a single event handler is active at any time. Calling this method multiple times will
+    /// override the previous event handler.
+    fn set_onprocessorerror(&self, callback: Box<dyn FnOnce(ErrorEvent) + Send + 'static>) {
```
orottier, line 330: The AudioNode trait cannot take any generic arguments in its method. Compiler will complain `this trait cannot be made into an object`. Hence we need the user to box it already to erase the generic fn type

orottier:
Performance impact appears minimal, nice.
I have added a checklist with other TODOs in the PR description

b-ma:
Hey, completely forgot this one as I was thinking you still wanted to add things, do you want me to review it?

orottier:
We should be good to review now. I just wrote [f532735](https://github.com/orottier/web-audio-api-rs/pull/247/commits/f532735e0d30c973aa8b711aef7160f730e4628a) to remove the panic from the worklet example, but you could revert that to see the change in action.

src/node/mod.rs
b-ma: I just wonder if we just not generalize this API to all events then for consistency?

b-ma:
Hey, apart from my small question above, this looks all good to me too!

src/node/mod.rs
orottier: Tough choice, I dislike that we need to ask users to box the closure. It really is a lot nicer to just pass the closure in raw. To the user it will not be very clear why they need to box it, anyway
I think I will merge this version, but feel free to open the discussion again if you think the consistency is more important

PR #250, opened 2023-01-18
Fix cpal build input

b-ma:
This fix the `cpal::build_input` on RPi (and maybe more generally on Linux/ALSA, I don't know) where asking for the max sample rate just returned `~2 * 10^9` (which seems to much to be supported :)

Also a bit of refactoring to align the implementation on `build_output`


orottier:
Thanks. Looks very good to me. The microphone code is a bit neglected (and developed before enumerateDevices was introduced here).  I have created https://github.com/orottier/web-audio-api-rs/issues/251

PR #252, opened 2023-01-18
Clean panner API

b-ma:
Finally took the time to check the error I have seen trying to generate the bindings. So here we are:
- A bit here, with the `impl Default` for some options
- A bit on the bindings side to ignore the deprecated functions that are not implemented here

### DIFF
--- a/src/node/panner.rs
+++ b/src/node/panner.rs
@@ -125,7 +125,7 @@ impl Default for PannerOptions {
     }
 }
 
-/// Assert that the channel count is valid for the StereoPannerNode
+/// Assert that the channel count is valid for the PannerNode
 /// see <https://webaudio.github.io/web-audio-api/#audionode-channelcount-constraints>
 ///
 /// # Panics
@@ -140,7 +140,7 @@ fn assert_valid_channel_count(count: usize) {
     }
 }
 
-/// Assert that the channel count is valid for the StereoPannerNode
+/// Assert that the channel count is valid for the PannerNode
 /// see <https://webaudio.github.io/web-audio-api/#audionode-channelcountmode-constraints>
 ///
 /// # Panics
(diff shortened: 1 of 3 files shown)


src/node/panner.rs
```diff
@@ -103,10 +116,45 @@ impl Default for PannerOptions {
...
+            channel_config: ChannelConfigOptions {
+                count: 2,
+                count_mode: ChannelCountMode::ClampedMax,
+                interpretation: ChannelInterpretation::Speakers,
+            },
         }
     }
 }
 
+/// Assert that the channel count is valid for the StereoPannerNode
```
orottier, line 128:
```suggestion
/// Assert that the channel count is valid for the PannerNode
```

src/node/panner.rs
```diff
@@ -103,10 +116,45 @@ impl Default for PannerOptions {
...
+///
+#[track_caller]
+#[inline(always)]
+fn assert_valid_channel_count(count: usize) {
+    if count > 2 {
+        panic!("NotSupportedError: PannerNode channel count cannot be greater than two");
+    }
+}
+
+/// Assert that the channel count is valid for the StereoPannerNode
```
orottier, line 143:
```suggestion
/// Assert that the channel count is valid for the PannerNode
```

src/node/panner.rs
```diff
@@ -262,23 +310,35 @@ impl AudioNode for PannerNode {
         1
     }
 
-    fn set_channel_count(&self, v: usize) {
-        if v > 2 {
-            panic!("NotSupportedError: PannerNode channel count cannot be greater than two");
-        }
-        self.channel_config.set_count(v);
+    // same limitations as for the PannerNode
```
orottier, line 313:
```suggestion
    // same limitations as for the StereoPannerNode
```

orottier (commented):
Great fixes

b-ma:
Thanks for fixing the issues in the comments!

PR #254, opened 2023-01-18
Change AnalyserNode API, take mutable reference for time/freq data

### LINKED ISSUE #253: AnalyzerNode API
Hey,

Trying to wrap the Analyzer I just ran on the fact that  `get_float_frequency_data` and `get_float_time_domain_data` are both waiting for a `Vec` while it seems to me that everywhere else the spec declare a `Float32Array` we used a ` &mut [f32]`, cf. `AudioBuffer::copy_from_channel` for example.

Is there any reason for that ? I don't really see why we couldn't use the same API here

orottier:
It requires some unsafe magic but it does make the API a lot nicer.
Fixes #253

### DIFF
--- a/examples/mic_playback.rs
+++ b/examples/mic_playback.rs
@@ -229,7 +229,7 @@ fn poll_frequency_graph(
     height: u16,
 ) -> ! {
     let bin_count = analyser.frequency_bin_count();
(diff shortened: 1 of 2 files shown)


b-ma:
(whow `unsafe` code :) 

Looks good to me! At least the API is consistent, whatever we do from that point

PR #255, opened 2023-01-25
Add fast path for adding two upmixed mono audio buffers

### LINKED ISSUE #117: Optimizations for AudioRenderQuantum when all channels are equal

orottier:
fixes #117

It shows a slight performance improvement in the benchmarks for me. Let's see what the CI runner tells us

### DIFF
--- a/src/render/quantum.rs
+++ b/src/render/quantum.rs
@@ -562,6 +562,17 @@ impl AudioRenderQuantum {
             ChannelCountMode::ClampedMax => max_channels.min(count),
         };
 
+        // fast path if both buffers are upmixed mono signals
+        if interpretation == ChannelInterpretation::Speakers
+            && self.all_channels_identical()
+            && other.all_channels_identical()
+        {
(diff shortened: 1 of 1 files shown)


b-ma:
Seems to be a bit better too on my side. In any case, I really think this is a good thing to include as we will be able to use it in several nodes render too.

Issue #236, opened 2022-11-08
AnalyserNode should recover from NaN


orottier:
Conform discussion in https://github.com/WebAudio/web-audio-api/issues/2512 we should write a patch for the time smoothing to replace NaNs.
Chrome patch is https://github.com/chromium/chromium/commit/b81148bafdc8cd89a6e4b3c9cccc479eacde4a46

PR #256, opened 2023-01-25
Refactor analyser node

b-ma:
Hey,

This is quite a big PR regarding the AnalyserNode and what we discussed in #253:
- use a thread-safe buffering queue (`Arc<AtomicF32>`) to ship audio data from audio thread to control thread
- all computations (FFT) are now done in control thread
- implement missing methods
- follow spec according to recomputation of FFT if in same render quantum, etc.
- added a very simple example (`analyser`) to complete the quite complex `mic_playback` one

I'm not completely sure that what have done for thread safety (i.e. moving the node to a different thread as you have done in the `mic_playback` example) is the cleanest / simplest possible way, but it builds and seems to work quite well. But any case having your thoughts on this point would be nice.

Let me know what you think

### DIFF
--- a/src/analysis.rs
+++ b/src/analysis.rs
@@ -796,42 +796,28 @@ mod tests {
     fn test_ring_buffer_concurrency() {
         let analyser = Arc::new(Analyser::new());
         let ring_buffer = analyser.get_ring_buffer_clone();
-
         let num_loops = 10_000;
+        let (sender, receiver) = crossbeam_channel::bounded(1);
 
-        let _ = thread::spawn(move || {
+        thread::spawn(move || {
             let mut rng = rand::thread_rng();
-            let mut counter = 0;
+            sender.send(()).unwrap(); // signal ready
 
-            loop {
+            for _ in 0..num_loops {
                 let rand = rng.gen::<f32>();
                 let data = [rand; RENDER_QUANTUM_SIZE];
                 ring_buffer.write(&data);
-
-                counter += 1;
-
-                if counter == num_loops {
-                    break;
-                }
 
                 std::thread::sleep(std::time::Duration::from_nanos(30));
             }
         });
 
-        std::thread::sleep(std::time::Duration::from_millis(1));
+        // wait for thread to boot
+        receiver.recv().unwrap();
 
-        let mut counter = 0;
-
-        loop {
+        for _ in 0..num_loops {
             let mut read_buffer = [0.; RENDER_QUANTUM_SIZE];
             analyser.get_float_time_domain_data(&mut read_buffer);
-
-            counter += 1;
-
-            if counter == num_loops {
-                break;
-            }
-
             std::thread::sleep(std::time::Duration::from_nanos(25));
         }
     }
@@ -840,9 +826,11 @@ mod tests {
     fn test_thread_safety() {
         let analyser = Arc::new(RwLock::new(Analyser::new()));
 
-        thread::spawn(move || {
+        let handle = thread::spawn(move || {
             analyser.write().unwrap().set_fft_size(MIN_FFT_SIZE);
             assert_eq!(analyser.write().unwrap().fft_size(), MIN_FFT_SIZE);
         });
+
+        handle.join().unwrap();
     }
 }
--- a/src/analysis.rs
+++ b/src/analysis.rs
@@ -3,12 +3,12 @@
 //! These are used in the [`AnalyserNode`](crate::node::AnalyserNode)
 
 use std::f32::consts::PI;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use std::sync::atomic::{AtomicU32, AtomicUsize, Ordering};
 
 use realfft::{num_complex::Complex, RealFftPlanner};
 
-use crate::RENDER_QUANTUM_SIZE;
+use crate::{RENDER_QUANTUM_SIZE, AtomicF64};
 
 
 // @todo - modify AtomicF32 in `lib.rs` to expose Ordering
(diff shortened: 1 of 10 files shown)


b-ma:
Just added a patch for #236 too

src/node/analyser.rs
```diff
@@ -30,33 +31,62 @@ pub struct AnalyserOptions {
...
+/// - `cargo run --release --example mic_playback`
+///
 pub struct AnalyserNode {
     registration: AudioContextRegistration,
     channel_config: ChannelConfig,
-    fft_size: Arc<AtomicUsize>,
-    smoothing_time_constant: Arc<AtomicU32>,
-    sender: Sender<AnalyserRequest>,
+    // RwLock is needed to make the AnalyserNode API immutable
+    analyser: Arc<RwLock<Analyser>>,
```
orottier, line 89: The `Arc` is not necessary

src/analysis.rs
```diff
@@ -27,358 +23,832 @@ pub fn generate_blackman(size: usize) -> impl Iterator<Item = f32> {
...
+        );
+    }
+}
+
+// as the queue is composed of AtomicF32 having only 1 render quantum of room should be enough
+const RING_BUFFER_SIZE: usize = MAX_FFT_SIZE + RENDER_QUANTUM_SIZE;
+
+// single producer / single consumer ring buffer
+pub(crate) struct AnalyserRingBuffer {
+    buffer: Arc<Vec<AtomicF32>>,
```
orottier, line 90: This `Arc` is not necessary

orottier: alternatively, make it like this:
```
#[derive(Clone)]
pub(crate) struct AnalyserRingBuffer {
    buffer: Arc<[AtomicF32]>,
    write_index: Arc<AtomicUsize>,
}
```
Then you never need to wrap the thing inside an Arc, just clone them. I think it saves a level of indirection so it is more performant

src/analysis.rs
```diff
@@ -27,358 +23,832 @@ pub fn generate_blackman(size: usize) -> impl Iterator<Item = f32> {
...
+        // computed data is returned.
+        if current_time != self.last_fft_time {
+            self.compute_fft();
+            self.last_fft_time = current_time;
+        }
+
+        // [spec] Write the current frequency data into array. If array’s byte
+        // length is less than frequencyBinCount, the excess elements will be
+        // dropped. If array’s byte length is greater than the frequencyBinCount
+        let len = dst.len().min(frequency_bin_count);
```
orottier, line 370: `zip` will stop when the end of either buffer is reached, so I don't think `take` is necessary

src/analysis.rs
```diff
@@ -27,358 +23,832 @@ pub fn generate_blackman(size: usize) -> impl Iterator<Item = f32> {
...
```
orottier, line 396: `zip` will stop when the end of either buffer is reached, so I don't think `take` is necessary

orottier (commented):
Hey, I took a first look and I think this very well reflects the 'safe' version we have discussed. Really nice!

One important thing to notice is that the outputs between the current version and yours differ by a lot. Try running your analyser example on main and your branch and see what I mean. You also notice this in the mic_playback example. I wonder if maybe the current version is broken?

I left some small nitpicks that could further improve performance. I will also reply to your comments in the issue in a minute. 

In any case, please proceed with your adventures

src/analysis.rs
orottier: You can create an `Arc<[..]>` with `Arc::from(&vec_of_atomics[..]).` (you need an intermediate allocation that hopefully the compiler will skip)

b-ma: ok I did the alternative but kept the `Arc<Vec<AtomicF32>>`. I didn't manage to create the `array` from the `Vec`, with your snippet the build failed because it gives `Arc<&[AtomicF32]>` rather than `Arc<[AtomicF32]>.

src/analysis.rs
b-ma: actually the `last_fft_output` buffer is never resized, so we need the `take` here

b-ma:
> One important thing to notice is that the outputs between the current version and yours differ by a lot. Try running your analyser example on main and your branch and see what I mean. You also notice this in the mic_playback example. I wonder if maybe the current version is broken?

Hum weird indeed, I will have a look

b-ma:
Ok, I checked the values and they are for sure wrong in `main` branch as there is no way a sine can trigger a 20dB peak. So I just recreated the example in JS to check the values of current branch against Chrome and Firefox and it appears consistent (10th bin converging around -14dB). 

So I think we are all good on this side!

src/analysis.rs
```diff
@@ -27,358 +23,830 @@ pub fn generate_blackman(size: usize) -> impl Iterator<Item = f32> {
...
-    fft_planner: RealFftPlanner<f32>,
+    ring_buffer: AnalyserRingBuffer,
+    fft_size: usize,
+    smoothing_time_constant: f64,
+    min_decibels: f64,
+    max_decibels: f64,
+    // If not wrapped into Arc<Mutex<T>>, compiler complains about thread safety:
+    // > `(dyn rustfft::avx::avx_planner::AvxPlannerInternalAPI<f32> + 'static)`
+    // > cannot be shared between threads safely
+    // But the Mutex is ok here as `Analyser` lives outside the audio thread.
```
orottier, line 164: Are you sure this is necessary? It works on my hardware without the Arc+Mutex.
It would also surprise me, Mutex does not magically make non-Send types Send. Being not-Send means you cannot leave the thread you were created on. Arc+Mutex specifically allows one to transfer objects from one thread to another

src/analysis.rs
orottier: It was easier than I thought: 1c29c2f8ca820ab46442222935b6596c9bd635ef
Showing a nice bit of extra performance, before:

```
bench_analyser_node
  Instructions:            41181322 (-18.52477%)
  L1 Accesses:             58665001 (-17.67864%)
  L2 Accesses:               154232 (-63.80796%)
  RAM Accesses:               84974 (-1.944403%)
  Estimated Cycles:        62410251 (-18.34027%)
```

after:
```
bench_analyser_node
  Instructions:            39385732 (-22.07732%)
  L1 Accesses:             56186630 (-21.15673%)
  L2 Accesses:               158969 (-62.67475%)
  RAM Accesses:               84981 (-1.936325%)
  Estimated Cycles:        59955810 (-21.55078%)
```

src/analysis.rs
orottier: good point

src/analysis.rs
orottier: I took the liberty to remove them. Let me know if building fails on your hardware

orottier: Hah, it fails on CI. The Mutex is required to make the thing `Sync`

orottier:
So to summarize, this PR
- fixes a giant bug in calculation of the actual frequency data
- shows a 20% instruction count reduction on the render thread
- fixes a weird construct where the control thread would need to wait a render quantum before the FFT data is shipped
- adds a lot of tests

I could not be more happier

PR #259, opened 2023-02-03
Fix: review some values compared to original benches

b-ma:


orottier:
Nice catch. The performance impact on the benchmarks seems minimal, interestingly.

PR #260, opened 2023-02-07
Offline context rendering

b-ma:
Hey, 

Here is small fix / refactor for the offline context rendering, mainly:
- Allocate the final `AudioBuffer` all at once instead of make it grow on each render quantum
- Perform rendering in a dedicated thread (which is required by the spec)

Does not make huge difference on benchmarks, but I can see an improvement of ~10-15% in the baseline, for the other ones it fall into noise I guess. But in any case it seems more logical to do like this.

Let me know what you think!

### DIFF
--- a/examples/benchmarks.rs
+++ b/examples/benchmarks.rs
@@ -97,7 +97,7 @@ fn main() {
     {
         let name = "Baseline (silence)";
 
-        let context = OfflineAudioContext::new(2, DURATION * sample_rate as usize, sample_rate);
+        let context = OfflineAudioContext::new(1, DURATION * sample_rate as usize, sample_rate);
 
         benchmark(&mut stdout, name, context, &mut results);
     }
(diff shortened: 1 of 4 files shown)


orottier:
Hey, cool fix. Definitely feels more natural this way.

I'm not sure why we should perform the rendering off thread though.
That will happen automatically once we implement something like #177 
In the meantime we are stuck with `start_rendering_sync` which is non-compliant anyway. Spawning a dedicated thread only introduces overhead and no actual benefit for the user.
The only benefit I could imagine is that the control thread does not suffer when the render thread panics. But that will be handled in #247  and since you call `handle.join().unwrap()` the panic still propagates to the control thread..

b-ma:
Yup, I was probably more thinking about the JS side where doing this in a new thread is nice to not block the main thread. But I agree with you as the method is not spec compliant and is  obviously marked as `sync` this is probably much more logical to keep thing simple here.

PR #261, opened 2023-02-07
Increase benchmark timing precision (from millis to micros)

orottier:
Some tests only take a few millis per iteration so the precision would be really low


b-ma:
All good to me!

PR #262, opened 2023-02-13
refactor: replace for loop with iter

b-ma:
Hey,

A simple (and quite unexpected) perf improvement, just replacing a for loop with iter

## before

```
- 5       | Simple source test with resampling (Mono)                                | 77            | 1558.4x               | 120
- 6       | Simple source test with resampling (Stereo)                              | 102           | 1176.5x               | 120
- 7       | Simple source test with resampling (Stereo and positionnal)              | 148           | 810.8x                | 120
````

## after

```
- 5       | Simple source test with resampling (Mono)                                | 63.559        | 1888.0x               | 120
- 6       | Simple source test with resampling (Stereo)                              | 85.878        | 1397.3x               | 120
- 7       | Simple source test with resampling (Stereo and positionnal)              | 134.003       | 895.5x                | 120
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ vecmath = "1.0"
 
 [dev-dependencies]
 alloc_counter = "0.0.4"
+criterion = "0.4.0"
 env_logger = "0.10"
 iai = "0.1.1"
 rand = "0.8"
--- a/benches/my_benchmark.rs
+++ b/benches/my_benchmark.rs
@@ -1,4 +1,6 @@
-use iai::black_box;
+// use iai::black_box;
+
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
 
 use web_audio_api::context::BaseAudioContext;
 use web_audio_api::context::OfflineAudioContext;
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by orottier** [comment]
Hey, I cannot reproduce the performance impact. Are you sure about your numbers?
Also your main branch may be out of date because it does not show the sub-millisecond numbers.

Alternatively, you could add a benchmark for the iai benchmark suite to get the absolute numbers.
We really ought to start with #188 ... using criterion for the local benchmarks will reduce the noise

**Comment by b-ma** [comment]
Hey, yup I was out of date indeed. Just merged everything, re-run the benches a few time, and yup on my side it's still quite better, particularly visible on `Simple source test with resampling (Mono)` e.g.:

**before**

```
- 5       | Simple source test with resampling (Mono)                                | 72.425        | 1656.9x               | 120
```

**after**

```
- 5       | Simple source test with resampling (Mono)                                | 63.129        | 1900.9x               | 120
```

orottier:
Interesting. I will try to put this specific case in criterion and measure again

b-ma:
Hey, just tested locally compared to main... and indeed there are no differences... so weird...

Maybe let's just keep the `criterion` stuff instead of `iai`, so at least we can use it locally?

orottier:
Hey sorry I have left this in a hurry. I'm actually looking to support both. Iai on CI and local criterion with the 'official' benchmarks. Will take a bit of time though

b-ma:
oh ok, sounds good. Let's close that then this is too blurry

PR #263, opened 2023-03-22
MediaStream spec compliance

orottier:


orottier:
@b-ma do you agree to releasing the current version as v1 release candidate? I will then publish an announcement on either rust audio discord / rust subreddit / users.rust-lang.org

b-ma:
Heiho, yup at some point I guess we must consider that is kind of stable :)

I guess my only concern is with the MediaStream API, what do you think? should we consider it non breaking? enough for v1 and we will see later (i.e. V2) ? that's not completely clear to me (and I don't have any strong opinion, as (as a JS guy) I am kind of ok with breaking changes :)

orottier:
To be continued.. Here's the direction I'm taking for now

src/media/decoding.rs
```diff
@@ -49,41 +49,7 @@ impl<R: Read + Send + Sync> symphonia::core::io::MediaSource for MediaInput<R> {
...
-/// // build a decoded audio stream the decoder
-/// let file = std::fs::File::open("samples/major-scale.ogg").unwrap();
-/// let stream = MediaDecoder::try_new(file).unwrap();
-/// // pipe the media stream into the web audio graph
-/// let context = AudioContext::default();
-/// let node = context.create_media_stream_source(stream);
-/// node.connect(&context.destination());
-/// ```
-///
 pub struct MediaDecoder {
```
b-ma, line 53: I saw this commit message: "Remove MediaDecoder from public API", shouldn't it be `pub(crate)` then ?

src/node/media_stream_destination.rs
```diff
@@ -96,15 +102,13 @@ impl MediaStreamAudioDestinationNode {
...
+    /// itself
     ///
-    /// Note that while you can call this function multiple times and poll all iterators concurrently,
+    /// Note that while you can call this function multiple times and poll all streams concurrently,
     /// this could lead to unexpected behavior as the buffers will only be offered once.
-    pub fn stream(&self) -> impl MediaStream {
-        AudioDestinationNodeStream {
-            receiver: self.receiver.clone(),
-        }
+    pub fn stream(&self) -> &MediaStream {
```
b-ma, line 110: "as the buffers will only be offered once." is this still true ? it seemed to me that `MediaStreamTrackIter.position` was there to prevent that, but maybe I misunderstood something

src/media_streams/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+
+struct MediaStreamTrackInner {
+    data: ArcSwap<FallibleBuffer>,
+    position: AtomicU64,
+    ended: AtomicBool,
+    provider: Mutex<Box<dyn Iterator<Item = FallibleBuffer> + Send + Sync + 'static>>,
+}
+
+impl MediaStreamTrack {
+    pub fn lazy<T: IntoIterator<Item = FallibleBuffer>>(iter: T) -> Self
```
b-ma, line 43: Why `lazy` rather than `new`?

examples/toy_webrtc.rs
```diff
@@ -160,7 +161,8 @@ fn run_client() -> std::io::Result<()> {
         sample_rate: context.sample_rate(),
         byte_buf: vec![0; 512],
     };
-    let stream_in = context.create_media_stream_source(stream);
+    let track = MediaStreamTrack::lazy(stream);
+    let stream_in = context.create_media_stream_track_source(&track);
```
b-ma, line 165: looks cool that we can pass a reference now

b-ma:
Hey, I just add a very rapid review but it looks really nice so far!
The new mod layout is much more clean I think

src/media/decoding.rs
orottier: The mod is private so this is not strictly necessary.  I will add it for clarity though!

src/media_streams/mod.rs
orottier: I will write some extensive documentation for this module. I'm preparing for two kinds of streams:
- lazy: these have no internal clock and only progress when polled. 
- forced: driven by an external clock mechanism

Streaming audio from a mic or webrtc will be a forced stream. They possibly need buffering to synchronise with the web audio context clock. 
Streaming audio from a disk or generating samples on the fly could benefit from the lazy stream as they will precisely follow the web audio context time progression

To be continued!

orottier: Maybe better names would be `new` for forced streams and `from_iter` for lazy. I expect lazy streams to be more niche. 

Both are obviously for advanced users only. Producing and consuming streams manually is only necessary when microphone and MediaElement input or speaker output is not sufficient for an application.

orottier:
The work on MediaStream opens the possibility to bring the MediaRecorder API into our library

b-ma:
> The work on MediaStream opens the possibility to bring the MediaRecorder API into our library

Nice :)

orottier:
I believe this could be merged now. Stay tuned for incremental improvements in other PRs.
I have decided not to put this in a v1-release-candidate yet so this will just land in the next 0.* release

src/media_devices/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+///
+/// // connect the node directly to the destination node (speakers)
+/// background.connect(&context.destination());
+///
+/// // enjoy listening
+/// std::thread::sleep(std::time::Duration::from_secs(4));
+/// ```
+// TODO, return Promise? How to provide constraints?
+pub fn get_user_media() -> MediaStream {
+    crate::io::build_input(AudioContextOptions::default())
```
b-ma, line 123: Don't really know what would be the proper API, but being able to pass the `deviceId` somehow would be very important (e.g. to work to with Jack or any external sound cards).

I could just re-wrap the whole thing on the Node.js side, but as everything is now `pub(crate)` I will be stuck

b-ma: maybe something like this (a bit verbose...):

```rs
#[non_exhaustive]
struct MediaStreamConstraints {
    audio: MediaTrackConstraintsOption
}

#derive(Default)
enum MediaTrackConstraintsOption {
   False: boolean,
   #[default] // not spec compliant, but makes sens in our case I guess
   True: boolean,
   MediaTrackConstraints: MediaTrackConstraints
}

#[non_exhaustive]
struct MediaTrackConstraints {
      deviceId: string
}
```

Then in the general case, we could just:
```rs
let stream = media_devices.get_user_media(Default::default())
```

b-ma:
Except my last comment on the input selection (that I really need these days), it seems very nice to me!

orottier:
Thanks. I will merge this and work on MediaStreamConstraints. Your suggestions are very helpful

Issue #264, opened 2023-03-29
MediaStream spec compliance


orottier:
- https://developer.mozilla.org/en-US/docs/Web/API/MediaStream
- https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack

Reconsider current MediaStream trait and how it interacts with MediaStreamAudioSourceNode, MediaStreamAudioDestinationNode, and MediaDecoder

Speaking of MediaDecoder, I think we should drop it from the public API because we have fixed #120

PR #267, opened 2023-05-02
Add inputs in enumerate_devices

b-ma:
Hey,

I hope you are well, quite a small change to add audio inputs into the `enumerate_devices` list

Cheers,

### DIFF
--- a/examples/sink_id.rs
+++ b/examples/sink_id.rs
@@ -1,6 +1,6 @@
 use web_audio_api::context::{AudioContext, AudioContextOptions, BaseAudioContext};
-use web_audio_api::enumerate_devices;
 use web_audio_api::node::{AudioNode, AudioScheduledSourceNode};
+use web_audio_api::{enumerate_devices, MediaDeviceInfo, MediaDeviceInfoKind};
 
 fn ask_sink_id() -> String {
     println!("Enter the output 'device_id' and press <Enter>");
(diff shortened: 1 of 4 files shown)


orottier:
Hey, thanks. All is well! I actually picked up the work from #264 again but it's a bit of a struggle. I hope to push a work in progress this week. I have read https://w3c.github.io/mediacapture-main for inspiration. It's a nice spec which we can definitely partly integrate in our lib.

What do you need this PR actually for? I will merge it surely.

I'm also contemplating to rewrite `Microphone::new` into `MediaDevices.getUserMedia`.  Do you think we need a `media_devices` mod where we can also put `enumerate_devices` ?

b-ma:
> I actually picked up the work from https://github.com/orottier/web-audio-api-rs/issues/264 again but it's a bit of a struggle. I hope to push a work in progress this week. I have read https://w3c.github.io/mediacapture-main for inspiration. It's a nice spec which we can definitely partly integrate in our lib.

Cool, I have read it too it's quite nice indeed, a bit busy these times but let me know if I can help

> What do you need this PR actually for? I will merge it surely.

We have made some tests to use Jack audio connection kit in raspberry pi which seems to greatly improve the latency and performances, but I don't see how to make it system default, thus I'd like to monkey patch the `AudioContext` and `Microphone` constructor to use jack if it is found in the device list and fallback to alsa if not present.

> I'm also contemplating to rewrite Microphone::new into MediaDevices.getUserMedia. Do you think we need a media_devices mod where we can also put enumerate_devices ?

+1 , I already started to do that in the node bindings, it seems quite clean to me

PR #268, opened 2023-05-02
Remove termion dependency from main crate

### LINKED ISSUE #266: termion does not support windows
This dependency causes many cargo subcommands to fail on windows.

orottier:
Fixes #266


tasogare3710:
directly cause of the error is that the `sys` module in termion is not included in windows build because it does not support windows. Thus, if any action triggers source code analysis, it will find the unknown name `sys` and output an error.

orottier:
Thanks for the info @tasogare3710 .
Could you confirm the project now builds on windows?

PR #270, opened 2023-05-07
MediaStreamConstraints for getUserMedia

orottier:
Seems to work on my device, however some caveats 
1. microphone input is not buffered so some frames are lost (about a single chunk of 128 frames every second on my device). This introduces some crackling. You can see this when you run `RUST_LOG=debug cargo run --release --example microphone`
2. our cubeb input implementation only supports stereo microphone input. Not sure how I can fix that.. The whole cubeb implementation is experimental anyway #187 
3. when selecting a non-default sink-id, closing the media track does not actually suspend the microphone. I suspect this is a `cpal` bug https://github.com/RustAudio/cpal/issues/771

1 and 2 are not related to this PR by the way.
In any case I think we can merge this despite the caveats.
Let me know how it works for you!

### DIFF
--- a/src/media_devices/mod.rs
+++ b/src/media_devices/mod.rs
@@ -4,7 +4,7 @@
 //!
 //! <https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices>
 
-use crate::context::AudioContextOptions;
+use crate::context::{AudioContextLatencyCategory, AudioContextOptions};
 use crate::media_streams::MediaStream;
 
 /// List the available media output devices, such as speakers, headsets, loopbacks, etc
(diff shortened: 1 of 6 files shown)


src/media_devices/mod.rs
```diff
@@ -89,6 +89,51 @@ impl MediaDeviceInfo {
     }
 }
 
+/// Dictionary used to instruct what sort of tracks to include in the [`MediaStream`] returned by
+/// [`get_user_media`]
```
orottier, line 93: I found this to be the most friendly way of modeling the constraints, but let me know if you think otherwise

b-ma:
Cool, the API looks ok to me

Just tested on my side and, a bit hard to tell, but point 1 doesn't seem to be an issue on my machine. Points 2 and 3 behave the same.

One last think I'm just thinking about, maybe we could just rename `get_user_media` to `get_user_media_sync` (just as `start_rendering_sync` and `decode_audio_data_sync`)?

PR #271, opened 2023-05-07
Feature/v1.0 rc

orottier:
Let's have another go at it.
It would probably be best if you could update the node bindings to the latest version, before publishing the rc @b-ma. No hurries though, I can wait


b-ma:
Yup sure, my plan is to work on that next week, I'll let you know!

PR #274, opened 2023-05-10
media recorder

orottier:
This one was a lot of fun.
Only Wav encoding for now.
Some of the IDL still needs implementing but I will address that in a later issue

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ repository = "https://github.com/orottier/web-audio-api-rs"
 keywords = ["web-audio-api", "audio", "sound", "dsp"]
 license = "MIT"
 categories = ["multimedia::audio"]
-exclude = ["/samples", "/snapshots", "/.github"]
+exclude = ["/samples", "/showcase", "/.github"]
 
 [dependencies]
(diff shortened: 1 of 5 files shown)


b-ma:
Really nice!

Launching the example, I just found that the `onstop` callback is not called on my side

b-ma:
Hey, I retested and seems to work better. Maybe just one thing I don't really understand, I just hooked the on_stop callback to add a log:

```rs
recorder.set_onstop(move |event| {
    eprintln!(
        "on stop event: timecode {:?}",
        event.type_,
    );

    let _ = send.send(());
});
```

and the output is this (with en end event triggered before the last data available event):

```
timecode 0.000000, data size 131140
timecode 0.338635, data size 132096
timecode 0.682546, data size 132096
timecode 1.026551, data size 132096
timecode 1.370504, data size 132096
on stop event: "StopEvent"
timecode 1.714490, data size 109568
```

I checked the created file and it looks fine, but this print order is a bit confusing, do you think is this ok?

orottier:
Interesting. This is indeed a potential footgun. I will have a look

PR #275, opened 2023-05-11
Add buffer upmixing rules for 8-channel outputs

orottier:
These are not specified in the spec so it was a bit of guessing

### DIFF
--- a/src/context/concrete_base.rs
+++ b/src/context/concrete_base.rs
@@ -152,11 +152,18 @@ impl ConcreteBaseAudioContext {
             inner: Arc::new(base_inner),
         };
 
+        // Online AudioContext should start with stereo channels by default
+        let initial_channel_count = if offline {
+            max_channel_count
+        } else {
+            2.min(max_channel_count)
+        };
+
         let (listener_params, destination_channel_config) = {
             // Register magical nodes. We should not store the nodes inside our context since that
             // will create a cyclic reference, but we can reconstruct a new instance on the fly
             // when requested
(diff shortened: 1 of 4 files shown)


b-ma:
Yup I don't know if we should support that, `ChannelInterpretation::Explicit` is made for such issue I guess?
But in the other hand this seems to be quite handy... Probably we should add a comment about this being non specified

orottier:
I believe `Explicit` mode is for custom audio setups such as at venues where there is no automatic up/down mixing desired because the speaker layout may not adhere to L, R, C, LFE, SL, SR etc.

However, you may be right we should not go in this direction as it is unspecified.

I found the corresponding Firefox code at https://searchfox.org/mozilla-central/source/dom/media/AudioChannelFormat.h#44 and they also seem to dismiss these speaker layouts.

Maybe a better solution would be to simply ignore these speaker layouts when polling the audio host. I will try to cook up something new.

b-ma:
Hey, I just realized that might be the issue : https://github.com/ircam-ismm/node-web-audio-api/issues/20#issuecomment-1554550037, then by default it's always stereo which should work everywhere.

b-ma:
Cool! Should we keep the upmixing rules to height channels then?

orottier:
> Should we keep the upmixing rules to height channels then?

Yeah I think that is better.
Because we still give users to option to set their destination to the max channel count (of possibly 8). And it will panic otherwise.
We can't limit the destination max channel count to 5, because that will hinder users that use a `Explicit` speaker channel layout.

PR #276, opened 2023-05-22
Fix issue with AudioBufferSourceNode playback rate and detune

b-ma:
This fix an issue with `AudioBufferSourceNode::playback_rate` and `AudioBufferSourceNode::detune` that were wrongly going into fast track in some cases and were therefore not applied.

### DIFF
--- a/examples/audio_buffer_source_pitching.rs
+++ b/examples/audio_buffer_source_pitching.rs
@@ -1,0 +1,51 @@
+use std::f32::consts::PI;
+
+use web_audio_api::context::{AudioContext, BaseAudioContext};
+use web_audio_api::node::{AudioNode, AudioScheduledSourceNode};
+
+fn main() {
+    let context = AudioContext::default();
+
+    // create a 1 second buffer filled with a sine at 200Hz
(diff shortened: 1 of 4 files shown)


orottier:
Thanks!

bradleat:
I think I am still running into this issue as of 0.42.0

What I am doing is this:

```
    head_source_node
      .playback_rate()
      .set_value_at_time(starting_playback_rate, grid_start_time);
    head_source_node
      .playback_rate()
      .linear_ramp_to_value_at_time(target_playback_rate, new_bpm_region_start_time);

    head_source_node
      .detune()
      .set_value_at_time(starting_detune, grid_start_time + 0.01);
    head_source_node
      .detune()
      .linear_ramp_to_value_at_time(target_detune, new_bpm_region_start_time + 0.01);
      ```
      
      
As soon as I take away the detunes (which were calculated to reverse the effect of the playback rate changes), the playback rate takes hold. However, with the detunes it seems that it has affected the playback_rate.

(This is with and without the `+ 0.01` term. )

bradleat:
I could be misunderstanding the intent here. Is detuning intended to change playback rate?

Issue #281, opened 2023-06-02
Issue with mic input on linux (raspberry-pi)


b-ma:
Hey,

I made some tests with the microphone on the Raspberry Pi with JACK backend and the fact is that we have an issue. For a reason I'm not sure how to explain, after a certain amount of time (from few seconds to few minutes depending of the Jack server block size) the mic audio stream just outputs zeros...

I tried to test the underlying libraries ([rust-jack](https://github.com/RustAudio/rust-jack) and [cpal](https://github.com/RustAudio/cpal)) and their example seem to work quite well (...still need to test for a longer time though).

But looking at the `cpal` example code (https://github.com/RustAudio/cpal/blob/master/examples/feedback.rs), I saw that they manually add some latency. I think it is to take into account that their API is not really duplex, which I guess implies that it's not possible to make sure the input is called before the output (?)
I tested with a latency of 3ms (around 1 block size) and it seems to work quite fine. So I wonder if we shouldn't add some double-buffering strategy to the input stream (maybe just for the Jack backend, it doesn't seems to be an issue on mac, and on other linux backend the buffer size is so large that it's not possible to do anything real-time anyway...)

What do you think? I can have a shot on that (I don't know if you can access some linux machine) see where it goes but I'm not completely sure where I should start?

orottier:
Interesting. In theory our mic implementation should not fail catastrophically when the input callback lags behind the output.
Could you run it again with more logging? `RUST_LOG=debug cargo run --release --example microphone`
It should print a message when it drops buffers.

The relevant code part is 
https://github.com/orottier/web-audio-api-rs/blob/5c73c3a0411aff3e505155c44baea577fd9f05bd/src/io/cpal.rs#L254 we set up a channel with max 3 items in flight of 128 samples
https://github.com/orottier/web-audio-api-rs/blob/5c73c3a0411aff3e505155c44baea577fd9f05bd/src/io/microphone.rs#L102 when the input callback runs too fast, we drop the new frames
https://github.com/orottier/web-audio-api-rs/blob/5c73c3a0411aff3e505155c44baea577fd9f05bd/src/io/microphone.rs#L50 when the output callback runs too fast, we emit a chunk of silence, but after that we should pick up the stream again.

I now notice the log statement at https://github.com/orottier/web-audio-api-rs/blob/5c73c3a0411aff3e505155c44baea577fd9f05bd/src/io/microphone.rs#L50 is commented out, could you put it back when you run the example with more logging?

[...]

PR #283, opened 2023-06-08
Fix cpal jack host

b-ma:
Ok, I think I got the problem for #281! Actually the input and output streams were not proper jack streams but some weird alsa inside jack stuff (I don't really understand...). It seems far more stable now

I also refactored a bit the `MicrophoneRenderer`, looked a bit more logical to create the `AudioBuffer` in the stream iterator rather than in the input callback.

### DIFF
--- a/src/io/cpal.rs
+++ b/src/io/cpal.rs
@@ -66,32 +66,39 @@ mod private {
 }
 use private::ThreadSafeClosableStream;
 
-fn get_device_label(sink_id: &str) -> String {
-    if !sink_id.is_empty() {
-        let device = CpalBackend::enumerate_devices_sync()
-            .into_iter()
-            .find(|e| e.device_id() == sink_id)
-            .unwrap();
-
-        return device.label().to_string();
-    }
-
-    "default".to_string()
-}
-
-// label is only used when the cpal-jack feature is activated
-#[allow(unused_variables)]
-fn get_host(label: &str) -> cpal::Host {
+fn get_host() -> cpal::Host {
     #[cfg(feature = "cpal-jack")]
-    if label == "jack" {
-        return cpal::host_from_id(cpal::available_hosts()
+    {
+        // seems to be always Some when jack is installed,
+        // even if it's not running
+        let some_jack_id = cpal::available_hosts()
             .into_iter()
-            .find(|id| *id == cpal::HostId::Jack)
-            .expect(
-                "make sure --features jack is specified. only works on OSes where jack is available",
-            )).expect("jack host unavailable");
+            .find(|id| *id == cpal::HostId::Jack);
+
+        if some_jack_id.is_some() {
+            let jack_id = some_jack_id.unwrap();
+            let jack_host = cpal::host_from_id(jack_id).unwrap();
+
+            // if jack is not running, the host can't access devices
+            // fallback to default host
+            return match jack_host.devices() {
+                Ok(devices) => {
+                    // no jack devices found, jack is not running
+                    if devices.count() == 0 {
+                        log::info!("No jack devices found, fallback to default");
+                        cpal::default_host()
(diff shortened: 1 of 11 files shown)


src/io/cpal.rs
```diff
@@ -66,6 +65,35 @@ mod private {
...
+
+        return device.label().to_string();
+    }
+
+    "default".to_string()
+}
+
+// label is only used when the cpal-jack feature is activated
+#[allow(unused_variables)]
+fn get_host(label: &str) -> cpal::Host {
```
orottier, line 83: You should probably call `get_host` as well in `enumerate_devices_sync` down below no, instead of `cpal::default_host`?
Otherwise you ask Alsa (when it is default host) for the devices, and then construct them via Jack, no?

And I'm not too sure why get_host should depend on the label.
Why not choose jack when it is available, and default_host otherwise?

Like so:
```rust
fn get_host() -> cpal::Host {
    #[cfg(feature = "cpal-jack")]
    {
        if let Some(jack_available) = cpal::available_hosts()
            .into_iter()
            .find(|id| *id == cpal::HostId::Jack) {
                 return cpal::host_from_id(jack_available).unwrap();
            };
    }
    cpal::default_host()
}

In that case `get_device_label` is not necessary at all.

File: src/io/cpal.rs [hunk 2]
```diff
@@ -205,14 +236,17 @@ impl AudioBackendManager for CpalBackend {
         }
     }
 
-    fn build_input(options: AudioContextOptions) -> (Self, Receiver<AudioBuffer>)
+    fn build_input(options: AudioContextOptions) -> (Self, Receiver<Vec<f32>>)
```
orottier, line 239: I need to think a bit more about this change, I don't really like flying `Vec` all over the place. I'd rather see we encode the important information alongside it asap (channels, samplerate). The code is a bit messy in `microphone.rs` I notice, I will try if there is another way to make this easier to follow

orottier (commented):
Hey, great to hear you were able to dissect the issue and fix it! 20 hours of running time gives me a pretty confident feeling about the setup.

I don't have a Jack-capable machine at hand so it is a bit hard for me to check if the methods you have added are fully necessary. Please have a look at my comments, I think we can get away with something easier, but maybe I'm not seeing the real issue at hand.

src/io/cpal.rs
b-ma: Yup you are right, I didn't try to see what `enumerate_device_sync` using jack host would give us actually, I will check that. Maybe it can simplify the setup yes (I must confess the current version is mostly "try this, try that, works, don't touch!" :)

src/io/cpal.rs
b-ma: I think the code is a bit more readable like that, but no strong opinion and maybe this is a bit less efficient too. 
I will revert to the previous version, let's keep the mic refactoring for a dedicated issue / PR.

b-ma:
> I don't have a Jack-capable machine at hand so it is a bit hard for me to check if the methods you have added are fully necessary.

Yup maybe it would be nice to ask `cpal` to expose jack in MacOS too (no real reason why they didn't do that at first sight), as it would be far more fluid for development / testing than doing things on a RPi. I tried to hack something to make a PR but miserably failed...

b-ma:
Ok, looks better now I think, the `get_host` function is a bit more complicated than your snippet to gracefully fallback to default when jack is installed but not running but that's all.

One thing to note is that `enumerate_devices_sync` with jack host (when the jack server is running) only return jack devices (which I guess is a decent default behavior) and they all appear in both input and output list (but this is not blocking and is rather a `cpal` issue I think).

Let me know what you think!

orottier:
This is great, looks good to me. I'm applying two cosmetic fixes and will then merge

PR #284, opened 2023-06-09
showcase: add round trip latency tester

b-ma:
Hey, just a simple tool to estimate full round trip latency

Let me know if you think it worth the merge!


orottier:
It's a bit niche but definitely nice to have listed!
Since you are not using any external dependencies there is actually no reason to put this in the `showcase/` folder. It can just be a regular `examples/` file. This prevents the expensive recompilation of the full project. Shall we move it there?

By the way, I don't have a separate mic-in but still ran the example. It showed a latency of 0 seconds which is impossible :) maybe change the display to "unable to perform measurement"?
When I crank up the volume it actually starts measuring the real thing. Very cool! Average latency on my Apple is 0.053s

b-ma:
Yup righ,t I moved everything to example and displayed some error message if estimated latency is 0

> Average latency on my Apple is 0.053s

Damn that's a lot, on RPi with Jack I'm around 13ms, I will try to find some sound card to test this on my mac too

orottier:
Looks great!

PR #287, opened 2023-06-17
waveshaper: Avoid allocations while resampling

### LINKED ISSUE #143: Use rubato realtime features
We are fixed at 0.10 currently
0.11 has some feature to be more 'realtime safe', it contains breaking changes
0.12 enables more SIMD-stuff but requires a very recent rustc

We should probably take our time to assess our usage of this lib and how we can utilize it better. There's many features and types of resamplers. We also ship our built-in linear interpolator which is maybe unneeded?

uklotzde:
Fixes #143.

### DIFF
--- a/src/node/waveshaper.rs
+++ b/src/node/waveshaper.rs
@@ -10,6 +10,7 @@ use rubato::{FftFixedInOut, Resampler as _};
 use crate::{
     context::{AudioContextRegistration, BaseAudioContext},
     render::{AudioParamValues, AudioProcessor, AudioRenderQuantum, RenderScope},
+    RENDER_QUANTUM_SIZE,
 };
 
 use super::{AudioNode, ChannelConfig, ChannelConfigOptions};
@@ -256,8 +257,6 @@ struct RendererConfig {
     receiver: Receiver<CurveMessage>,
 }
 
-const DEFAULT_CHUNK_SIZE: usize = 128;
-
 #[derive(Debug, Clone, PartialEq, Eq)]
 struct ResamplerConfig {
     channels: usize,
@@ -268,7 +267,7 @@ struct ResamplerConfig {
 
 impl ResamplerConfig {
     fn upsample_x2(channels: usize, sample_rate: usize) -> Self {
-        let chunk_size_in = DEFAULT_CHUNK_SIZE * 2;
+        let chunk_size_in = RENDER_QUANTUM_SIZE * 2;
         let sample_rate_in = sample_rate;
         let sample_rate_out = sample_rate * 2;
         Self {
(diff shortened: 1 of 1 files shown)


src/node/waveshaper.rs
```diff
@@ -626,3 +720,23 @@ mod tests {
         assert_float_eq!(channel[..], expected[..], abs_all <= 0.);
     }
 }
+
+#[inline]
+fn apply_curve(curve: &[f32], input: f32) -> f32 {
```
orottier, line 725: Could you move this method above the `tests`?

src/node/waveshaper.rs
```diff
@@ -256,6 +256,122 @@ struct RendererConfig {
     receiver: Receiver<CurveMessage>,
 }
 
+const DEFAULT_CHUNK_SIZE: usize = 128;
```
orottier, line 259: If you have chosen this constant to correspond to the default render quantum size, I'd rather have you import `crate::RENDER_QUANTUM_SIZE`. It will make the migration easier for #217

src/node/waveshaper.rs
```diff
@@ -308,82 +424,85 @@ impl AudioProcessor for WaveShaperRenderer {
             match self.oversample.load(Ordering::SeqCst).into() {
                 OverSampleType::None => {
                     output.modify_channels(|channel| {
-                        channel.iter_mut().for_each(|o| *o = self.apply_curve(*o));
+                        // curve is always set at this point
+                        let curve = self.curve.as_deref().unwrap();
```
orottier, line 428: This line can be moved after line 423 I think, which will deduplicate it in the 3 match arms.
Also perhaps you could rewrite 423 as a `if let Some(curve) = self.curve {`

orottier (commented):
Thanks a lot for you contribution!

I need a bit more time to review and verify it in full, but I left a few first remarks. Could you have a look?

If you wish, you could add a benchmark to `benches/my_benchmark.rs` to see if it actually outperforms the previous version (it will run in the CI and post a comment on your PR).

uklotzde:
> If you wish, you could add a benchmark to `benches/my_benchmark.rs` to see if it actually outperforms the previous version (it will run in the CI and post a comment on your PR).

Benchmarking is not helpful here. The allocation might only cause issues very rarely but in fatal ways. Same applies to mutexes. Lock-free code is often less performant on average than locking.

orottier:
> Lock-free code is often less performant on average than locking.

Fair point! Although I'm reasonably sure `process_into_buffer` is both lock-free *and* more performant :)

I have merged your updated code. Thanks again

PR #290, opened 2023-06-18
Add pre-commit config and CI workflow

uklotzde:
Proven pre-commit config(s) and the corresponding CI workflow, adopted from other projects. Required various minor fixes, mostly spelling and formatting issues. The CI workflow will prevent occurrence of new issues in the future.

I have deleted the redundant steps from rust.yml.

TODO

- [x] Rebase after #289 has been merged


uklotzde:
The GitHub diff doesn't show a helpful diff of the changes in the renamed (.yml -> .yaml) files. I could remove the last commit and you could apply it later.

CHANGELOG.md
```diff
@@ -15,47 +17,49 @@
...
 
-# Version 0.28.0 (2023-01-30)
+## Version 0.28.0 (2023-01-30)
 
 - Improved AnalyserNode (performance and correctness)
 - Fixed microphone input on Raspberry Pi
 - Fixed overriding channel config for PannerNode
 
-# Version 0.27.0 (2023-01-17)
+---
```
orottier, line 26: Why the horizontal bar?

.github/workflows/pre-commit.yaml
```diff
@@ -0,0 +1,58 @@
...
+        run: |
+          sudo apt-get update && sudo apt-get install -y libasound2-dev libjack-jackd2-dev
+
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.x"
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@master
```
orottier, line 33: Is this an improved version over `actions-rs/toolchain@v1` ? In that case we should probably also use it in `rust.yaml`

Also, maybe best to not depend on bleeding edge?
```suggestion
        uses: dtolnay/rust-toolchain@v1
```

.github/workflows/pre-commit.yaml
```diff
@@ -0,0 +1,58 @@
...
+  CARGO_TERM_COLOR: always
+
+jobs:
+  pre-commit:
+    name: Detecting code style issues
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install ALSA and Jack dependencies
+        run: |
+          sudo apt-get update && sudo apt-get install -y libasound2-dev libjack-jackd2-dev
```
orottier, line 25: Jack is not strictly necessary to build the lib (but it won't hurt either). We don't install it in the `rust.yaml` run

orottier (approved):
Thanks for your next contribution!

I must admit I felt a bit hesitant to bring all this complexity and the number of dependencies to this project. But since none of this runs on my local machine, and we don't store any secrets in CI, I figured we could try it out.

I left a few nitpicks. Could you have a look?
Also, do you think it is possible to move all these config files (spell, prettierrc, etc) into a `config/` folder? If that is a lot of work, don't bother.

CHANGELOG.md
uklotzde: Oops. Left over from a previous attempt to fix the issues.

.github/workflows/pre-commit.yaml
uklotzde: Yes. actions-rs/toolchain@v1 is unmaintained and has issues. I have switched all my projects over.

.github/workflows/pre-commit.yaml
uklotzde: This way we could simply use `--all-features` for the checks.

CHANGELOG.md
uklotzde: I have fixed the wrong commit and force pushed the changes. Hope this is ok.

.github/workflows/pre-commit.yaml
uklotzde: I don't remember why I have chosen `master` over `v1`. It never caused any issues and it doesn't hurt unless the CI build breaks. Since we use the most recent Rust version for the CI builds we should better rely on the most recent tooling.

uklotzde:
> I must admit I felt a bit hesitant to bring all this complexity and the number of dependencies to this project. But since none of this runs on my local machine, and we don't store any secrets in CI, I figured we could try it out.

pre-commit is convenient for running all those checks on your local machine without depending on GitHub CI. Each roundtrip to GitHub is time consuming and inconvenient. It introduces a considerable delay in the feedback cycle during development.

.github/workflows/pre-commit.yaml
uklotzde: According to the README `@stable` is what we want: https://github.com/orottier/web-audio-api-rs/pull/290/commits/336f65b647196fc0bd392655cafb1675e23e7eef

uklotzde:
> Also, do you think it is possible to move all these config files (spell, prettierrc, etc) into a `config/` folder? If that is a lot of work, don't bother.

Uncommon, I have not seen this yet. As a workaround to reduce the clutter I always use the hidden file name starting with `.` if available.

PR #291, opened 2023-06-18
Simplify and align AtomicF32/F64 with std::sync::atomic

uklotzde:
- Adhere to the _Principle Of Least Surprise_ for developers.
- Overly restrictive memory barriers could reduce performance considerably.

I expect that `SeqCst` is only needed for very few use cases. Most of the orderings could probably be downgraded to `Relaxed`. Stronger guarantees are only needed for synchronization purposes. Not for shared parameters that could change at any time in no particular order.

### DIFF
--- a/src/control.rs
+++ b/src/control.rs
@@ -23,24 +23,24 @@ impl Scheduler {
 
     /// Retrieve playback start value
     pub fn get_start_at(&self) -> f64 {
-        self.start.load()
+        self.start.load(Ordering::SeqCst)
     }
 
     /// Schedule playback start at this timestamp
     pub fn start_at(&self, start: f64) {
         // todo panic on invalid values, or when already called
-        self.start.store(start);
+        self.start.store(start, Ordering::SeqCst);
     }
 
     /// Retrieve playback stop value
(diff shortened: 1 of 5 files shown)


orottier (approved):
Thanks! Makes sense to align these calls.

The usage of `SeqCst` is my (probably misguided) attempt to make sure audio node configuration changes do not appear out of order, which in theory could have an effect on the audio output if the render thread just happens to come by at the wrong time.

I'd be fine with changing all of these to `Relaxed`. Mainly because we have other 'out of order issues' because some audio node config settings are not propagated by atomics, but by message passing. The atomic settings are thus able to 'overtake' the settings that were passed by message.

uklotzde:
If the ordering is critical then you must solely use message passing for feeding input parameters to the real-time kernel. Atomics are the wrong tool for this purpose. If  I would have designed it I would not have exposed writable parameters as atomics.

On the other hand exposing the current value in the real-time kernel as an atomic (read-only) is valid. But for this purpose `Relaxed` is totally appropriate.

uklotzde:
Revisiting all the memory orderings and double-checking it with the spec should be done in separate PRs.

PR #293, opened 2023-06-18
cpal-jack: Fix panic caused by unwrap()

### LINKED ISSUE #288: Choppy and heavily distorted output
On Fedora 38 none of the examples produces a clean output. The output is always choppy and heavily distorted, even if ALSA does not report any buffer underruns.

I noticed that the results vary on different runs. A clean run followed by a completely distorted run is possible. The distorted output even continues after the executable has terminated.

In contrast the cpal examples work as expected.

What am I missing? Why do even the simple examples do not work out of the box on a common platform?

uklotzde:
Using a latency of 0.0 as a fallback seems to work. I have no clue how this undocumented latency calculation is supposed to work.

```
[2023-06-18T23:44:55Z INFO  web_audio_api::io::cpal] Audio Output Host: cpal Jack
[2023-06-18T23:44:55Z INFO  web_audio_api::io::cpal] Output device: Ok("cpal_client_out")
[2023-06-18T23:44:55Z DEBUG web_audio_api::io::cpal] Attempt output stream with preferred config: StreamConfig { channels: 2, sample_rate: SampleRate(48000), buffer_size: Fixed(1024) }
[2023-06-18T23:44:56Z DEBUG web_audio_api::io::cpal] Output stream set up successfully
```

Also fixes #288 for me. No wonder with such a generous buffer size.

### DIFF
--- a/src/io/cpal.rs
+++ b/src/io/cpal.rs
@@ -425,11 +425,11 @@ impl AudioBackendManager for CpalBackend {
 
 fn latency_in_seconds(infos: &OutputCallbackInfo) -> f64 {
     let timestamp = infos.timestamp();
-    let delta = timestamp
+    timestamp
         .playback
         .duration_since(&timestamp.callback)
-        .unwrap();
-    delta.as_secs() as f64 + delta.subsec_nanos() as f64 * 1e-9
+        .map(|delta| delta.as_secs() as f64 + delta.subsec_nanos() as f64 * 1e-9)
+        .unwrap_or(0.0)
 }
 
 /// Creates an output stream


b-ma:
Nice catch thanks, I didn't realized the `cpal/jack` backend didn't report that value correctly. Looks like a `cpal` issue at first sight, but maybe this is a bit more tricky as the chosen `buffer_size` may not align with the Jack block size. To be investigated (opened #294 to that end)

> I have no clue how this undocumented latency calculation is supposed to work.

This function is only used to report the output latency of the system, cf. [https://webaudio.github.io/web-audio-api/#dom-audiocontext-outputlatency](https://webaudio.github.io/web-audio-api/#dom-audiocontext-outputlatency). In your case, if you configured the Jack server with a block size of 1024 and kept the default period (which is 2), it should report `1024 / 48000 * 2`, or ~0.0426 seconds. Then, returning 0 will indeed work but that's a lie :)

> Also fixes https://github.com/orottier/web-audio-api-rs/issues/288 for me. No wonder with such a generous buffer size.

Yup sure, but using JACK I think you should be able to go with the default render size of 128 without any issues

orottier:
Okay, I'll merge this because I think lying is better than crashing ;)
Let's continue discussion in #294

PR #296, opened 2023-06-19
AudioParam: Reduce code duplication and Arc usage

uklotzde:
- Extract common fields into `AudioParamBase`
- Extract atomic fields into `AudioParamShared`, wrapped into a single `Arc`

The code might not be elegant. But revealing and making those dependencies explicit is the first step before improving the design.

The static decomposition should not have any noticeable impact on performance or memory usage, although the compiler now has less options for reordering and packing the fields. Using only a single `Arc` instead of two could improve performance slightly.

### DIFF
--- a/src/param.rs
+++ b/src/param.rs
@@ -194,24 +194,23 @@ impl AudioParamEventTimeline {
 /// AudioParam controls an individual aspect of an AudioNode's functionality, such as volume.
 pub struct AudioParam {
     registration: AudioContextRegistration,
-    is_a_rate: Arc<AtomicBool>,
-    automation_rate_constrained: bool,
+    raw_parts: AudioParamRaw,
+}
+
+// helper struct shared between the actual param and the processor
+#[derive(Debug, Clone)]
+pub(crate) struct AudioParamBase {
+    shared_parts: Arc<AudioParamShared>,
     default_value: f32, // readonly
     min_value: f32,     // readonly
     max_value: f32,     // readonly
-    current_value: Arc<AtomicF32>,
-    sender: Sender<AudioParamEvent>,
 }
 
 // helper struct to attach / detach to context (for borrow reasons)
 #[derive(Clone)]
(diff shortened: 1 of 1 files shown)


orottier:
Thanks again.

The explicit layout of `AudioParam > AudioParamRaw > AudioParamBase > AudioParamShared` has some merits but snippets like this are rather daunting:

```rust
        self.raw_parts
            .base_parts
            .shared_parts
            .current_value
            .store(clamped, Ordering::SeqCst);
```

> But revealing and making those dependencies explicit is the first step before improving the design.

Do you have any specific ideas or issues to address? It would help me decide whether this new layout pulls its weight

uklotzde:
> Do you have any specific ideas or issues to address? It would help me decide whether this new layout pulls its weight

These chains only appear internally and the actual memory layout is not visible from the outside. Reducing the redundancies upfront might be helpful when sorting out the synchronization between `AudioParam` and `AudioParamProcessor` as proposed in #297. Sure, they are not elegant. But that's a consequence of the current design which does not seem to be ideal.

I prefer to group multiple fields that appear repeatedly in different contexts as distinct types. Replicating them anonymously across different types would hide these dependencies. Makes it easier to spot commonalities and differences for subsequent refactorings. We should end up with very few types that share only a single, common core or base.

orottier:
Thanks for the elaborate reply.

> current design which does not seem to be ideal.
> subsequent refactorings

My question about specific ideas or issues to address still stands. I'm hesitant to refactor for the sake of abstract reasons such as redundancies and commonalities if there is no pressing issue at hand. Surely, you are technically correct about the flat layout hiding possible future improvements, but to refactor also incurs a slight cost since the original authors of the file will have to grok the changes.

It would really help me make a judgement if you could share the next steps you have in mind with this module.

uklotzde:
The issues I tried to fix are probably only the tip of the iceberg.

PR #298, opened 2023-06-20
document and add example option when running on ALSA

b-ma:
Proposal according to the discussion on ALSA render size in #288 

- added some documentation in the README
- sketch adding an argument in the `amplitude_modulation` example

If it's ok for you, I can generalize the pattern to all examples, 

Let me know!


orottier:
Thanks, this is a great start.

Since some of the examples takes arguments themselves, I think it is better to pass the setting with an environment variable so they don't clutter. Like so:

`WEB_AUDIO_LATENCY=playback cargo run --release --example amplitude_modulation`

Or more verbose:
```
export WEB_AUDIO_LATENCY=playback
cargo run --release --example amplitude_modulation
```

The env vars can be read with https://doc.rust-lang.org/std/env/fn.var.html

Now all examples will share some common code:
1. initialize the env_logger
2. setup an AudioContext with the desired playback rate

I wonder if we should introduce a (hidden) method to our library `init_example() -> AudioContext` which all examples call.
The benefit is that we don't write this snippet over and over again. Drawback is that examples are meant as copy and paste inspiration for users and an exotic helper function is not really great in that case

b-ma:
Ok, except for the toy_webrtc which was not working properly for some reason I didn't dig into, I updated all the examples to optionally handle the env variable.

I chose the verbose solution, I think it is best to keep things really simple for new users

b-ma:
Yup, nice it's shorter indeed, I will make the changes tomorrow

b-ma:
Done!

orottier:
Thanks for the hard work!

PR #299, opened 2023-06-20
Fix: use cpal feature only for clippy

b-ma:
Just a small fix in the pre-commit file that was crashing on Mac because of the clippy `--all-features` flag, I guess because the JACK backend is not implemented by cpal for Mac or Windows:

```
error[E0599]: no variant or associated item named `Jack` found for enum `cpal::HostId` in the current scope
  --> src/io/cpal.rs:77:45
   |
77 |             .find(|id| *id == cpal::HostId::Jack)
   |                                             ^^^^ variant or associated item not found in `HostId`
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -64,7 +64,7 @@ repos:
             --locked,
             --workspace,
             --features,
-            cpal,
+            cubeb,
             --all-targets,
             --,
             -D,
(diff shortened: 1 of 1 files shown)


orottier:
Thanks for the fix. Indeed `jack` won't be available on devices that don't have it installed.
I have slightly changed your work. `cpal` is a default feature so no need to specify it. I added `cubeb` because that is what we test in CI as well and should never fail on your local machine.

Issue #297, opened 2023-06-19
Prevent out-of-order arrival of audio node settings


orottier:
Some of the audio node settings are propagated to the render thread via message passing, e.g. `OscillatorNode.set_periodic_wave`.
Others are passed with atomics, e.g. `OscillatorNode.set_frequency`.

The atomic settings are applied the moment the node's `AudioProcessor` is called, while the settings that are set via the message bus are only consumed at the start of the render quantum.

This means there is a non-negligible chance that (for a single render quantum) only the frequency setting is changed for this snippet:

```rust
osc.set_periodic_wave(w);
osc.set_frequency(440.);
```

We should fix this by applying all settings via the message bus.
CC @uklotzde  and https://github.com/orottier/web-audio-api-rs/pull/291#pullrequestreview-1485173569

uklotzde:
I also noticed that each `AudioParam`/`AudioParamProcessor` pair are connected by individual channels? This is probably not what you want. Messages should be dispatched and applied sequentially in the same order as they are received among the `AudioParamProcessor`s.

I also consider using _unbounded_ channels for this purpose as inappropriate. For any for real world use case _bounded_ channels are recommended. Moreover the receiver side of theses crossbeam channels cannot be used safely in a real-time context AFAIK.

The crossbeam channels seem to have serious unresolved issues: https://github.com/crossbeam-rs/crossbeam/issues/821

uklotzde:
> This is a rather terrible implementation I must admit ...

No offense, "make it work" is the appropriate strategy for getting started. Almost everything could be fixed later.

uklotzde:
Leaving (TODO) comments with your thoughts, assumptions, and any shortcomings you are aware of when writing code would help others (or your future self) to catch up when needed.

PR #305, opened 2023-06-21
AudioParam: Use Acquire/Release ordering for handover in atomics

uklotzde:
I consider this as the _canonical ordering_ for passing of values in atomics to achieve a handover like a batton.

Unfortunately the spec explicitly allows write access to the _current value_ slot by anyone, not only the processor.

If this applies to more use cases then the load and store orderings should be defined as constants and documented.

Thoughts?

### DIFF
--- a/src/param.rs
+++ b/src/param.rs
@@ -257,9 +257,7 @@ impl AudioNode for AudioParam {
 impl AudioParam {
     /// Current value of the automation rate of the AudioParam
     pub fn automation_rate(&self) -> AutomationRate {
-        self.raw_parts
-            .shared_parts
-            .load_automation_rate(Ordering::SeqCst)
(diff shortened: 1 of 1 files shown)


orottier:
I'm no expert on this so my stance is always:
- `Relaxed` when the threads don't really interact but just need to share some info
- `SeqCst` otherwise

Your reasoning sounds sensible and will be more performant, so I am all for it.

In the repo you will probably find the following other atomics usage:
- `Relaxed` for sharing render thread state (number of frames processed, etc). I think this is fine.
- `SeqCst` for passing audio node configs. This we will refactor anyway in #297

orottier:
By the way, thanks again for your contribution. Your expertise is much appreciated!

PR #307, opened 2023-06-24
ChannelConfig is now a single Arc instead of 3

orottier:


b-ma:
All good to me too!

b-ma:
Do you think it would be possible to make a small release soon? I would need to update the node bindings to work with Jack in the next weeks

PR #310, opened 2023-06-25
Prevent needless clone in offline rendering + in the online happy path

orottier:


b-ma:
Hey, the benches seems to show an interesting improvement indeed, but there is something I don't quite understand here:
As the `AudioRenderQuantum` is supposed to have "copy on write" semantic, why is this different with the `Cow`, at first sight it looks really similar to me?

orottier:
Yes, definitely interesting and a bigger change than I had expected. Note however that this is all quite synthetical because of these benches don't really resemble real life workloads.

I came up with this fix because I was looking a flamegraphs for #311 and saw the `clone` method taking a non-negligible amount of time.

My hand-wavy argument would be that `bench_ctor` and `bench_buffer_src` do so little work per render quantum that the shallow copy of our copy-on-write audio buffers are still a factor. 
As you can see in the diff we have completely elided any overhead for the offline rendering path.
In any real life scenario with 5+ audio nodes, or online audio, this will not be interesting. I could drop the `Cow` from the online rendering part and for simplicity opt for a clone there.

I looked into the actual `clone` code inside the AudioRender and I see 2 reasons why it is not maximally performant:
- `AudioRenderQuantumChannel` could use a single `Rc` instead of 2
- the `clone` method of the underlying `ArrayVec` in `AudioRenderQuantum` has protection for panics. But we know a clone of `AudioRenderQuantumChannel` is panic free so we could use a specialized implementation

I could amend this PR with some further experiments

orottier:
Okay, none of my suggestions brought any real life performance gains so I'm making the PR simpler and I'm leaving it at that.

PR #313, opened 2023-07-02
Use unified function syntax for ref-counted types

uklotzde:
Indicates that this is a shallow clone operation.


orottier:
Thanks! Explicit is good

PR #316, opened 2023-07-11
Update MSRV to 1.70 and remove lazy_static and once_cell

uklotzde:
Reduce some inconsistencies and remove obsolete crates.

### DIFF
--- a/src/node/audio_buffer_source.rs
+++ b/src/node/audio_buffer_source.rs
@@ -1,4 +1,4 @@
-use once_cell::sync::OnceCell;
+use std::cell::OnceCell;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
--- a/src/node/stereo_panner.rs
+++ b/src/node/stereo_panner.rs
@@ -4,7 +4,7 @@ use crate::param::{AudioParam, AudioParamDescriptor};
 use crate::render::{AudioParamValues, AudioProcessor, AudioRenderQuantum, RenderScope};
 
 use super::{
-    sine_table, AudioNode, ChannelConfig, ChannelConfigOptions, ChannelCountMode,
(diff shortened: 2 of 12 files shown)


orottier (approved):
Thanks, fixes #302 
Interesting performance improvement as well..
A MSRV of the current version is potentially disruptive. Let's wait at least for 1.71 to arrive (any time now, in pre-release)

src/node/audio_buffer_source.rs
```diff
@@ -1,4 +1,4 @@
-use once_cell::sync::OnceCell;
+use std::cell::OnceCell;
```
uklotzde, line 1: Switching from `sync` to *non-sync* here might make a difference by avoiding a mutex.

src/node/stereo_panner.rs
```diff
@@ -71,8 +71,10 @@ fn assert_valid_channel_count_mode(mode: ChannelCountMode) {
 #[inline(always)]
 fn get_stereo_gains(x: f32) -> [f32; 2] {
     let idx = (x * TABLE_LENGTH_BY_4_F32) as usize;
-    let gain_left = SINETABLE[idx + TABLE_LENGTH_BY_4_USIZE];
-    let gain_right = SINETABLE[idx];
+
+    let sine_table = sine_table();
```
uklotzde, line 75: Now the reference is only resolved once instead of twice per call.

uklotzde: Now the access to OnceLock only happens once during construction.

orottier (approved):

uklotzde:
I don't see changes that would require bumping the MSRV to 1.71 right now: https://releases.rs/docs/1.71.0/

We could do this at any time when actually needed.

uklotzde:
I have added a GitHub action for checking the MSRV. Let's see if this works.

PR #317, opened 2023-07-12
Fix broken resampling in MediaElement, support more media types

orottier:
For your info @b-ma, @uklotzde and I have taken over maintenance of the `creek` dependency.

Some features of the media element that we don't have implemented right now (reverse playback, and perhaps pitch preserving playback settings) could be implemented there.

### DIFF
--- a/tests/media_element.rs
+++ b/tests/media_element.rs
@@ -17,7 +17,9 @@ fn test_media_element_source_progress() {
     src.connect(&context.destination());
     media.play();
 
-    // sleep for a bit, make sure the audio thread has started
+    // TODO improve test setup of online AudioContext
+    // <https://github.com/orottier/web-audio-api-rs/issues/323>
+    // Sleep for a bit, make sure the audio thread has started
     std::thread::sleep(std::time::Duration::from_millis(100));
 
     // assert the media has progressed
--- a/tests/media_element.rs
+++ b/tests/media_element.rs
@@ -1,0 +1,25 @@
+use web_audio_api::context::{AudioContext, AudioContextOptions, BaseAudioContext};
+use web_audio_api::node::AudioNode;
+use web_audio_api::MediaElement;
+
+#[test]
+fn test_media_element_source_progress() {
+    let options = AudioContextOptions {
+        sink_id: "none".into(),
+        ..AudioContextOptions::default()
+    };
+    let context = AudioContext::new(options);
+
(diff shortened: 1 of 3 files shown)


uklotzde:
How did you recognize that resampling was broken? How do the changes prevent this from happening again in the future?

orottier:
> How did you recognize that resampling was broken? How do the changes prevent this from happening again in the future?

Good question, I should probably add a regression test, but those are always a bit tricky for nodes that can only exist in live `AudioContext`s. I can cook up something though.

The actual issue was that a paused stream emits silence with only a single channel. Then, if the media starts playing with more channels, it confused the dead-simple Resampler we have for the media element:

```
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `2`', src/buffer.rs:277:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/panicking.rs:578:5
   1: core::panicking::panic_fmt
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/panicking.rs:67:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
   4: web_audio_api::buffer::AudioBuffer::extend
   5: <web_audio_api::resampling::Resampler<M> as core::iter::traits::iterator::Iterator>::next
   6: <web_audio_api::node::MediaStreamRenderer<R> as web_audio_api::render::processor::AudioProcessor>::process
   7: web_audio_api::render::graph::Graph::render
   8: web_audio_api::render::thread::RenderThread::render
```

Looking at the actual Resampler, it could use some love. It's not entirely realtime safe

orottier:
The integration test is not perfect (using sleep to wait for the render thread to start) but will prevent regressions for now.

tests/media_element.rs
```diff
@@ -0,0 +1,25 @@
...
+    let context = AudioContext::new(options);
+
+    let mut media = MediaElement::new("samples/major-scale.ogg").unwrap();
+    media.set_loop(true);
+
+    let src = context.create_media_element_source(&mut media);
+    src.connect(&context.destination());
+    media.play();
+
+    // sleep for a bit, make sure the audio thread has started
```
uklotzde, line 20: Please add a `TODO` or `FIXME` comment that allows to find and resolve these little workarounds or hacks later.

PR #319, opened 2023-07-15
Fix `AudioDestination` constraints

b-ma:
Just a very small fix


b-ma:
I don't really get why clippy is complaining, all checks pass on my machine and I didn't touch the files that are faulty, any ideas?

orottier:
> I don't really get why clippy is complaining, all checks pass on my machine and I didn't touch the files that are faulty, any ideas?

It's a new rust version with new lints enabled by default. You will notice them when you upgrade! I will fix them on the main branch

orottier:
Thanks, nice catch!

PR #325, opened 2023-07-16
RenderThread: Rename method according to code comments

uklotzde:
The code comments seem to be inconsistent with the code. This is what I would expect according to the comments.


orottier:
Fair enough.
For context, we switched to the (non-compliant `_sync` methods) some time ago to delay implementation of the async interface #104 #177

PR #327, opened 2023-07-16
Assess performance impact of MAX_CHANNEL_COUNT (32 -> 256)

orottier:
Some investigations for https://github.com/ircam-ismm/node-web-audio-api/issues/23#issuecomment-1637078022

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,7 +48,7 @@ use std::sync::atomic::{AtomicU32, AtomicU64, Ordering};
 pub(crate) const RENDER_QUANTUM_SIZE: usize = 128;
 
 /// Maximum number of channels for audio processing
-pub const MAX_CHANNELS: usize = 32;
+pub const MAX_CHANNELS: usize = 256;
 
 mod buffer;
 pub use buffer::*;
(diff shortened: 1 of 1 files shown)


b-ma:
Whow quite impressive

orottier:
Indeed. But we should realize these CI benchmark are rather synthetic because they are microbenchmarks, not real world workloads. But the benchmarks.rs example shows a similar - less extreme - trend.

For fun and giggles I will perform another run with 4 channels max, but this is not permissable.
E.g. even if the audio output hardware is stereo, it is possible to have intermediate buffers in the audio graph with 32 channels (when using a large ChannelMergerNode maybe) which are downmixed before they reach the destination node.

All in all this shows we should not increase the MAX_CHANNEL_COUNT to a large number because this will negatively affect many users. I will continue to explore solutions

orottier:
@b-ma with the issues around 32+ channels counts resolved, I think it's better to leave this as it is and focus on other things. Unless you think there's merit to supporting larger channels counts

b-ma:
Yup this is really ok I think. That's really niche requirements and it's always possible to make a specific build changing the MAX_CHANNEL_COUNT constant if really required

PR #328, opened 2023-07-17
Fix output stream number of channels

### LINKED ISSUE #320: Possible issues with soundcards with number of channels > 32
_disclaimer: I didn't make the actual tests so I might be mistaking, I will try to do that in the next weeks_

The problem:
- If we have a soundcard with a number of output larger than 32 (e.g. using blackhole w/ 256 channels or equivalent)
- Then we need to configure the destination, e.g.:
```js
audioContext.destination.channelCount = 12;
audioContext.destination.channelInterpretation = 'discrete';
```
to not fall into the number of channels and/or mixing rules errors

The question:
- Can we do that safely as the context is resumed by default?

b-ma:
Fixes #320 and https://github.com/ircam-ismm/node-web-audio-api/issues/23

Actually the issue was that the buffer returned by the destination node is filled with silence up to the backend number of channel by the render thread using `AudioRenderQuantum::mix`, cf. https://github.com/orottier/web-audio-api-rs/blob/ec5f9efecf696683f69c4c05fcd4e6737ac75126/src/render/thread.rs#L264-L266

But the backend streams were actually not clamped to `MAX_CHANNELS` therefore the `assert_valid_number_of_channels` in `AudioRenderQuantum::mix` was failing.

Also `destination.maxNumberOfChannels` now returns 32 even if the soundcard has actually more output channels (behaviour checked on Chrome and Firefox)

Note that I've added a Max patch in the examples folder to help checking the multichannel output with a virtual soundcard. Even if we agree to keep this kind of stuff, I'm not sure this is nice as Max is a paid software, would be better with a free solution (e.g. PureData)

### DIFF
--- a/src/io/cpal.rs
+++ b/src/io/cpal.rs
@@ -16,7 +16,7 @@ use crate::context::AudioContextOptions;
 use crate::io::microphone::MicrophoneRender;
 use crate::media_devices::{MediaDeviceInfo, MediaDeviceInfoKind};
 use crate::render::RenderThread;
-use crate::AtomicF64;
+use crate::{AtomicF64, MAX_CHANNELS};
 
 use crossbeam_channel::Receiver;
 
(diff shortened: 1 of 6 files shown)


src/io/cpal.rs
```diff
@@ -366,7 +382,7 @@ impl AudioBackendManager for CpalBackend {
     }
 
     fn output_latency(&self) -> f64 {
-        self.output_latency.load(Ordering::SeqCst)
+        self.output_latency.load(Ordering::Relaxed)
```
b-ma, line 385: I choose the relaxed ordering because I don't think this is really important, let me know if you see any issue with that

orottier: Yes, this is fine

orottier (approved):
Great debugging. Glad to have this fixed

Issue #315, opened 2023-07-11
Rewrite audio nodes to use proper message passing instead of atomics


orottier:
Continuation from the initial work of #311 , these nodes still need changes:

- [x] BiquadFilterNode
- [x] ConvolverNode
- [x] PannerNode
- [x] WaveShaperNode

For further consideration
- [ ]  also rewrite the ChannelConfig to use message passing

b-ma:
Actually about https://github.com/orottier/web-audio-api-rs/pull/329, I think I will just remove the new `LoopControl` struct as there is finally no `Arc`, anything on your sides against that?

orottier:
Fixed with #337

PR #329, opened 2023-07-17
Merge `Controller` into `AudioBufferSourceNode`

b-ma:
Following comment in #311

Compared to the previous version, the strategy for the loop parameters is as follow: 
store the given values in the `loop_parameters` thread safe instance which only lives in control thread(s) and send a message to the render thread which stores only the raw values.
Thus, the values between control and render side might be desynchronised for a little while but the developer experience will appear more logical, i.e.

```rs
node.set_loop(true);
println!("{:?}", node.loop_());
> true // is guaranteed
```

Note that this seems to be the strategy used by Firefox.

Let me know if you disagree with that


orottier:
I don't understand why the benchmarks are acting up. My assumption is the mod layout does not affect the actual compilation as a crate is treated as a single compilation unit. To be investigated

src/node/audio_buffer_source.rs
```diff
@@ -49,9 +48,55 @@ struct PlaybackInfo {
...
+// Values between control and render side might be desynchronised for little while
+// but the developer experience will appear more logical, i.e.
+// ```no_run
+// node.set_loop(true);
+// println!("{:?}", node.loop_());
+// > true // is guaranteed
+// ```
+// Note that this seems to be the strategy used by Firefox
+#[derive(Debug)]
+struct LoopParameters {
```
uklotzde, line 63: How about renaming this to `LoopControl` because it is a stateful, mutable container? The name `LoopParameters` suggests that this is just a simple value object, i.e. a POD.

src/node/audio_buffer_source.rs
```diff
@@ -333,7 +384,9 @@ struct AudioBufferSourceRenderer {
     stop_time: f64,
     offset: f64,
     duration: f64,
-    controller: Controller,
+    loop_: bool,
```
uklotzde, line 387: I like the idea of keeping the atomics out of the render thread as much as possible.

Ideally, the render thread should only store public output values in atomics for monitoring live values from external threads. With only a single writer there won't be any race conditions and all changes are repeatable.

The render thread should not load inputs (= control values) from atomics that could change unexpectedly when being modified by multiple writers concurrently. Modifications of input values should usually performed by submitting messages. This ensures that the derived internal state remains consistent. Hidden internal state does not need to be stored in atomics, only the public output values as outlined above.

src/node/audio_buffer_source.rs
b-ma: ok done

b-ma:
> I don't understand why the benchmarks are acting up. My assumption is the mod layout does not affect the actual compilation as a crate is treated as a single compilation unit. To be investigated

Yup that's strange and a bit random: sometimes only `bench_buffer_src_biquad` is really affected, sometimes `bench_buffer_src_biquad` and `bench_stereo_positional` are... really no idea on my side

orottier:
Okay the benchmarks are weird for other PRs as well so I will just merge this!

src/node/audio_buffer_source.rs
orottier: agreed! We should verify this for all nodes that need to be changed still #315

PR #332, opened 2023-07-18
AudioProcessor::onmessage(): Pass message as mutable reference

uklotzde:
As discussed.

### DIFF
--- a/examples/worklet.rs
+++ b/examples/worklet.rs
@@ -143,7 +143,7 @@ impl AudioProcessor for WhiteNoiseProcessor {
         true // source node will always be active
     }
 
-    fn onmessage(&mut self, msg: &mut Box<dyn Any + Send + 'static>) {
+    fn onmessage(&mut self, msg: &mut dyn Any) {
         // handle incoming signals requesting for change of color
         if let Some(color) = msg.downcast_ref::<NoiseColor>() {
             self.color = *color;
--- a/examples/worklet.rs
+++ b/examples/worklet.rs
@@ -1,3 +1,5 @@
+use std::any::Any;
+
 use rand::Rng;
 
 use web_audio_api::context::{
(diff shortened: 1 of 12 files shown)


examples/worklet.rs
```diff
@@ -141,15 +143,12 @@ impl AudioProcessor for WhiteNoiseProcessor {
         true // source node will always be active
     }
 
-    fn onmessage(&mut self, msg: Box<dyn std::any::Any + Send + 'static>) {
+    fn onmessage(&mut self, msg: &mut Box<dyn Any + Send + 'static>) {
```
orottier, line 146:
```suggestion
    fn onmessage(&mut self, msg: &mut dyn Any) {
```

This will work as well and looks a lot cleaner. Is there any drawback?

orottier (commented):
Thanks! I have one suggestion, care to have a look?

src/buffer.rs
```diff
@@ -64,7 +64,7 @@ pub struct AudioBufferOptions {
 ///
 /// - `cargo run --release --example audio_buffer`
 ///
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
```
orottier, line 67: This default implementation will create an audio buffer with zero channels, which we deem invalid - `assert_valid_number_of_channels` requires at least one.
This is a bit tricky, I'd rather not expose this, but there is no other way provide a default implementation that does not allocate..

orottier: One solution would be to change
`channels: Vec<ChannelData>` to
`channels: SmallVec<[ChannelData; 2]>,`
which is probably a nice performance gain anyway.

orottier: I have written up #334 in case you would like to defer this discussion and focus on the PR at hand

uklotzde: Adding `Default` was just for convenience. I can create a tombstone instance manually that replaces the message content.

examples/worklet.rs
uklotzde: That didn't work and I don't know why yet.

uklotzde: I have dropped the non-working commit. Downcasting doesn't seem to work then. We probably need to keep it boxed for preserving the `Send` and `'static` bounds.

uklotzde: Fixed. Working with `Any` is tricky. I simply forgot to invoke `Box::as_mut` instead of using `&mut` when passing the parameter :see_no_evil:

orottier (approved):
@b-ma do you agree with this signature of `onmessage`? It's a bit niche but I don't expect many users to actually run into this

b-ma:
Hey, not sure I understand everything going on there but seems quite clean to me indeed!

Also seems that we wouldn't be far from having a really compliant [messageport](https://html.spec.whatwg.org/multipage/web-messaging.html#messageport) API from there, that's nice

PR #336, opened 2023-07-19
Add garbage collector thread as sidecar of render thread

uklotzde:
Based on #322.

### TODO

- [x] Rebase after #322 has been merged.


orottier (approved):
A linked list queue with pre-allocated nodes, nifty!
I was a bit taken aback by the low download count of the `llq` crate but the author appears to be an expert in this field so I'll happily add it to the dependency list.

uklotzde:
We could check later if it would be possible and convenient to use `basedrop` for our use case. Might be slightly more efficient.

PR #337, opened 2023-07-19
Refactor message passing

b-ma:
Work in progress regarding #315

### DIFF
--- a/src/node/constant_source.rs
+++ b/src/node/constant_source.rs
@@ -11,12 +11,13 @@ use super::{AudioNode, AudioScheduledSourceNode, ChannelConfig};
 // dictionary ConstantSourceOptions {
 //   float offset = 1;
 // };
-//
 // @note - Does not extend AudioNodeOptions because AudioNodeOptions are
 // useless for source nodes as they instruct how to upmix the inputs.
 // This is a common source of confusion, see e.g. https://github.com/mdn/content/pull/18472
+>>>>>>> 83a0107 (refactor: renaming variables)
 #[derive(Clone, Debug)]
 pub struct ConstantSourceOptions {
+    /// Initial parameter value of the constant signal
     pub offset: f32,
 }
 
@@ -114,14 +115,16 @@ impl AudioScheduledSourceNode for ConstantSourceNode {
 impl ConstantSourceNode {
     pub fn new<C: BaseAudioContext>(context: &C, options: ConstantSourceOptions) -> Self {
         context.register(move |registration| {
-            let param_opts = AudioParamDescriptor {
+            let ConstantSourceOptions { offset } = options;
+
+            let param_options = AudioParamDescriptor {
                 min_value: f32::MIN,
                 max_value: f32::MAX,
                 default_value: 1.,
                 automation_rate: AutomationRate::A,
             };
-            let (param, proc) = context.create_audio_param(param_opts, &registration);
-            param.set_value(options.offset);
+            let (param, proc) = context.create_audio_param(param_options, &registration);
+            param.set_value(offset);
 
             let render = ConstantSourceRenderer {
                 offset: proc,
--- a/src/node/constant_source.rs
+++ b/src/node/constant_source.rs
@@ -11,10 +11,11 @@ use super::{AudioNode, AudioScheduledSourceNode, ChannelConfig};
 // dictionary ConstantSourceOptions {
 //   float offset = 1;
 // };
+// https://webaudio.github.io/web-audio-api/#ConstantSourceOptions
+//
 // @note - Does not extend AudioNodeOptions because AudioNodeOptions are
-// useless for source nodes as they instruct how to upmix the inputs.
-// This is a common source of confusion, see e.g. https://github.com/mdn/content/pull/18472
->>>>>>> 83a0107 (refactor: renaming variables)
+// useless for source nodes, because they instruct how to upmix the inputs.
+// This is a common source of confusion, see e.g. mdn/content#18472
 #[derive(Clone, Debug)]
 pub struct ConstantSourceOptions {
     /// Initial parameter value of the constant signal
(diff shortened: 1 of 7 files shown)


b-ma:
test

b-ma:
\bench

src/node/convolver.rs
```diff
@@ -1,15 +1,15 @@
...
-use crossbeam_channel::{Receiver, Sender};
-use realfft::{num_complex::Complex, ComplexToReal, RealFftPlanner, RealToComplex};
-use std::sync::{
-    atomic::{AtomicBool, Ordering},
-    Arc, Mutex,
-};
+use super::{AudioNode, ChannelConfig, ChannelConfigOptions, ChannelInterpretation};
 
 /// Scale buffer by an equal-power normalization
 fn normalization(buffer: &AudioBuffer) -> f32 {
```
uklotzde, line 15: Unrelated: Why does this function name does not start with a *verb* as usual?

src/node/panner.rs
```diff
@@ -271,16 +282,14 @@ pub struct PannerNode {
...
     orientation_y: AudioParam,
     orientation_z: AudioParam,
-    cone_inner_angle: Arc<AtomicF64>,
-    cone_outer_angle: Arc<AtomicF64>,
-    cone_outer_gain: Arc<AtomicF64>,
-    distance_model: Arc<AtomicU8>,
-    ref_distance: Arc<AtomicF64>,
-    max_distance: Arc<AtomicF64>,
-    rolloff_factor: Arc<AtomicF64>,
+    cone_inner_angle: AtomicF64,
```
uklotzde, line 285: Why still individual atomics instead of an inner typed wrapped into a single `RefCell` as demonstrated in in the `AudioBufferSourceNode`? This also applies to all other node types as far as I could see.

uklotzde (commented):
It would be easier for review to refactor the node types one after another. All these changes are not dependent on each other.

uklotzde:
This PR does not seem to fix the soundness errors that are potentially causing inconsistencies in multi-threaded contexts.

orottier:
> This PR does not seem to fix the soundness errors that are potentially causing inconsistencies in multi-threaded contexts.

Yeah, I think @b-ma is keeping these things separate for now.
This PR addresses the leftover nodes that do not use message passing yet, as described in #315 
Indeed this introduces soundness issues, but let's decide on the way forward first in #344 . I won't release in the meantime
(I'm a bit busy at the moment but will propose something shortly)

b-ma:
> Edit, I just read
> 
> > Edit: I tried to follow the pattern proposed by @uklotzde in the AudioBufferSourceNode, let me know if I misunderstood something
> 
> but I still think it is fine to keep matters separate for now

Yup that was the idea, I was actually thinking about the post_message / onmessage pattern used for the AudioBuffer and control messages

src/node/constant_source.rs
```diff
@@ -11,8 +11,11 @@ use super::{AudioNode, AudioScheduledSourceNode, ChannelConfig};
 // dictionary ConstantSourceOptions {
 //   float offset = 1;
 // };
+// @note: this does not extend AudioNodeOptions in the spec, why?
+// https://webaudio.github.io/web-audio-api/#ConstantSourceOptions
```
orottier, line 15: `AudioNodeOptions` are useless for source nodes, because they instruct how to upmix the inputs.
This is a common source of confusion, see e.g. https://github.com/mdn/content/pull/18472
Also for the spec editors because they accidentally let OscillatorOptions extend AudioNodeOptions

src/node/convolver.rs
```diff
@@ -359,6 +344,13 @@ impl ConvolverRendererInner {
         }
     }
 
+    // dummy convolver used to init renderer
+    fn tombstone() -> Self {
+        // just use arbitrary common sample rate
+        let padded_buffer = AudioBuffer::from(vec![vec![0.; 0]; 1], 48000.);
```
orottier, line 350: This allocates (non-empty vec). I think you can reuse the tombstone from `AudioBufferSourceRenderer`

orottier:
I have improved the HRTF-loading in the panner node.
You ran into issues because the `onmessage` handler receives an `&mut` of the message. So if you need to take out a value that is not `Copy` or `Clone`, you need to place back a tombstone or, when that's not easily created, wrap the inner value in an `Option` so you can `take` it out.

I will have a closer look at the other changes tomorrow.

src/node/convolver.rs
b-ma: Found a way to make it more clean in any case

src/node/constant_source.rs
b-ma: Added your note and link to the comment, so we don't forget again. I will try to align this to other nodes as well later

b-ma:
> I have improved the HRTF-loading in the panner node.
You ran into issues because the onmessage handler receives an &mut of the message. So if you need to take out a value that is not Copy or Clone, you need to place back a tombstone or, when that's not easily created, wrap the inner value in an Option so you can take it out.

Ok thanks! Think I got it (until next time borrow checker shows up :)

src/node/convolver.rs
```diff
@@ -151,33 +150,34 @@ impl ConvolverNode {
...
 
-            if let Some(buffer) = buffer {
-                node.set_buffer(buffer);
-            }
-
             (node, Box::new(renderer))
-        })
+        });
+
+        // renderer has been sent to render thread, we can sent it messages
```
orottier, line 175:
```suggestion
        // renderer has been sent to render thread, we can send it messages
```

src/node/convolver.rs
```diff
@@ -414,6 +411,11 @@ impl ConvolverRendererInner {
     }
 }
 
+struct ConvolverRenderer {
```
orottier, line 414: As said, I will rewrite this.
A pattern of `is_set` and a tombstone value is way better represented as simply `Option<ConvolverRendererInner>`

orottier (commented):
Thanks for the hard work. I will pick up the stuff I have mentioned.

src/node/convolver.rs
b-ma: sure this is a redondant

orottier:
Ready to merge but I want to publish a release before merging as we still have some bugfixes lined up.
Merging this will force us to address #344 to fix the logic bugs from the atomics

PR #339, opened 2023-07-19
Dependencies

uklotzde:
- Specifying .0 patch versions is pointless
- Update dependencies to the latest minor version. Not essential, only to keep track.

I am actually an advocate of [PSA](https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277).


orottier:
Interesting.
I knew cargo would treat `x.y.z` as `x.y.[z..]` but I had not realized it will also up the minor version when possible.

PR #340, opened 2023-07-19
Update bench bot behavior

### LINKED ISSUE #335: Bot question
Could we somehow have a better strategy with the bot?

I'm not sure how to do that because 1. the "on-demand" can (very) rapidly becomes "I don't care / I forgot that thing" but 2. the constant (huge) reporting also creates noise and makes conversations very of hard to follow...

b-ma:
Fix #335

- split `rust` workflow into `build` and `benchmark`
- `benchmark` workflow now runs only if "/bench" is put into a comment

Hope this works... not sure in which branch the workflow actually runs...

### DIFF
--- a/.github/workflows/benchmark-post-result.yaml
+++ b/.github/workflows/benchmark-post-result.yaml
@@ -1,6 +1,6 @@
 # yaml-language-server: $schema=https://json.schemastore.org/github-workflow
 
-name: Post benchmark results
+name: benchmark post results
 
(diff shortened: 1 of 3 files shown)


b-ma:
Convert to draft, this is probably broken, I will recheck that

b-ma:
ok looks good now (hopefuly)

orottier:
Let's see! Thanks

PR #342, opened 2023-07-19
Use local loop state in both render and control thread

uklotzde:
Neither Firefox nor Chrome/Chromium use atomics for this purpose.

### DIFF
--- a/src/node/audio_buffer_source.rs
+++ b/src/node/audio_buffer_source.rs
@@ -6,7 +6,6 @@ use crate::buffer::AudioBuffer;
 use crate::context::{AudioContextRegistration, AudioParamId, BaseAudioContext};
 use crate::param::{AudioParam, AudioParamDescriptor, AutomationRate};
 use crate::render::{AudioParamValues, AudioProcessor, AudioRenderQuantum, RenderScope};
-use crate::AtomicF64;
 use crate::RENDER_QUANTUM_SIZE;
 
 use super::{AudioNode, AudioScheduledSourceNode, ChannelConfig};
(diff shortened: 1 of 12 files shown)


uklotzde:
Using atomics just for convenience is dangerous and deceiving. You will then experience all kinds of race conditions, because you can't set the value and send the message atomically at the same time. If you invoke the setters from multiple threads all outcomes are possible.

uklotzde:
Btw, while inspecting the code I was confused why atomics were used here. I expected that the render thread would be the single writer and other threads are just reading the state values. But this was not the case.

b-ma:
Hey,

I definitely agree with @orottier here. Just to add my (opinionated) 2 cents to the discussion:

- Having all nodes non mutable is an API design choice which I think is very nice also for developper experience (let's call it "principle of least surprise"). I.e. for a user, having to know that this node should be `mut` while this other one should not because of some implementation detail is very cumbersome

- It's probably true that theoretically that _"you [can] then experience all kinds of race conditions, because you can't set the value and send the message atomically at the same time"_ but, if such thing ever happens in real world, I would first tend to question the general design and architecture of an application where you can end up doing 2 contradictory things at the exact same time on the same audio node

> I expected that the render thread would be the single writer and other threads are just reading the state values. But this was not the case.

Could you elaborate please?

src/node/audio_buffer_source.rs
```diff
@@ -327,34 +286,44 @@ impl AudioBufferSourceNode {
         &self.detune
     }
 
+    pub fn loop_state(&self) -> &LoopState {
```
b-ma, line 289: This is not in the spec and exposes an internal implementation detail, probably we should stick to the "Keep Your API Surface Small" principle here

uklotzde:
Due to the race conditions of the current implementation the render and the control state will get out of sync when setting control state properties concurrently from different threads.

The API seems to be designed with the assumption of a single control thread as in the browser context. This 1:1 entangling between control and render state needs to be respected by the implementation. Otherwise you need to use a different API.

uklotzde:
> > I expected that the render thread would be the single writer and other threads are just reading the state values. But this was not the case.
> 
> Could you elaborate please?

That would probably require a different API and a less tight coupling between control and render state. A distinction between *desired* and *observed/actual* values as found in control systems with a unidirectional data flow.

b-ma:
> Due to the race conditions of the current implementation the render and the control state ~will~[can] get out of sync when setting control state properties concurrently from different threads.

Sorry to be consistently picky but "can" and "will" are not synonyms

uklotzde:
> > Due to the race conditions of the current implementation the render and the control state ~will~[can] get out of sync when setting control state properties concurrently from different threads.
> 
> Sorry to be consistently picky but "can" and "will" are not synonyms

When talking about concurrency everything that *could* happen *will* happen.

orottier:
Alright, so to summarize:

Before we fixed the out-of-order arrival of node settings, we used atomic variables shared between the control and render thread. There was no issue with out of sync settings. All methods took `&self` which we considered a nice feature.

Then we moved to message passing for these settings. However we kept the atomics on the control side. As @uklotzde rightly points out, this could lead to a logic error when e.g. one of multiple control side threads are pre-empted right before updating the atomic variable. This is an oversight on my end, and could with small, but nonzero, chance end up with a very confusing state for end users

Now these errors are easily mitigated by taking `&mut` for these methods that update node settings. This is slightly inconvenient, but nothing out of the ordinary, for most usage patterns.

@b-ma do you foresee any issue with &mut for your nodejs bindings? If not, I think it is best to side with correctness over convenience. 

Our library has only one unique selling point over the bazillion other audio libs. That is strict adherence to a well known specifications. I don't think &mut will hurt this case, even though it might not be what the average JS developer would expect. Let's not hurt our long term potential by introducing possibilities for subtle logic bugs

orottier (approved):

b-ma:
Alright

> @b-ma do you foresee any issue with &mut for your nodejs bindings? If not, I think it is best to side with correctness over convenience.

No idea actually, must be tested...

But I would still argue that introducing the `mut` stuff is not very nice in terms of API ergonomics. 
If we don't care about the atomic (no real opinion on my side), what about using something like `Cell` or `RefCell` instead? 
Then we would just have the best of both worlds, correctness and nice API

uklotzde:
The API assumes strict eventual consistency between the control element (leader) and its corresponding render element (follower). Otherwise you will end up with inconsistent states.

Enclosing the loop state into a mutex would also work. This would be slightly opinionated but could preserve the implicit mutability feature of the API.

Sacrificing correctness for convenience is not an option. Never.

uklotzde:
...or `RefCell` if `Sync` is not required.

b-ma:
Final remark on my side, then I let you both decide what you think is best and I will stick to it, that's no big deal.

@uklotzde That's a _point of view_ to consider usability and appropriation by users as "convenience", that's not mine. That's why I'm _arguing_ to try to find, if possible, what I think would be a better solution for average developers / creative coders (potentially beginning with Rust) who, I think, are users we should really not forget to consider in this project (hardcore DSP guys for whom usability is not an issue will probably stick to C++ and JUCE for the years to come...)

uklotzde:
> Final remark on my side, then I let you both decide what you think is best and I will stick to it, that's no big deal.
> 
> @uklotzde That's a _point of view_ to consider usability and appropriation by users as "convenience", that's not mine. That's why I'm _arguing_ to try to find, if possible, what I think would be a better solution for average developers / creative coders (potentially beginning with Rust) who, I think, are users we should really not forget to consider in this project (hardcore DSP guys for whom usability is not an issue will probably stick to C++ and JUCE for the years to come...)

That's out of question. But then you need to pick the right implementation depending on the requirements (`Sync` or `!Sync`). Atomics don't work here.

uklotzde:
I have restored the internal mutability by using using `RefCell`. This is ok, because `AudioBufferSourceNode` was never `Sync` since it already uses `std::cell::OnceCell`. Therefore this PR should not change the public API.

PR #343, opened 2023-07-20
Fix bench workflow

b-ma:
ok that was not working properly... benches were triggered on every comment (so basically worse than the previous situation...)

So now
- benchmark workflow should only run if comment is "/bench"
- "post benchmark" is skipped if "benchmark" is not a success

Hopefully should be good now (these things are a real pain to test...)


orottier:
Seems to be working now :)

PR #345, opened 2023-07-21
Remove redundant 'static lifetime bounds

uklotzde:
`Any` already implies `'static'`.


orottier (approved):
Nice catch.
I had to read up on it because this was not familiar to me.
I believe technically, it's not `Any` but `Box<dyn ...>` which implies `'static`: https://doc.rust-lang.org/reference/lifetime-elision.html#default-trait-object-lifetimes

PR #346, opened 2023-07-21
fix: rename `AudioDestination` max_channels_count to max_channel_count

b-ma:
cf. https://webaudio.github.io/web-audio-api/#dom-audiodestinationnode-maxchannelcount

### DIFF
--- a/src/node/destination.rs
+++ b/src/node/destination.rs
@@ -104,6 +104,7 @@ impl AudioDestinationNode {
     }
     /// The maximum number of channels that the channelCount attribute can be set to (the max
     /// number of channels that the hardware is capable of supporting).
+    /// <https://www.w3.org/TR/webaudio/#dom-audiodestinationnode-maxchannelcount>
     pub fn max_channel_count(&self) -> usize {
         self.registration.context().base().max_channel_count()
     }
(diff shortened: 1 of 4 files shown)


src/node/destination.rs
```diff
@@ -104,7 +104,7 @@ impl AudioDestinationNode {
     }
     /// The maximum number of channels that the channelCount attribute can be set to (the max
     /// number of channels that the hardware is capable of supporting).
-    pub fn max_channels_count(&self) -> usize {
+    pub fn max_channel_count(&self) -> usize {
```
uklotzde, line 107: Please also add a [link](https://www.w3.org/TR/webaudio/#AudioDestinationNode-attributes) to the spec as discussed.

src/node/destination.rs
```diff
@@ -104,7 +104,7 @@ impl AudioDestinationNode {
     }
     /// The maximum number of channels that the channelCount attribute can be set to (the max
     /// number of channels that the hardware is capable of supporting).
```
orottier, line 106:
```suggestion
    /// number of channels that the hardware is capable of supporting).
    /// <https://www.w3.org/TR/webaudio/#dom-audiodestinationnode-maxchannelcount>
```

orottier (approved):
Thanks, nice catch

Issue #344, opened 2023-07-21
AudioNode ergonomics: atomics, interior mutability, Send/Sync?


orottier:
Continuing discussion from #342 

For historical reasons, many of the audio node properties are stored as atomics.
This results in all audio node methods taking `&self` instead of requiring `&mut self`.
As pointed out in the linked discussion, the usage of atomic may result in logic errors and should thus be rewritten.

Replacing the atomics by regular values will change the signature of all methods to `&mut`.
This is not bad per se, but is a breaking change and does require some rewrites on https://github.com/ircam-ismm/node-web-audio-api

Alternatively, we can leave the signatures unchanged but that will require to implement the interior mutability ourselves:
- via RefCell, which is low overhead but marks the nodes not being Send/Sync, which I think is a big disadvantage [1]
- via Mutex, which I think is very much overkill. Users that need to update the node from multiple thread can choose to do that themselves instead.

My preference would be to move all methods to `&mut` and make the necessary adjustments to the node bindings.

[1] this prevents e.g. a usage pattern where a thread is spawned to handle keyboard input which controls an AudioBufferSourceNode directly

orottier:
Apparently I already test some structs to be `Send + Sync + 'static` https://github.com/orottier/web-audio-api-rs/blob/main/tests/online.rs#L26
We should probably test this for all possible nodes, to prevent accidental regression (like the AudioBufferSource node using OnceCell instead of OnceLock)

b-ma:
I don't think the node bindings should be actually considered part of the discussion. 
Except very weird unexpected problem, that's probably just adding some `mut` here and there (and the probably related `clippy::please_shutup_about_non_required_mut` directive)

orottier:
Alright, then I think it is best to not use interior mutability and mark all methods `&mut` when required.
This is idiomatic, sound and will not be a burden in almost  all use cases.
The node bindings will need a slight update. Most nodes are currently stored for napi as `Rc<...Node>`. This will need to be `Rc<RefCell<...Node>>` and we will need a bunch of `borrow_mut` calls here and there.

I will start migrating some nodes to the new style

orottier:
Fixed with #350

PR #347, opened 2023-07-22
Ensure all concrete AudioNode types are 'static, Send and Sync

orottier:
While exploring #344 I noticed all audio nodes already adhere to this requirement, so I decide to bring the AudioBufferSourceNode in line by changing RefCell to Mutex and OnceCell to OnceLock

Please note that this is not an endorsement of the pattern "use a Mutex for interior mutability pattern so we can continue to use `&self`". It's just that I want to have this regression test in place before we refactor any further.

For context, I think it is important that the nodes fulfill these bounds, so the users never run into limitations when using the library.
I could imagine the following usage (pseudo code):
- create buffer source node
- connect to gain node
- connect to destination node
- ship gain node to a dedicated thread
- on that thread, listen for keyboard input to adjust the volume of the gain node

### DIFF
--- a/src/node/audio_buffer_source.rs
+++ b/src/node/audio_buffer_source.rs
@@ -1,5 +1,5 @@
 use std::any::Any;
-use std::cell::{OnceCell, RefCell};
+use std::sync::{Mutex, OnceLock};
 
 use crate::buffer::AudioBuffer;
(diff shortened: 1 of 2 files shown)


b-ma:
makes sens to me

PR #348, opened 2023-07-23
Add note on missing `AudioNodeOptions` in source and properly document `AudioDestination`

b-ma:


src/node/destination.rs
```diff
@@ -5,33 +5,36 @@ use super::{
     AudioNode, ChannelConfig, ChannelConfigOptions, ChannelCountMode, ChannelInterpretation,
 };
 
-/// Representing the final audio destination and is what the user will ultimately hear.
+/// The AudioDestinationNode interface represents the terminal node of an audio
+/// graph in a given context. usually the speakers of your device, or the node that
+/// will "record" the audio data with an OfflineAudioContext.
+///
+/// AudioDestinationNode has no output (no AudioNode can be linked after it in the
```
orottier, line 12: This is not true, I think you copied it from MDN but they have it wrong

The destination node has a single output.
https://webaudio.github.io/web-audio-api/#AudioDestinationNode
> The output of a AudioDestinationNode is produced by summing its input, allowing to capture the output of an AudioContext into, for example, a MediaStreamAudioDestinationNode, or a MediaRecorder (described in [mediastream-recording]).

b-ma: Huhu, I'm not really surprised, the explicit routing of multichannel is completely broken too (and Chrome and Firefox behave differently, didn't tested Safari but it seems we are the only ones who are compliant here :)

PR #351, opened 2023-07-27
Add value-like newtypes for offsets, durations, and sample rate

uklotzde:
Using unspecific value types like `f64` for everything is error prone and confusing. This could be improved by using type-safe abstractions, at least behind the scenes. It helps to reduce common errors even if the public API is required to remain loosely typed.

This explorational PR only implements the required subset that is needed for `AudioBufferSource`.

Drawback: Additional debug assertions within the wrappers invalidate the tests, see the comments in Cargo.toml. Either the implementation or the tests seem to be fragile and susceptible to rounding errors.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,3 +63,11 @@ cpal = ["dep:cpal"]
 cubeb = ["dep:cubeb"]
 cpal-jack = ["cpal", "cpal/jack"]
 cpal-asio = ["cpal", "cpal/asio"]
+
+[profile.test.package."web-audio-api"]
+# Debug assertions in arithmetic operations in the module `types` seem to prevent
+# register optimizations for floating point types (64-bit vs. 80-bit) that then
+# cause rounding/precision errors in tests.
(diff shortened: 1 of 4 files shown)


orottier:
Apologies for the delay in response. I will not be available a lot up to Sept 1 but will try to respond every now and then.

Thanks for the contribution. Programming with units of measurement is fun and indeed prevents certain classes of errors.

I think we should use units on all audio processors, or none of them.
The issue is that rewriting the full project to use units is a very large task.
Also my opinion is that using units is only a slight improvement over using raw values. While writing my own processors I have not had issues discerning the physical quantities.
Also,  classes of bugs such as mixing two values of the same unit, flipping signs etc are not prevented.
This really is also a black hole for effort. `detune` should be written with units as well. And probably we want `playback_rate` and `channel_index` to be type safe as well?

So in order to continue:
- are you up for the task of rewriting all processors? My help will be limited because I would rather focus on other issues that we track.
- can we agree on the list of variable types to be rewritten (e.g. only physical quantities? only quantities with dimension? also channel index, input port etc?)

PR #354, opened 2023-08-15
Expose current playhead position

uklotzde:
A proposed extension of the v2 API, see the doc comment for details.

Any serious playback application needs this essential feature and it is easy to implement. Each of the proposed workarounds for the v1 API is ugly and has shortcomings.


orottier (approved):
I'm fine with this, given the comment `/// Unofficial v2 API extension, not part of the spec yet.`
@b-ma do you have any opinion regarding this?

Issue #356, opened 2023-08-15
cpal: Unstable device identifiers


uklotzde:
The numbers returned when invoking `enumerate_devices_sync()` and in `is_valid_sink_id()` differ depending on the time of invocation. Later, when creating the `AudioContext`, the input devices seem to be missing and the output devices are renumbered. It is not possible to create an `AudioContext` by setting `sink_id` to one of the `device_id`s returned by `enumerate_devices_sync()` invoked before.

See also #355, that unfortunately doesn't fix this bug.

uklotzde:
Example:

```
[2023-08-15T22:39:58Z INFO  app] Found audio output device with sink id "5": "pipewire"
[2023-08-15T22:39:58Z INFO  app] Found audio output device with sink id "6": "default"
[2023-08-15T22:39:58Z INFO  app] Found audio output device with sink id "7": "sysdefault:CARD=PCH"
[2023-08-15T22:39:58Z INFO  app] Found audio output device with sink id "8": "hdmi:CARD=PCH,DEV=0"
[2023-08-15T22:39:58Z INFO  app] Found audio output device with sink id "9": "hdmi:CARD=PCH,DEV=1"
[2023-08-15T22:39:58Z INFO  app] Found audio output device with sink id "10": "hdmi:CARD=PCH,DEV=2"
[2023-08-15T22:39:58Z INFO  app] Found audio output device with sink id "11": "sysdefault:CARD=DJ"
[2023-08-15T22:39:58Z INFO  web_audio_api::context::online] ### Device: MediaDeviceInfo { device_id: "1", group_id: None, kind: AudioInput, label: "pipewire", device: Any { .. } }
[2023-08-15T22:39:58Z INFO  web_audio_api::context::online] ### Device: MediaDeviceInfo { device_id: "2", group_id: None, kind: AudioInput, label: "default", device: Any { .. } }
[2023-08-15T22:39:58Z INFO  web_audio_api::context::online] ### Device: MediaDeviceInfo { device_id: "3", group_id: None, kind: AudioInput, label: "sysdefault:CARD=PCH", device: Any { .. } }
[2023-08-15T22:39:58Z INFO  web_audio_api::context::online] ### Device: MediaDeviceInfo { device_id: "4", group_id: None, kind: AudioInput, label: "sysdefault:CARD=DJ", device: Any { .. } }
[...]

PR #360, opened 2023-09-06
Fix: improve device id

b-ma:
cf. #356

Note that I had to modify the pre-commit hook because for some reason cubeb build w/ CMake through clippy was failing, weird thing is that if I ran an example with the cubeb feature that was working...

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -64,7 +64,7 @@ repos:
             --locked,
             --workspace,
             --features,
-            cpal,
+            cubeb,
             --all-targets,
             --,
             -D,
@@ -83,7 +83,7 @@ repos:
             --locked,
             --workspace,
             --features,
-            cpal,
+            cubeb,
             --no-deps,
             --document-private-items,
           ]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -64,7 +64,7 @@ repos:
             --locked,
             --workspace,
             --features,
-            cubeb,
+            cpal,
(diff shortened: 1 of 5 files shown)


src/media_devices/mod.rs
```diff
@@ -32,6 +36,19 @@ pub enum MediaDeviceInfoKind {
...
 }
 
+pub(crate) fn create_device_id(
+    kind: u8,
+    host: String,
+    device_name: String,
+    num_channels: u16,
+    index: u8,
+) -> String {
+    let mut hasher = FxHasher::default();
```
orottier, line 46: Instead of manually hashing it is probably best to include a private struct with DeviceInfo and derive Hash on it

.pre-commit-config.yaml
```diff
@@ -64,7 +64,7 @@ repos:
             --locked,
             --workspace,
             --features,
-            cubeb,
+            cpal,
```
orottier, line 67: This should probably be reverted and investigated separately

src/media_devices/mod.rs
b-ma: ok done

.pre-commit-config.yaml
b-ma: The problem is that I was even not able to make a commit, because the pre-commit hook was crashing

orottier:
Thanks for the work!
I reverted the pre-commit thing locally, let me know if you still run into issues on your side

b-ma:
Ok cool, I'll let you know

PR #361, opened 2023-09-20
Fix and unify source/sink device_id selection in examples

orottier:
I think we had it mixed up in some cases

Use empty ('') for default device
Use 'none' to disable the output device (not for inputs)


b-ma:
...and probably some nonchalant copy / paste :), nice catch!

Issue #359, opened 2023-09-05
(de-)allocation in render thread


b-ma:
Follow up on #353, plus few ideas:

- [ ] Implement `Drop` to drop `AudioBuffer`s cleanly in AudioBufferSourceNode, ConvolverNode, ... (cf. [Otto's comment](https://github.com/orottier/web-audio-api-rs/pull/353#issuecomment-1705553704))
- [ ] Use `ArrayVec` instead of `Vec` in `AudioParam` (this is already a dependency so that's cool)
- [ ] Install https://github.com/Windfisch/rust-assert-no-alloc to make sure our render thread is clean

b-ma:
Hey,

Following the discusion here: https://github.com/orottier/web-audio-api-rs/pull/353#issuecomment-1705553704 , I had a look on how we could reuse the garbage collector thread to drop `AudioBuffer`s and co. outside render thread and here are my first impressions:
- If we want to be able to give some `sender` toward the garbage collector to the `AudioNode`s, it seem to mean that somehow the audio context should be able to deliver it, so that we grab a clone of the `sender` when we create the renderers.
- But with the current implement of the garbage collector, it does not seems possible to implement this pattern because `llq` is single producer single consumer channel, while we would need a multiple consumer single consumer channel.

What do you think? Should I try to proceed this way and replace the `llq` channel with a `crossbeam-channel`?

(aside, maybe we could even use raw `std::mpsc` instead of `crossbeam`, cf. https://github.com/rust-lang/rust/pull/93563)

orottier:
Thanks for sharing your thoughts.

The reason we are using `llq` is because it offers a channel of unbounded size, without (de)allocating in the render thread. This is not possible with crossbeam. The catch is that with `llq` we need to ship a pre-allocated node beforehand.

We could try to expose a method in the `RenderScope` to get a hold of the sender. I will need to time to sketch this out though. It's not trivial.

And maybe the `llq` message type should be enum, instead of `Box<dyn Any>`.

```rust
enum Garbage {
     Message(Box<dyn Any>),
     AudioBuffer(AudioBuffer),
     // etc
}

b-ma:
Ok, I see, probably it's best if I let you define the general architecture then

b-ma:
I have an idea about deallocation which I have the impression could solve several problems with rather low overhead. 
I didn't write any code yet as maybe I missed something, so I prefer having your opinion before digging into it.

[...]

PR #363, opened 2023-09-22
Perf - use `ArrayVec` instead of `Vec` for internal `AudioParam` buffer

b-ma:
cf. #359

### DIFF
--- a/src/param.rs
+++ b/src/param.rs
@@ -2,7 +2,9 @@
 use std::any::Any;
 use std::slice::{Iter, IterMut};
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::Arc;
+use std::sync::{Arc, OnceLock};
+
+use arrayvec::ArrayVec;
 
 use crate::context::AudioContextRegistration;
 use crate::node::{
@@ -10,8 +12,6 @@ use crate::node::{
 };
 use crate::render::{AudioParamValues, AudioProcessor, AudioRenderQuantum, RenderScope};
 use crate::{AtomicF32, RENDER_QUANTUM_SIZE};
-
-use std::sync::OnceLock;
 
 /// For SetTargetAtTime event, that theoretically cannot end, if the diff between
 /// the current value and the target is below this threshold, the value is set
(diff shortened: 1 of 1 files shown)


b-ma:
\bench

b-ma:
\bench

src/param.rs
```diff
@@ -1088,7 +1088,11 @@ impl AudioParamProcessor {
             match some_event {
                 None => {
                     if is_a_rate {
-                        self.buffer.resize(count, self.intrinsic_value);
+                        let buffer = [self.intrinsic_value; RENDER_QUANTUM_SIZE];
```
orottier, line 1091: Maybe it's more succinct to write as:

```rust
for _ in self.buffer.len() .. count {
    self.buffer.try_insert(self.intrinsic_value).unwrap();
}
```

Or did you benchmark this to be the fastest way?

orottier: Yeah, the L2 and RAM accesses are quite unstable. But since these numbers are so low I don't think it is very important (and that's probably also the reason for the large deviations). I think instruction count is the main metric to track.

I have read https://kobzol.github.io/rust/rustc/2023/09/23/rustc-runtime-benchmarks.html recently with some tips to look further into:

> After the initial refactoring was completed, I needed to decide how will we actually define the benchmarks and what tool we should use to gather the execution metrics. Both cargo bench and [criterion](https://github.com/bheisler/criterion.rs) are not a bad choice for running benchmarks, but they only measure wall-time, while I also wanted to measure hardware counters. I was considering to use [iai](https://github.com/bheisler/iai) for a while. However, it uses [Cachegrind](https://valgrind.org/docs/manual/cg-manual.html) for the measurements, while I wanted the benchmarks to be executed natively, without simulation. Also, using Cachegrind wouldn’t produce realistic wall-time results.

> In the end, I decided to write a small library called [benchlib](https://github.com/rust-lang/rustc-perf/tree/1d1400b66e232cd968c3598cb725c4133cea63a3/collector/benchlib), so that we would have ultimate control of defining, executing and measuring the benchmarks, instead of relying on external crates. benchlib uses Linux perf events to gather hardware metrics, using the [perf-event](https://github.com/jimblandy/perf-event) crate. I also took bits and pieces from other mentioned tools, like the [black_box](https://github.com/rust-lang/rustc-perf/blob/414230abc695bd715054d633c8203090f7423ce5/collector/benchlib/src/benchmark.rs#L161) function from iai.

PR #364, opened 2023-09-26
Prevent `ChannelConfig.clone` by changing `AudioRenderQuantum.add` args

orottier:


orottier:
Interesting, I guess the compiler is smarter than me.
I'm working on a rewrite of the ChannelConfig and this was something I though could be made faster. Ah well

PR #365, opened 2023-09-28
Propagate changes to node's ChannelConfig via the main message bus

orottier:
To prevent out of order arrival with other node settings.

For simplicity, I left the interior mutability pattern of the channel config but instead of using atomics, a mutex is used now. This should be fixed later to take &mut references to the node's channel config.

### DIFF
--- a/src/message.rs
+++ b/src/message.rs
@@ -3,7 +3,7 @@
 use std::any::Any;
 
 use crate::context::AudioNodeId;
-use crate::node::ChannelConfig;
+use crate::node::{ChannelConfigInner, ChannelCountMode, ChannelInterpretation};
 use crate::render::graph::Graph;
 use crate::render::AudioProcessor;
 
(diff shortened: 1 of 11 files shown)


src/message.rs
```diff
@@ -51,4 +51,19 @@ pub(crate) enum ControlMessage {
         id: AudioNodeId,
         msg: llq::Node<Box<dyn Any + Send>>,
     },
+
+    /// Update the channel count of a node
+    SetChannelCount { id: AudioNodeId, count: usize },
+
+    /// Update the channel count of a node
```
orottier, line 58:
```suggestion
    /// Update the channel count mode of a node
```

src/node/mod.rs
```diff
@@ -179,44 +179,70 @@ impl ChannelConfig {
...
     }
 
-    fn set_interpretation(&self, v: ChannelInterpretation) {
-        self.inner.interpretation.store(v as u32, Ordering::Release)
+    fn set_interpretation(
+        &self,
+        v: ChannelInterpretation,
+        registration: &AudioContextRegistration,
+    ) {
+        self.inner.lock().unwrap().interpretation = v;
```
orottier, line 206: We need to hold on to this lock while sending the control message and drop the lock at the end of the method.
This prevent an inconsistent state when the thread is pre-empted after altering the `interpretation` and a concurrent call is made to this method.

PR #366, opened 2023-10-05
Async deallocations

orottier:
Apologies for the messy proof of concept. Furthermore, this is untested. But I hope it gives some idea!
When I have a bit more time I will make it more performant, more readable and perhaps add testcases that actually validate no (de)allocations take place in the render thread*

* as an aside, there are some more possible (de)allocations we need to look at. Such as resizing of the main hashmap that holds the nodes, and the edge vectors, etc


orottier:
Continuing in #368

PR #370, opened 2023-10-11
Fix #369 - AudioListener would have a connection to a dropped PannerNode 

### LINKED ISSUE #369: PannerNode always crashes
Machine: AMD, 16 GB, Windows 10 Home, Rust 1.7, web-audio 0.33.0

I knew web audio from a javascript 3d development environment I made, so thank you for this project!  While working in rust, calling this anywhere in my code:

let panner = ctx.create_panner();

Will eventually cause (usually when it's called a second time):

thread 'cpal_wasapi_out' panicked at 'called `Option::unwrap()` on a `None` value', C:\Users\ggadwa\.cargo\registry\src\index.crates.io-6f17d22bba15001f\web-audio-api-0.33.0\src\render\graph.rs:261:14

Note the panner is not connected to an output or connected from an input source, just calling create_panner() causes this.  Note that if I instead use Panner::new with options the same thing happens, but more immediately (just when I call it the first time, or at least in my testing.)

Further notes: Creating a source, connecting it to an output, and playing it within the same code work perfectly; but within the same code creating a panner node will crash it, so it seems related specific to the panner node.

As I am new at Rust, it is possible this is some mistake on my part, i.e., a lifetime/de-allocation problem.

orottier:


tests/online.rs
```diff
@@ -121,3 +121,30 @@ fn test_channels() {
...
+    drop(panner);
+
+    // allow the audio render thread to boot and handle adding and dropping the panner
+    std::thread::sleep(std::time::Duration::from_millis(200));
+
+    // creating a new panner node should not crash the render thread
+    let mut _panner = context.create_panner();
+
+    // A crashed thread will not fail the test (only if the main thread panics).
+    // Instead inspect if there is progression of time in the audio context.
```
b-ma, line 145: damn tricky one...

Issue #371, opened 2023-10-12
Setting HRTF on panner node will freeze for 800 milliseconds


ggadwa:
Trying out 0.34, there's another oddity with the Panner Nodes (you'll be getting a lot of in a game testing from me!)  Again, thank you for this lib, and I'll try my best to shake out bugs for you.

If you do this:

`panner.set_panning_model(PanningModelType::HRTF);`

You get a pause (on my AMD/windows home, haven't tested it on my mac yet) or about 800 milliseconds.  My physics thread was complaining a lot about this :)

If you use ::EqualPower, it works fine.  BTW, no crashing on panner nodes anymore, so that issue has been fixed.

orottier:
Thanks for the report!

A small audio glitch is expected when the panning model is changed for an active panner, because HRTF for example might delay the sound signal for the left ear if the sound is coming from the right. PanningModel equalpower assumes both ears are at the same physical locations so no delays are present. 

However, 800ms is an eternity and this is not expected.

When I run `cargo run --release --example spatial` and press `<Enter>` a few times, I notice a 300ms delay on the control thread. Not the render thread by the way - audio playback is fine.

300 ms is quite a lot of time to load the HRTF data, but since this happens on the control thread it should not really be an issue either. I will have a look if is any speedup possible, but perhaps you should review your application and make sure that any interaction with our library is not happening on a thread with realtime constraints.

ggadwa:
So this one I might not be explaining properly.  I went back to chrome (I'll use this as reference form now on, as I should have) and did this:

`
console.info('start='+performance.now());
let panner = ctx.createPanner();
panner.panningModel='HRTF';
panner.distanceModel='linear';
panner.refDistance=100.0;
panner.maxDistance=1100.0;
panner.rolloffFactor=1.0;
panner.coneInnerAngle=360.0;
panner.coneOuterAngle=0.0;
panner.coneOuterGain=0.0;
panner.positionX.value=0.0;
panner.positionY.value=0.0;
panner.positionZ.value=0.0;
panner.orientationX.value=1.0;
panner.orientationY.value=0.0;
panner.orientationZ.value=0.0;
console.info('end='+performance.now());
`

In the console:

start=1658.9000000357628
test.html:52 end=1659.199999988079

[...]

PR #375, opened 2023-10-16
Improve HRTF panner initialization time by caching the HRIR sphere

orottier:
relates to #371


b-ma:
Cool thanks, as far as I can tell looks good to me!

PR #377, opened 2023-10-20
Flush denormal float values in audio graph processing

### LINKED ISSUE #194: How to deal with subnormal floats?

orottier:
Fixes #194


orottier:
Okay, with the final version I think the performance regression is acceptable

b-ma:
Cool looks nice ! 
I think there are still some places in the code (i.e. `AudioParam`) where we check this manually, I will try to have a look soon

(for info, I will be completely off next week)

PR #378, opened 2023-10-20
Derive Clone for AudioParam, could help the node bindings

orottier:
For motivation, see https://github.com/ircam-ismm/node-web-audio-api/pull/37

### DIFF
--- a/src/param.rs
+++ b/src/param.rs
@@ -201,7 +201,7 @@ impl AudioParamEventTimeline {
 }
 
 /// AudioParam controls an individual aspect of an AudioNode's functionality, such as volume.
-#[derive(Clone)]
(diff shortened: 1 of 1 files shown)


src/param.rs
```diff
@@ -201,8 +201,9 @@ impl AudioParamEventTimeline {
 }
 
 /// AudioParam controls an individual aspect of an AudioNode's functionality, such as volume.
+#[derive(Clone)]
```
orottier, line 204:
```suggestion
#[derive(Clone)] // for the node bindings, see #378
```

b-ma:
Thanks!

orottier:
Released in 0.36

Issue #208, opened 2022-08-13
Reduce `size_of` `Node` for better cache performance in render loop


orottier:
`std::mem::size_of::<Node>() == 160` currently. I think we gain some nice performance (better cache utilization) if we trim it down a bit.

Some suggestions:

[ ] Use `u8` instead of `usize` for input/output indices 
[ ] Use `u32` for `NodeIndex` instead of `u64`
[ ] Store `outgoing_edges` on the heap
[ ] Pack `ChannelConfig` as a single `Arc<(count, mode, interp)>`
[ ] Use bitflags to combine the boolean fields in a single u8

On a different note, we can explore to store the Nodes as an "struct of arrays" instead of "array (hashmap in our case) of structs" https://news.ycombinator.com/item?id=17981306

b-ma:
> On a different note, we can explore to store the Nodes as an "struct of arrays" instead of "array (hashmap in our case) of structs" https://news.ycombinator.com/item?id=17981306

Very nice direction I think, kind of related to this comment there https://github.com/orottier/web-audio-api-rs/issues/55#issuecomment-1206192591

orottier:
For the record, in the current implementation:

```rust
        assert_eq!(160, std::mem::size_of::<RefCell<Node>>());

        assert_eq!(8, std::mem::size_of::<RefCell<()>>());
        assert_eq!(8, std::mem::size_of::<Option<llq::Node<AudioNodeId>>>());
        assert_eq!(16, std::mem::size_of::<Box<dyn AudioProcessor>>());
        assert_eq!(24, std::mem::size_of::<Vec<AudioRenderQuantum>>()); // times 2
        assert_eq!(8, std::mem::size_of::<ChannelConfig>());
        assert_eq!(64, std::mem::size_of::<SmallVec<[OutgoingEdge; 2]>>());
        assert_eq!(1, std::mem::size_of::<bool>()); // times 3
```

This is expected to grow because we need to use more SmallVecs in #368

PR #379, opened 2023-10-22
Experiment: store nodes in contiguous Vec instead of HashMap

orottier:
Don't mind the unreadable code, I'm seeing +-10% performance gains in some benchmarks.

A big todo is to reuse deallocated node slots, or the Vec may grow indefinitely for long running processes

### DIFF
--- a/src/context/concrete_base.rs
+++ b/src/context/concrete_base.rs
@@ -21,14 +21,14 @@ use std::sync::{Arc, Mutex, RwLock, RwLockWriteGuard};
 ///
 /// It reuses the ids of decommissioned nodes to prevent unbounded growth of the audio graphs node
 /// list (which is stored in a Vec indexed by the AudioNodeId).
-struct AudioNodeIdIssuer {
+struct AudioNodeIdProvider {
     /// incrementing id
     id_inc: AtomicU64,
     /// receiver for decommissioned AudioNodeIds, which can be reused
     id_consumer: Mutex<llq::Consumer<AudioNodeId>>,
 }
 
-impl AudioNodeIdIssuer {
+impl AudioNodeIdProvider {
     fn new(id_consumer: llq::Consumer<AudioNodeId>) -> Self {
         Self {
             id_inc: AtomicU64::new(0),
@@ -36,7 +36,7 @@ impl AudioNodeIdIssuer {
         }
     }
 
-    fn issue(&self) -> AudioNodeId {
+    fn get(&self) -> AudioNodeId {
         if let Some(available_id) = self.id_consumer.lock().unwrap().pop() {
             llq::Node::into_inner(available_id)
         } else {
@@ -75,8 +75,8 @@ struct ConcreteBaseAudioContextInner {
     sample_rate: f32,
     /// max number of speaker output channels
     max_channel_count: usize,
-    /// issuer for new AudioNodeIds
-    audio_node_id_issuer: AudioNodeIdIssuer,
+    /// provider for new AudioNodeIds
+    audio_node_id_provider: AudioNodeIdProvider,
     /// destination node's current channel count
     destination_channel_config: ChannelConfig,
     /// message channel from control to render thread
@@ -112,7 +112,7 @@ impl BaseAudioContext for ConcreteBaseAudioContext {
         f: F,
     ) -> T {
         // create a unique id for this node
-        let id = self.inner.audio_node_id_issuer.issue();
+        let id = self.inner.audio_node_id_provider.get();
         let registration = AudioContextRegistration {
             id,
             context: self.clone(),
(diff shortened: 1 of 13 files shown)


src/render/node_collection.rs
```diff
@@ -0,0 +1,105 @@
...
+use std::ops::{Index, IndexMut};
+
+pub struct NodeCollection {
+    nodes: Vec<Option<RefCell<Node>>>,
+}
+
+impl NodeCollection {
+    pub fn new() -> Self {
+        let mut instance = Self {
+            nodes: Vec::with_capacity(64),
```
b-ma, line 12: I wonder if we should be more agressive here and grab more memory at startup here, e.g. `Vec::with_capacity(2048)`

src/render/graph.rs
```diff
@@ -126,6 +125,7 @@ impl Graph {
         let inputs = vec![AudioRenderQuantum::from(self.alloc.silence()); number_of_inputs];
         let outputs = vec![AudioRenderQuantum::from(self.alloc.silence()); number_of_outputs];
 
+        let index = index.0 as usize;
```
b-ma, line 128: Hey, really interesting perf improvements
Just thinking (maybe I'm mistaking, not really sure and maybe that's what you have in mind but...) renaming these to `id` rather than `index` could help continue refactoring I think. 
Because then we could have two parallel `Vec`: a `Vec<Option<AudioNodeId>>` and a `Vec<Option<RefCell<Node>>>` where the indexes would correspond. Then we could reuse `None` places in the `Vec` to prevent the it from growing on indefinitely (would maybe imply a third `Vec<Option<usize>>` containing the dropped positions for housekeeping)

orottier: Hey thanks for the suggestion. I actually have another solution in mind where the audio graph lets the AudioContext know ( realtime safe) which AudioNodeIds are available again. Would be interesting to benchmark against your solution

orottier:
I will have to evaluate the performance regression between the first and second commit

src/render/node_collection.rs
orottier: `std::mem::size_of::<RefCell<Node>>()` is 160 bytes, so that would mean over 300kB allocated.
That's not terrible of course, but I think most audio graphs really don't need that many nodes.

Whatever the initial size, we do need to think about the performance impact of outgrowing that size.
- for the reallocation, we could perform it on the control thread because it will know when we run out of space
- for the memcpy, we should maybe look at reducing the side of `Node` #208 or an amortized growth like https://github.com/jonhoo/griddle

orottier:
Alright, very happy with the results. This is ready for review @b-ma, but take your time. I'm not in a hurry

src/render/graph.rs
```diff
@@ -142,9 +148,7 @@ impl Graph {
     }
 
     pub fn add_edge(&mut self, source: (AudioNodeId, usize), dest: (AudioNodeId, usize)) {
-        self.nodes
-            .get_mut(&source.0)
-            .unwrap_or_else(|| panic!("cannot connect {:?} to {:?}", source, dest))
+        self.nodes[source.0 .0 as usize]
```
b-ma, line 151: I'm a bit confused with this syntax: `source.0 .0 as usize`, what does it mean?

orottier: Yeah that looks weird. It is a double tuple unpack for which rustfmt has added an extra space. This appears to be not necessary anymore if we change the config: https://github.com/rust-lang/rustfmt/issues/5745 
I will pick that up in a separate PR

Then again I agree the nested tuple is not nice to read. I will change the NodeCollection interface to take AudioNodeId instead of usize for indexing. This will make the code nicer!

b-ma: > I will change the NodeCollection interface to take AudioNodeId instead of usize for indexing. This will make the code nicer!

+1

b-ma:
I think I got it, no special comment on my side, looks very nice, great work!

src/context/concrete_base.rs
```diff
@@ -17,6 +17,34 @@ use crossbeam_channel::{Receiver, SendError, Sender};
 use std::sync::atomic::{AtomicU64, AtomicU8, Ordering};
 use std::sync::{Arc, Mutex, RwLock, RwLockWriteGuard};
 
+/// This struct assigns new [`AudioNodeId`]s for [`AudioNode`]s
+///
+/// It reuses the ids of decommissioned nodes to prevent unbounded growth of the audio graphs node
+/// list (which is stored in a Vec indexed by the AudioNodeId).
+struct AudioNodeIdIssuer {
```
b-ma, line 24: maybe rename to `AudioNodeIdProvider`? 
I find `audio_node_id_issuer.issue()` a bit weird (maybe `audio_node_id_provider.get()`? But really no big deal

PR #380, opened 2023-11-02
Run benchmarks locally with criterion

### LINKED ISSUE #188: Setup proper benchmark suite

orottier:
Fixes #188 

You can now run `cargo bench` locally to run the same benches as we run in CI.
It's a lot of boilerplate currently, I'm still looking for improvements.
Also we need to move to unify  examples/benchmarks.rs in here.

### DIFF
--- a/benches/my_benchmark.rs
+++ b/benches/my_benchmark.rs
@@ -26,7 +26,6 @@ fn get_audio_buffer(ctx: &OfflineAudioContext) -> AudioBuffer {
     static BUFFER: OnceLock<AudioBuffer> = OnceLock::new();
     BUFFER
         .get_or_init(|| {
-            println!("decoding now");
             let file = File::open("samples/think-stereo-48000.wav").unwrap();
             ctx.decode_audio_data_sync(file).unwrap()
         })
--- a/benches/my_benchmark.rs
+++ b/benches/my_benchmark.rs
@@ -7,14 +7,42 @@ use criterion::{black_box, criterion_group, criterion_main, Criterion};
 use web_audio_api::context::BaseAudioContext;
 use web_audio_api::context::OfflineAudioContext;
 use web_audio_api::node::{AudioNode, AudioScheduledSourceNode, PanningModelType};
+use web_audio_api::AudioBuffer;
+
+use std::fs::File;
+use std::sync::OnceLock;
 
 const SAMPLE_RATE: f32 = 48000.;
 const DURATION: usize = 10;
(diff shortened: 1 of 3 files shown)


benches/my_benchmark.rs
```diff
@@ -214,11 +237,13 @@ pub fn bench_hrtf_panners() {
     osc.connect(&panner2);
     osc.start();
 
-    assert_eq!(ctx.start_rendering_sync().length(), SAMPLES);
+    assert_eq!(ctx.start_rendering_sync().length(), SAMPLES_SHORT);
 }
 
+#[cfg(feature = "iai")]
```
b-ma, line 243: maybe a macro would be justified to generate the code from here to the bottom of the file

b-ma:
Cool!

> Also we need to move to unify examples/benchmarks.rs in here.

Do you want to port all the benches? If yes I can do it once you are satisfied with the general layout (...I would like to have more time to contribute, but at least I can do the small things :)

In any case, I think it's nice to keep `examples/benchmarks.rs` as it allows to compare with browser's perfs (even if it's quite noisy)

orottier:
> Do you want to port all the benches? 

Yes and no.
Yes, because criterion helps us to accurately run these benches for academic publishing. No because some of the benchmarks are way too heavy to run often and in CI. Maybe there's some middle ground?

In any case the current CI benches are not bad. Although I would like to have some coverage on resampling, upmixing and granular synth.

Any help is appreciated, maybe you can help with the macro's? and maybe copy over 3 or 4 of the lightweight benches from examples/ to benches/?

Another interesting observation is that `iai` and `criterion` don't seem to agree always. Both agreed that #379 was a performance win, but `criterion` tells me that #279 is also beneficial, whereas `iai` reports +10% instruction count.. To be investigated

orottier:
> https://github.com/orottier/web-audio-api-rs/pull/380#issuecomment-1790387222

hah, you triggered the bench action with your previous comment.
It ran criterion in CI because I believe the bench command (which now needs the `--features iai` flag) is invoked from the main branch but is then run on the local checkout. This should be fixed after merging the PR; CI should run iai

PR #381, opened 2023-11-03
refactor: move benches boilerplate code into macro

b-ma:
Not completely sure how this works, but it seems to works :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ criterion = "0.5.1"
 env_logger = "0.10"
 iai = "0.1.1"
 rand = "0.8"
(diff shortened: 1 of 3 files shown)


b-ma:
Arg nop...

orottier:
Looking nice already. If you need me to move some things to main already for further testing, let me know

b-ma:
I think the changes in https://github.com/orottier/web-audio-api-rs/pull/381/commits/1f140d4d65eb7a5a5c76427a6fd679ac13b08b19 needs to be put on main so we can run the benches here and make sure everything is ok. Except that I think it's ready to merge, I will try to add some new benches in a separate PR

PR #382, opened 2023-11-04
Refactor param automation mess

b-ma:
Hey @orottier,

Here is an attempt for cleaning the huge messy `match` in `AudioParamProcessor::compute_buffer` (which has been boring me since a long time...). It's not finished at all, I just rewrote the `SetValue` stuff for now, I prefer to have your feedback before spending more time on it.

What do you think, should I go on?

(Maybe that could help also with perf by minimizing cache, I don't really know)

### DIFF
--- a/src/param.rs
+++ b/src/param.rs
@@ -87,7 +87,7 @@ pub struct AudioParamDescriptor {
     pub max_value: f32,
 }
 
-#[derive(PartialEq, Eq, Debug)]
+#[derive(PartialEq, Eq, Debug, Copy, Clone)]
 enum AudioParamEventType {
     SetValue,
     SetValueAtTime,
@@ -128,7 +128,7 @@ struct AudioParamEventTimeline {
 impl AudioParamEventTimeline {
     fn new() -> Self {
         Self {
-            inner: Vec::new(),
+            inner: Vec::with_capacity(32),
             dirty: false,
         }
     }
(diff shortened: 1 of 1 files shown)


src/param.rs
```diff
@@ -1075,22 +1131,34 @@ impl AudioParamProcessor {
...
             self.buffer.push(self.intrinsic_value);
             // nothing to compute in timeline, for both k-rate and a-rate
             if is_constant_block {
                 return;
             }
         }
 
+        // For borrow check reasons, we move move out the `event_timeline`,
+        // and pass it around as &mut.
+        let mut event_timeline = std::mem::take(&mut self.event_timeline);
```
b-ma, line 1144: Just used the same strategy as in graph render to workaround the borrow checker. Maybe there is a more simple solution but I couldn't find it...

b-ma: Also, `EventTimeline` contains a `Vec`, do it allocates when replacing with default?

b-ma: Hum, I guess this is my answer: https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-Default-for-Vec%3CT,+Global%3E

orottier:
Hey, definitely anything that reduces the `match` method complexity is fine with me.

We are maybe able to make the pattern you are suggesting a bit more elegant. I don't think you need to take out the `&mut event_timeline` to work around the borrow checker

If you make the EventType a Copy-type:
```
#[derive(PartialEq, Eq, Debug, Copy, Clone)]
enum AudioParamEventType {
```

And start the big match with
```
        loop {
            let next_event_type = self.event_timeline.peek().map(|e| e.event_type);

            match next_event_type {
                None => {
// snip
                    break;
                }
                Some(AudioParamEventType::SetValue) | Some(AudioParamEventType::SetValueAtTime) => {
                    if self.compute_set_value_automation(&block_infos) {
                        break;
                    }
                }
               // etc
```

A drawback with your suggested solution is that we `peek` twice for every handled event, but maybe that is acceptable

b-ma:
Ok thanks, your method worked like a charm!

Pretty happy this is cleaned up :)

orottier:
Tests pass, code is more readable. Great work

PR #385, opened 2023-11-09
AudioWorkletNode

orottier:
Hey, this is already getting out of hand so I wanted to show you the current progress.

I think mainly the `examples/worklet.rs` file would be interesting to inspect. It should resemble the spec way more than the AudioProcessor trait we used before.

Lots of work to do:
- [x] support multiple inputs/outputs
- [x] support outputChannelCount
- [x] construct the processor inside the render thread and support constructor options
- [x] remove the `Send` bound from the trait
- [x] support scope (current time, frame , sample rate)
- [ ] message port
- [x] AudioParamDescriptor should contain `name: String`
- [x] Add bitcrusher example 
- [x] decide mod layout
- [x] documentation
- [ ] rename RenderScope to AudioWorkletGlobalScope ?
- [ ] decide: hide the raw AudioProcessor trait?
- [x] benchmarking


examples/worklet.rs
```diff
@@ -1,161 +1,58 @@
...
-
-impl AudioProcessor for WhiteNoiseProcessor {
-    fn process(
+    fn process<'a, 'b>(
         &mut self,
-        _inputs: &[AudioRenderQuantum],
-        outputs: &mut [AudioRenderQuantum],
-        params: AudioParamValues<'_>,
-        _scope: &RenderScope,
+        inputs: &'b [&'a [f32]],
```
orottier, line 31: These different lifetimes are necessary unfortunately

test.rs
```diff
@@ -0,0 +1,21 @@
+trait AsMeh {
```
orottier, line 1: This experiment could help with https://webaudio.github.io/web-audio-api/#AudioWorkletProcessor-instantiation

b-ma:
Hey, I'll try to have a close look in the next few days

b-ma:
It looks quite nice indeed! Great work

I read the spec and the code quite a few time (...but probably not enough :) and here are a few questions / remarks:
- rename `AudioWorkletProcessor::construct` to `AudioWorkletProcessor::constructor`?
- rename `ConstructorOptions` or `ProcessorOptions`

So if I understand well you consider the instantiation should occur in render thread (which is confirmed by the spec by the way) because you think processors should be only stack allocated and thus deallocation should be free? But I still don't really get how this new API would help to solve our issues with e.g. delay ring buffer (but maybe that's not the point yet...)

If I'm not missing something the _bitcrusher_ example could already be implemented no? I can try to do it if you want, would be an interesting test maybe. 
The [_vumeter_](https://webaudio.github.io/web-audio-api/#vu-meter-mode) example would also be interesting to implement at some point to test message ports.

In any case, that's really interesting! Let me know if I can help in any way (maybe porting existing nodes see were it get's us?)

examples/worklet.rs
b-ma: A bit odd/complex indeed, but not really shocking neither in my opinion as this is boilerplate code... 

Do you think it is due to the fact that you need to deal with the existing `AudioWorkletRenderer` so this could potentially be removed once (and if) all nodes are ported to this new API, or is it something we will have to live with forever (with an additional depth when several inputs/outputs are supported) in any case?

src/node/worklet.rs
```diff
@@ -0,0 +1,321 @@
...
+
+impl AudioProcessor for AudioWorkletRenderer {
+    fn process(
+        &mut self,
+        inputs: &[AudioRenderQuantum],
+        outputs: &mut [AudioRenderQuantum],
+        params: crate::render::AudioParamValues<'_>,
+        scope: &RenderScope,
+    ) -> bool {
+        // Bear with me, to construct a &[&[&[f32]]] we first build a backing vector of all the
```
orottier, line 223: To get rid of this allocation in safe rust is going to be tricky. But I will try a few more things

orottier:
```
bench_sine_gain
  Instructions:            77498406
bench_sine_gain_with_worklet
  Instructions:            78255322
```

I'm happy to see the new worklet overhead is quite low compared to the processors that handle the RenderQuantum raw.
I have added unsafe code to prevent reallocations per render quantum. I might ask for an unsafe review on the rust user forum though, it's quite tricky.

orottier:
Are you okay with this mod layout?
I'm leaving the message port for now, that can be added later.
How do you feel about `rename RenderScope to AudioWorkletGlobalScope ?`

b-ma:
> I'm happy to see the new worklet overhead is quite low compared to the processors that handle the RenderQuantum raw.

Yup quite impressive indeed, congrats! (I would be quite curious to test this with other nodes as well at some point)

> Are you okay with this mod layout?

I think I would simply put it in `web_audio_api::node`, as finally that's just some kind of `AudioNode` too:
```
interface AudioWorkletNode : AudioNode 
```
But no strong opinion here

> How do you feel about rename RenderScope to AudioWorkletGlobalScope ?

On the one side I don't feel it has exactly the same responsibility as in the spec. But on the other side I don't think we can really have such "scope" behavior in Rust (?). So, maybe that's the closest thing we can provide and somehow it makes sens...

Maybe that's just a matter of properly documenting the tradeoff with the spec? I think you read the spec more than me here, what your guts are telling you? :)

Another thing I'm a bit unsure is the argument order for:

```rs
   fn process<'a, 'b>(
        &mut self,
        scope: &'b RenderScope,
        inputs: &'b [&'a [&'a [f32]]],
        outputs: &'b mut [&'a mut [&'a mut [f32]]],
        params: AudioParamValues<'b>,
    ) -> bool {
```
Why did you decided to put `score` as first argument here rather than as last one? I would make more sens to me to have it at the end I guess, as it is kind of an "extension" of the `AudioWorkletProcessCallback` API

orottier:
> I think I would simply put it in web_audio_api::node, as finally that's just some kind of AudioNode too:

But it also contains `AudioParamValues` and the `AudioWorkletProcessor` interface

> Why did you decided to put scope as first argument here rather than as last one?

I haven't put much thought in it, but it felt like an extension of `self`. I will move it to the end though, good suggestion

b-ma:
> > I think I would simply put it in web_audio_api::node, as finally that's just some kind of AudioNode too:

> But it also contains AudioParamValues and the AudioWorkletProcessor interface

Hum right, then let's go with `web_audio_api::worklet`, that makes sens

orottier:
I have decided to merge this current work and leave the leftovers for a new issue #391

Issue #389, opened 2023-11-18
Bug: `thread '<unnamed>' panicked at 'Index 0 for dropped Node in NodeCollection'`


b-ma:
I run this quite a few time `RUST_BACKTRACE=1 cargo run --example granular`

Got that:

```
thread '<unnamed>' panicked at 'Index 0 for dropped Node in NodeCollection', src/render/node_collection.rs:104:32
stack backtrace:
   0: rust_begin_unwind
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/panicking.rs:67:14
   2: <web_audio_api::render::node_collection::NodeCollection as core::ops::index::IndexMut<web_audio_api::context::AudioNodeId>>::index_mut::{{closure}}
             at ./src/render/node_collection.rs:104:32
   3: core::option::Option<T>::unwrap_or_else
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/option.rs:979:21
   4: <web_audio_api::render::node_collection::NodeCollection as core::ops::index::IndexMut<web_audio_api::context::AudioNodeId>>::index_mut
             at ./src/render/node_collection.rs:100:9
   5: web_audio_api::render::graph::Graph::render
             at ./src/render/graph.rs:483:10
   6: web_audio_api::render::thread::RenderThread::render_inner
             at ./src/render/thread.rs:299:42
   7: web_audio_api::render::thread::RenderThread::render::{{closure}}
             at ./src/render/thread.rs:220:39
   8: no_denormals::no_denormals
             at /Users/me/.cargo/registry/src/index.crates.io-6f17d22bba15001f/no_denormals-0.1.2/src/lib.rs:102:12
   9: web_audio_api::render::thread::RenderThread::render
             at ./src/render/thread.rs:220:9
  10: web_audio_api::io::cpal::spawn_output_stream::{{closure}}
             at ./src/io/cpal.rs:481:17
  11: cpal::traits::DeviceTrait::build_output_stream::{{closure}}
             at /Users/me/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cpal-0.15.2/src/traits.rs:169:17
  12: cpal::host::coreaudio::macos::Device::build_output_stream_raw::{{closure}}
[...]

PR #390, opened 2023-11-18
fix: make sure destination is in node collection to consider graph active

b-ma:
cf. https://github.com/orottier/web-audio-api-rs/issues/389


orottier:
Great work

PR #392, opened 2023-11-18
Add a note on the lack of WebAssembly support in README.md

orottier:
Do you think it's worth it?


b-ma:
It seems to be something that several people expect, so yes this probably a good idea

Not sure that should go into the `Audio Backend` section though, as this is more a compilation target issue no? 
Maybe we could introduce a `MISC` section, e.g.

```md
## MISC

### Node.js bindings

...

### WebAssembly 

Compiling to WebAssembly / [web-sys](https://rustwasm.github.io/wasm-bindgen/api/web_sys/) is currently not supported. 
Please read up on the [wasmtracking issue](https://github.com/orottier/web-audio-api-rs/issues/265) for alternative solutions.

### Note for Linux users

...
```

Issue #386, opened 2023-11-11
PannerNode behaves differently on the web


cybersoulK:
The sound is higher in the web when the position is farther away, in a factor of 3-5 times.
This is the setup code:

Web:
```rust
        let panner_node = context.create_panner().unwrap();
        panner_node.set_distance_model(DistanceModelType::Inverse);
        panner_node.set_panning_model(PanningModelType::Hrtf);
        panner_node.set_ref_distance(8.0);
        panner_node.set_max_distance(80.0);
        panner_node.set_rolloff_factor(0.2);
```

web-audio-api:
```rust
let mut options = PannerOptions::default();
        options.distance_model = DistanceModelType::Inverse;
        options.panning_model = PanningModelType::HRTF;
        options.ref_distance = 8.0;
        options.max_distance = 80.0;
        options.rolloff_factor = 0.2;
        let panner_node = PannerNode::new(context, options);
 ```
 
 might be related to the following issue when he noticed some differences as well:
 https://github.com/orottier/web-audio-api-rs/issues/373

**Comment by orottier** [comment]
Thanks for the report. What value do you mean with farther away?
The expected behaviour is 100% volume for distance <= 8 and then rolling off when moving further away.
For your information, `max_distance` is unused in the "inverse" distance model

**Comment by cybersoulK** [comment]
I am about 60 distance,
I might have exaggerated with 3x, but i also updated parameters since:

```rust
panner_node.set_distance_model(DistanceModelType::Inverse);
panner_node.set_panning_model(PanningModelType::Hrtf);
panner_node.set_ref_distance(*distances.start() as f64);
panner_node.set_rolloff_factor(1.0);
```

```rust
panner_node.set_distance_model(DistanceModelType::Inverse);
panner_node.set_panning_model(PanningModelType::Hrtf);
panner_node.set_ref_distance(8.0);
panner_node.set_rolloff_factor(1.0);
```
```rust
options.distance_model = DistanceModelType::Inverse;
options.panning_model = PanningModelType::HRTF;
options.ref_distance = 8.0;
options.rolloff_factor = 1.0;
```

(volume is 0.8, distance is around 40)
the first section of the video shows the web-audio-api, second part shows web. 
Pawn is at exact position, the bullet does not have a PannerNode (it's the only difference), so the sound is the reference point.
A looped sound is running to show the PannerNode differences, the sound notices around 30% higher volume in the web? or i am not sure what the difference is.

[...]

Issue #241, opened 2022-11-22
Handle stereo input in PannerNode


orottier:
Both the equalpower and hrtf panning mode assume mono input

PR #393, opened 2023-11-19
PannerNode fixes: volume, stereo-to-stereo and clipping of rolloff

orottier:
Relates to #386 #241


src/node/panner.rs
```diff
@@ -734,51 +759,30 @@ impl AudioProcessor for PannerRenderer {
                 projected_source = [0., 0., 1.];
             }
 
+            // Only handle mono for now (todo issue #44)
+            *output = input.clone();
+            let mut overall_gain_correction = 1.;
+            if output.number_of_channels() == 2 {
+                overall_gain_correction *= 2.; // stereo-to-stereo panning typically doubles volume
```
b-ma, line 766:
```rs
overall_gain_correction = 2.; 
```

This is bit confusing and appear arbitrary, would be nice to have a bit more explanation

b-ma:
Looks good to me! (...as far as not a big user of this node can tell :)

PR #394, opened 2023-11-19
Fix crashed control thread when communicating with dropped render thread

### LINKED ISSUE #387: Panic when shutting down windowing by close button
```rust
panicked at 'called `Result::unwrap()` on an `Err` value: "SendError(..)"', /Users/.../.cargo/registry/src/index.crates.io-6f17d22bba15001f/web-audio-api-0.36.1/src/context/concrete_base.rs:415:40

panicked at 'called `Result::unwrap()` on an `Err` value: "SendError(..)"', /Users/.../.cargo/registry/src/index.crates.io-6f17d22bba15001f/web-audio-api-0.36.1/src/context/concrete_base.rs:415:40

panicked at 'panic in a function that cannot unwind', library/core/src/panicking.rs:126:5
thread caused non-unwinding panic. aborting.
```

this happens when i shutdown the winit windowing by the close UI button, 
but it does not happen when i close the terminal by Ctrl + C

might wanna use?
```rust
self.send_control_msg(message).ok();
```

orottier:
Fixes #387

### DIFF
--- a/src/context/concrete_base.rs
+++ b/src/context/concrete_base.rs
@@ -137,7 +137,7 @@ impl BaseAudioContext for ConcreteBaseAudioContext {
                 self.inner.queued_audio_listener_msgs.lock().unwrap();
             queued_audio_listener_msgs.push(message);
         } else {
-            self.send_control_msg(message).unwrap();
+            self.send_control_msg(message);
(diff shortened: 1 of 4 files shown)


b-ma:
Nice, looks good!

PR #395, opened 2023-11-20
Try to prevent pre-commit hook failing on `cubeb`

### LINKED ISSUE #383: cargo-doc precommit fails if the lib has not been build with cubeb previously
Not sure what is going on but, add this issue after `rm -Rf target` then tried to make a commit
The pre-commit hook passed after I did `cargo build --features cubeb`

orottier:
If we run cargo check first, cubeb should be built properly and can be used for the next steps.

Might fix #383


orottier:
Merging as this won't hurt either way

Issue #396, opened 2023-11-21
Audio starts jittering the more sounds are played


cybersoulK:
after a few hundred source nodes connect and disconnect, there are a lot of loud pops or sound jitter. / abrupt interruptions (some leak in web-audio-api or cpal usage?)

[pop.mp3.zip](https://github.com/orottier/web-audio-api-rs/files/13377884/pop.mp3.zip) (internal recording of the bullets sounds, 

- Cpal backend is used
- Web equivalent does well
- One buffer per source node. (Cloned)
- A Node graph is created (and later disconnected) per sound: Source -> gain -> destination
- None of these nodes are connected with anything else.
- This issues happens on both Mac and windows

Render capacity shows increasing numbers:
<img width="1069" alt="Screenshot 2023-11-21 at 2 41 38 AM" src="https://github.com/orottier/web-audio-api-rs/assets/102260392/39d2beb1-e424-4d0d-b6ea-1d630f3a66b3">
<img width="1088" alt="Screenshot 2023-11-21 at 2 39 54 AM" src="https://github.com/orottier/web-audio-api-rs/assets/102260392/8c0a850d-9962-4454-8c59-4f547b0c195e">
<img width="1104" alt="Screenshot 2023-11-21 at 2 41 01 AM" src="https://github.com/orottier/web-audio-api-rs/assets/102260392/af8df810-35bf-459e-9a37-5cda4c664560">

I used kira.rs with the cpal backend, and it never did this.
I just compared them side by side with exact same sound conditions.

My minimal code for each sound that plays:
```rust
 let options = AudioBufferSourceOptions::default();
 let mut source_node = AudioBufferSourceNode::new(context, options);
source_node.set_buffer(buffer);

let mut options = GainOptions::default();
options.gain = volume;
let gain_node = GainNode::new(context, options);

source_node.connect(&gain_node);
gain_node.connect(&context.destination());

source_node.start();

Sound {
   source_node,
   gain_node,
}
```
and later when the sound stops:
```rust
impl Drop for Sound {
   self.source_node.disconnect();
   self.gain_node.disconnect();
}
```

I made sure that all of the sounds are being dropped and their nodes disconnected

orottier:
Thanks for the detailed report.
I can confirm on my machine that we run into max load with about 500 sounds buffersourcenodes+gain concurrently and adding 10 nodes in bursts every 5 milliseconds.
[...]

PR #399, opened 2023-11-22
How to starve cpu 

b-ma:
Not meant to be merged, but maybe a test bed regarding #396


orottier:
Next objective, starve CPU with detached Oscillator --> Gain legs

Issue #397, opened 2023-11-22
Free audio processor resources when the node has stopped/ended and cannot be started again


orottier:
Currently, we drop the AudioProcessor when both:
- the control thread has dropped the corresponding AudioNode
- the processor has no inputs and tail-time = false

But we can release the resources earlier, e.g. when
- an AudioBufferSourceNode has ended and does not loop
- a ConstanceSourceNode has stopped

Take care, since the control thread still has the AudioNode they could call connect/disconnect/etc on it.
This means we should not recycle the audio_node_id immediately

b-ma:
Cool to re-open this discussion, I really think we have a lot of room for perf improvements in handling the audio graph more smartly (...but that's quite/really tricky for sure...)

> an AudioBufferSourceNode has ended and does not loop

I don't really get that, if the `AudioBufferSourceNode` is in a loop, it's not ended and the tail time should remain `true` unless `stop` is called explicitly (and if it is not the case, I would consider it an issue ...and my fault actually :)

> a `ConstantSourceNode` has stopped

This is true for all `AudioScheduledSourceNode` I think

--- 

Related note regarding #396 and `AudioBufferSourceNode`, because I'm not really sure about what we do in the case:
- We have a buffer of 10mn long
- We start the source
- After 1sec, we `disconnect()` the node and drop it _**but** without calling `stop()` explicitly_...

Do the renderer continue to hardly compute things that cannot be heard anymore? (I'm actually afraid this is the case, and maybe that could explain the issue)

And we could just imagine the exact same setup with an `OscillatorNode` or a `ConstantSourceNode`, which would be actually worse because they would render forever (as `tail_time == true` unless `stop` is explicitly called on these nodes...)

If I'm not mistaking in the first place, maybe a new rule for `AudioScheduledSourceNode`s could be: "if disconnected and dropped on main thread (so no possibility to reconnect anymore), tail time is `false` whatever the renderer pretends"

And then (not sure) maybe this could be generalized to:

> We drop the AudioProcessor **_also_** when both:
> + the control thread has dropped the corresponding AudioNode
> + the processor has no output

orottier:
Yes, all valid points.

> Do the renderer continue to hardly compute things that cannot be heard anymore? 

Reading the spec literally, yes it should continue processing regardless if the output are used.
https://webaudio.github.io/web-audio-api/#rendering-loop - Step 4 requires to consider all nodes

[...]

PR #400, opened 2023-11-23
Release processor resources for 'orphaned' nodes (even with tail time)

orottier:
Relates to #397 and should solve problem of #399

Note that this does not solve for orphaned legs (e.g. oscillator -> gain) but we will deal with that later

### DIFF
--- a/src/render/graph.rs
+++ b/src/render/graph.rs
@@ -174,12 +174,17 @@ impl Graph {
     }
 
     pub fn remove_edges_from(&mut self, source: AudioNodeId) {
+        // Remove outgoing edges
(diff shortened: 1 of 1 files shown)


orottier:
Hmm it does fix the starve_cpu example but now I see a `Panic occurred in Audio Processor: 'Index 12 for dropped Node in NodeCollection'. Removing node from graph.` every now and then

b-ma:
All good to me! 

(...and nice catch for [7b048c2](https://github.com/orottier/web-audio-api-rs/pull/400/commits/7b048c2f60a750543c8d8ecdef7910ce6686aee0))

PR #401, opened 2023-11-24
Run diagnostics on live AudioContext

orottier:
Usage: `context.run_diagnostics(|s| println!("Diagnostics:\n{}", s));`
It will generate a report on the next render quantum, in a somewhat realtime safe manner.
This will hopefully aid debugging of e.g. #396

### DIFF
--- a/src/context/online.rs
+++ b/src/context/online.rs
@@ -346,6 +346,19 @@ impl AudioContext {
 
     #[allow(clippy::missing_panics_doc)]
     pub fn run_diagnostics<F: Fn(String) + Send + 'static>(&self, callback: F) {
+        let mut buffer = Vec::with_capacity(32 * 1024);
+        {
+            let backend = self.backend_manager.lock().unwrap();
+            use std::io::Write;
+            writeln!(&mut buffer, "backend: {}", backend.name()).ok();
+            writeln!(&mut buffer, "sink id: {}", backend.sink_id()).ok();
+            writeln!(
+                &mut buffer,
+                "output latency: {:.6}",
+                backend.output_latency()
+            )
+            .ok();
+        }
         let callback = move |v| match v {
             EventPayload::Diagnostics(v) => {
                 let s = String::from_utf8(v).unwrap();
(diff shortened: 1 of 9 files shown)


src/context/online.rs
```diff
@@ -344,6 +344,26 @@ impl AudioContext {
         self.base().clear_event_handler(EventType::SinkChange);
     }
 
+    #[allow(clippy::missing_panics_doc)]
+    pub fn run_diagnostics<F: Fn(String) + Send + 'static>(&self, callback: F) {
```
orottier, line 348: Wondering if we should mark this method `#[doc(hidden)]` because I could imagine we are not keeping the API stable over different versions. Would be interesting to return more structured data in the future

src/context/online.rs
```diff
@@ -344,6 +344,26 @@ impl AudioContext {
...
+            }
+            _ => unreachable!(),
+        };
+
+        self.base().set_event_handler(
+            EventType::Diagnostics,
+            EventHandler::Once(Box::new(callback)),
+        );
+
+        let buffer = Vec::with_capacity(32 * 1024);
```
orottier, line 362: > 32kB ought to be enough for anybody.

-- Bill Gates

orottier:
/bench because we are messing with vtables here in tight loops

src/context/online.rs
b-ma: yup looks reasonable

src/render/processor.rs
```diff
@@ -116,6 +116,18 @@ pub trait AudioProcessor: Send {
     fn onmessage(&mut self, msg: &mut dyn Any) {
         log::warn!("Ignoring incoming message");
     }
+
+    /// Return the name of the actual AudioProcessor type
+    #[doc(hidden)] // not meant to be user facing
+    fn name(&self) -> &'static str {
```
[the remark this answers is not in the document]
b-ma: Nice

b-ma:
Looks all good to me!

PR #402, opened 2023-11-25
Fix panic in AudioBufferSource and Oscillator when supplying some options

orottier:
Pretty silly bug that has been present for quite some time. A panic occures when supplying a buffer in AudioBufferSourceOptions or a periodic wave in OscillatorOptions
The processor needs to be registered into the audio graph before we send it messages. We already do this correctly for all other nodes. Add regression tests.


b-ma:
Nice catch!

PR #403, opened 2023-11-30
Prevent the wpt-runner from crashing by avoiding some panics

orottier:
These are needed to make the wpt runner in the node bindings project work


b-ma:
Looks noce to me! The strategy for up/down mix is sensible indeed

Issue #404, opened 2023-12-02
MediaRecorder not working properly with MediaStream from get_user_media_sync


CallMeMSL:
Hi there,
thank you for your work on this great library!

I want to use this library for a Uni Project where I do live transcription and analyze spoken sentences with an LLM.
Somehow recording the mic doesn't seem to work.
This code produces an 11 Minute .wav file that blasts your ears away with garbage:

```
fn main() {
    let mic = media_devices::get_user_media_sync(MediaStreamConstraints::Audio);
    let recorder = MediaRecorder::new(&mic);
    let recording: Arc<Mutex<Vec<u8>>> = Default::default();
    let arc = Arc::clone(&recording);
    recorder.set_ondataavailable(move |mut event| {
        let mut r = arc.lock().unwrap();
        r.append(&mut event.blob);
    });
    recorder.start();
    std::thread::sleep(std::time::Duration::from_secs(4));
    recorder.stop();
    std::thread::sleep(std::time::Duration::from_secs(1));
    let data = recording.lock().unwrap();
    let mut file = fs::OpenOptions::new()
        .create(true)
        .write(true)
        .open("unprocessed.wav").unwrap();
    file.write_all(&data).unwrap();
}
```
Is there something obvious that I am missing? The example for `get_user_media_sync` for the [docs](https://docs.rs/web-audio-api/latest/web_audio_api/media_devices/fn.get_user_media_sync.html) works without any problems and I can hear myself clearly and without any latency.

I am using Linux with PipeWire.

Kind Regards,
David

orottier:
Thanks for the report. This is very interesting since you are combining the MediaDevices API and the MediaRecorder API, without an AudioContext in between. This is of course fine on the web, but not something we have really accounted for in our implementation - which main focus is the Web Audio API.

Tacking an AudioContext in between makes it work:

```
use std::sync::{Arc, Mutex};
use web_audio_api::context::{AudioContext, AudioContextLatencyCategory, AudioContextOptions};
use web_audio_api::media_devices;
use web_audio_api::media_devices::MediaStreamConstraints;
use web_audio_api::media_recorder::MediaRecorder;
use web_audio_api::node::AudioNode;

// If you are on Linux and use ALSA as audio backend backend, you might want to run
// the example with the `WEB_AUDIO_LATENCY=playback ` env variable which will
// increase the buffer size to 1024
//
// `WEB_AUDIO_LATENCY=playback cargo run --release --example recorder
fn main() {
[...]

PR #406, opened 2023-12-03
Avoid HRTF panic when constructing an AudioContext with low sample rate

orottier:
regarding https://github.com/orottier/web-audio-api-rs/issues/404#issuecomment-1837264275


b-ma:
Yup, better solution than panicking I think (and hopefully it should be temporary), I guess these low sample rate are more common than what we imagine, e.g. low-fi / retro gaming, etc.

PR #407, opened 2023-12-04
Trying out designs for OfflineAudioContext::suspend

orottier:
Some ideas for #100 . It does not adhere to the spec but it was very simple to implement. Probably has performance impact To be continued.


orottier:
To be continued in #412

PR #412, opened 2023-12-18
Add async OfflineAudioContext startRendering, suspend and resume

orottier:
Todo:

- [x] Decide if we like the `sync` interface this way
- [x] Check if it solves our node-binding issues
- [x] Address performance regression (caused by checking the need to suspend every quantum)

### DIFF
--- a/src/context/offline.rs
+++ b/src/context/offline.rs
@@ -12,6 +12,9 @@ use crate::{assert_valid_sample_rate, RENDER_QUANTUM_SIZE};
 use futures::channel::oneshot;
 use futures::sink::SinkExt;
 
+pub(crate) type OfflineAudioContextCallback =
+    dyn FnOnce(&mut OfflineAudioContext) + Send + Sync + 'static;
+
 /// The `OfflineAudioContext` doesn't render the audio to the device hardware; instead, it generates
 /// it, as fast as it can, and outputs the result to an `AudioBuffer`.
 // the naming comes from the web audio specification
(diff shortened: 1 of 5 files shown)


orottier:
I'm ready for a first round of reviews!

b-ma:
Ok cool! I will try to have a closer look this week, and (hopefully) try to see where this gets us with the bindings

src/context/offline.rs
```diff
@@ -101,13 +171,147 @@ impl OfflineAudioContext {
...
+    ///     context_clone.resume().await;
+    /// });
+    ///
+    /// let render_promise = context.start_rendering();
+    ///
+    /// let buffer = executor::block_on(async move { join!(suspend_promise, render_promise).1 });
+    /// assert_eq!(buffer.number_of_channels(), 1);
+    /// assert_eq!(buffer.length(), 512);
+    /// ```
+    pub async fn suspend_at(&self, suspend_time: f64) {
```
b-ma, line 216: Maybe we could just call it `suspend` as there is no alternative signature

src/context/offline.rs
```diff
@@ -101,13 +171,147 @@ impl OfflineAudioContext {
...
+        let quantum = (suspend_time * self.base.sample_rate() as f64 / RENDER_QUANTUM_SIZE as f64)
+            .ceil() as usize;
```
b-ma, line 218: should check the other conditions as well, cf. https://webaudio.github.io/web-audio-api/#dom-offlineaudiocontext-suspend, but maybe for another PR (this one may be a bit tricky "is less than or equal to the current time"

src/context/offline.rs
```diff
@@ -101,13 +171,147 @@ impl OfflineAudioContext {
...
+    ///     let mut src = context.create_constant_source();
+    ///     src.connect(&context.destination());
+    ///     src.start();
+    /// });
+    ///
+    /// let buffer = context.start_rendering_sync();
+    /// assert_eq!(buffer.number_of_channels(), 1);
+    /// assert_eq!(buffer.length(), 512);
+    /// ```
+    pub fn suspend_at_sync<F: FnOnce(&mut Self) + Send + Sync + 'static>(
```
b-ma, line 277: Same, maybe `suspend_sync` is clear enough?

b-ma:
Hey, except my small comments, look all good to me! I will try to implement the bindings, I let you know

orottier:
Thanks for the comments. I will address those, plus the leftover performance regression.

orottier:
Cool, the baseline performance regression is gone and even exceeds performance of the previous release. I guess due to ab48aff

PR #415, opened 2023-12-29
fix: implement clear_onstatechange and fix some event types

b-ma:
Some small fixes on issues I encountered while trying to review the EventTarget implementation


b-ma:
(looks like there is some kind of new clippy stuff...)

PR #417, opened 2024-01-01
Assert `AudioScheduledSourceNode` node start and stop arguments are valid

b-ma:
Ensure values given to start and stop are valid according to spec and wpt

cf. 
- https://webaudio.github.io/web-audio-api/#AudioScheduledSourceNode-methods
- https://webaudio.github.io/web-audio-api/#AudioBufferSourceNode-methods


orottier:
Thanks!

Issue #416, opened 2023-12-31
`OfflineAudioContext::state` is not updated when calling `supend` and `resume`


b-ma:

orottier:
Thanks, I will address this together with #410

Issue #411, opened 2023-12-11
Emit oncomplete event for OfflineAudioContext


orottier:
https://www.w3.org/TR/webaudio/#dom-offlineaudiocontext-oncomplete

PR #419, opened 2024-01-02
Set BaseAudioContext::state by render thread to signal device readiness

### LINKED ISSUE #410: Improve BaseAudioContext onChangeState to signal device readiness
See related discussion in #219 and #408

orottier:
- A context is instantiated in Suspended state
    - Online context will change to Running at the first render quantum
    - Online context will change to Closed when the render thread is dropped after a `close` call
    - Online context suspend/resume state is still managed by the control thread - TODO
    - Offline context will change to Running after startRendering
    - Offline context will set state to suspended when suspending, running when resumed
    - Offline context will set change to closed when rendering has finished

Fixes #410  #416  #101

It probably also fixes https://github.com/ircam-ismm/node-web-audio-api/pull/61#issuecomment-1871977771
> One thing that I found that is more a concern on the rust side, is that the first "resume" statechange event is sometimes triggered, sometimes not. I think this some kind of race condition but this is a bit weird from a user perspective.

Todo

- [x] find a solution for online AudioContext resume/suspend to signal device 'readiness'. Possible solution: first pause audio rendering by sending a ctrl msg to suspend the audio graph. When that message is handled, then emit the state_change event and update the state and suspend the cpal stream. Similar setup for resuming.
- [x] close_sync on online AudioContext is not truly sync. We don't guarantee the operation has finished before the function returns. Solution: use a channel to communicate if we are done
- [ ] implement oncomplete for offline context - tricky! to be addressed in #411 
- [x] add async methods for suspend/resume/close on online AudioContext

### DIFF
--- a/src/context/concrete_base.rs
+++ b/src/context/concrete_base.rs
@@ -147,6 +147,7 @@ impl BaseAudioContext for ConcreteBaseAudioContext {
 
 impl ConcreteBaseAudioContext {
     /// Creates a `BaseAudioContext` instance
+    #[allow(clippy::too_many_arguments)] // TODO refactor with builder pattern
     pub(super) fn new(
         sample_rate: f32,
         max_channel_count: usize,
--- a/src/context/concrete_base.rs
+++ b/src/context/concrete_base.rs
@@ -91,8 +91,8 @@ struct ConcreteBaseAudioContextInner {
     listener_params: Option<AudioListenerParams>,
     /// Denotes if this AudioContext is offline or not
     offline: bool,
-    /// Describes the current state of the `ConcreteBaseAudioContext`
-    state: AtomicU8,
+    /// Current state of the `ConcreteBaseAudioContext`, shared with the RenderThread
+    state: Arc<AtomicU8>,
     /// Stores the event handlers
     event_loop: EventLoop,
     /// Sender for events that will be handled by the EventLoop
@@ -150,6 +150,7 @@ impl ConcreteBaseAudioContext {
     pub(super) fn new(
         sample_rate: f32,
         max_channel_count: usize,
+        state: Arc<AtomicU8>,
         frames_played: Arc<AtomicU64>,
         render_channel: Sender<ControlMessage>,
         event_channel: Option<(Sender<EventDispatch>, Receiver<EventDispatch>)>,
@@ -175,7 +176,7 @@ impl ConcreteBaseAudioContext {
             queued_audio_listener_msgs: Mutex::new(Vec::new()),
             listener_params: None,
             offline,
-            state: AtomicU8::new(AudioContextState::Suspended as u8),
+            state,
             event_loop: event_loop.clone(),
             event_send,
         };
@@ -263,7 +264,6 @@ impl ConcreteBaseAudioContext {
         if self.state() != AudioContextState::Closed {
             let result = self.inner.render_channel.read().unwrap().send(msg);
             if result.is_err() {
-                self.set_state(AudioContextState::Closed);
                 log::warn!("Discarding control message - render thread is closed");
             }
         }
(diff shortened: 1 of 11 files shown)


src/context/concrete_base.rs
```diff
@@ -333,14 +334,14 @@ impl ConcreteBaseAudioContext {
...
     #[must_use]
     pub(super) fn state(&self) -> AudioContextState {
-        self.inner.state.load(Ordering::SeqCst).into()
+        self.inner.state.load(Ordering::Acquire).into()
     }
 
     /// Updates state of current context
     pub(super) fn set_state(&self, state: AudioContextState) {
         let current_state = self.state();
         if current_state != state {
```
orottier, line 343: A race condition may occur in concurrent calls to this method (e.g. making a massive amount of suspend calls) causing the statechange event to be sent too many times. This could be fixed with https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU64.html#method.compare_exchange but this method will probably be refactored away once we solve how to make suspend_sync better

b-ma:
Cool! Looks all good to me

This will be very simple to implement #411 from this point, I was unsure of the behaviour of await

orottier:
> This will be very simple to implement #411 from this point

Ahah, that is what you think (and I thought), but this is especially tricky because we don't spawn an event loop in the case of the OfflineAudioContext. Something to reconsider maybe. In any case we can't use the existing methods to spawn the complete event. I tried but it does not emit the event properly.

b-ma:
Ah damn I didn't realized that... 

> Something to reconsider maybe.

Indeed, and their will probably be another issue after that I just discovered yesterday... :) As we send the ended event for `AudioScheduledSourceNode` in the next render block, it might never be triggered in offline contexts if the rendering finishes on this block (there is a wpt test failling and I was thinking that was the actual issue)

orottier:
Indeed that will also be an issue. But to be fair, these events simply don't make sense on an OfflineAudioContext because there is not actionable thing you can do with them.
What information does it give you that a AudioBufferSourceNode has ended when the offline context is rendering? You don't have time to modify the audio graph at that point anyway because the rendering will continue while you handle the event...
I think the only sensible (but deprecated) event for OfflineAudioContext is oncomplete. All the others (statechange, ended, processorerror) could be informational but not actionable..

b-ma:
Hehe yup you are right, I guess the most important use case is actually testing and, eventually, some kind of monitoring (I do my best here :)...

orottier:
Ready for merging

src/context/online.rs
```diff
@@ -378,8 +380,86 @@ impl AudioContext {
...
+    /// * The audio device is not available
+    /// * For a `BackendSpecificError`
+    pub async fn resume(&self) {
+        // First ask the audio host to resume the stream
+        self.backend_manager.lock().unwrap().resume();
+
+        // Then, ask to resume rendering via a control message
+        let (sender, receiver) = oneshot::channel();
+        let notify = OneshotNotify::Async(sender);
+        let suspend_msg = ControlMessage::Resume { notify };
```
b-ma, line 420: `suspend_msg` is a bit misleading, should be `resume_msg`?

src/context/online.rs
```diff
@@ -378,8 +380,86 @@ impl AudioContext {
...
+    /// the progression of the currentTime, and stop processing audio data.
+    ///
+    /// # Panics
+    ///
+    /// Will panic when this function is called multiple times
+    pub async fn close(&self) {
+        // First, stop rendering via a control message
+        let (sender, receiver) = oneshot::channel();
+        let notify = OneshotNotify::Async(sender);
+        let suspend_msg = ControlMessage::Close { notify };
```
b-ma, line 440: same `close_msg`?

b-ma:
Nice!

orottier:
Released in v0.41

PR #422, opened 2024-01-03
fix: unwanted panic at Analiser initialization

b-ma:
There was an issue when initializing the Analyser node, e.g.:

```js
new AnalyserNode(c, {"minDecibels":-10,"maxDecibels":20})
```

Maybe not the cleanest way to go, I wonder if the default values should live in the `node/analyser.rs` and configure the inner analyzer from outside. What do you think?

### DIFF
--- a/src/analysis.rs
+++ b/src/analysis.rs
@@ -188,8 +188,8 @@ impl Analyser {
             ring_buffer,
             fft_size: DEFAULT_FFT_SIZE,
             smoothing_time_constant: DEFAULT_SMOOTHING_TIME_CONSTANT,
-            min_decibels: DEFAULT_MIN_DECIBELS,
-            max_decibels: DEFAULT_MAX_DECIBELS,
+            min_decibels: f64::NEG_INFINITY,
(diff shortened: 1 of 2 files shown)


src/analysis.rs
```diff
@@ -636,16 +636,25 @@ mod tests {
...
         let mut analyser = Analyser::new();
+        analyser.set_max_decibels(DEFAULT_MAX_DECIBELS); // init value
         analyser.set_min_decibels(DEFAULT_MAX_DECIBELS);
     }
 
     #[test]
     #[should_panic]
     fn test_max_decibels_constraints_lt_min_decibels() {
         let mut analyser = Analyser::new();
+        analyser.set_min_decibels(DEFAULT_MIN_DECIBELS); // init value
```
b-ma, line 647: these two test looks a bit silly with this change too

orottier:
Thanks, I understand the issue now. I think I can make it a bit nicer with a combined method `set_min_max_decibels(&mut self, min, max)` on the `Analyser`

PR #424, opened 2024-01-04
Fix AudioParam behaviour

b-ma:
Hey, some new stuff regarding AudioParam

- Fix `AudioParam` argument parsing
- Fix AutomationRate set behaviour on control thread
- ~Fix return value for raw setters (i.e. `set_value` and `set_automation_rate`)~~
- Fix return value for raw `set_value` setter
- Some refactoring

### DIFF
--- a/src/param.rs
+++ b/src/param.rs
@@ -1,7 +1,7 @@
 //! AudioParam interface
 use std::any::Any;
 use std::slice::{Iter, IterMut};
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{Ordering};
 use std::sync::{Arc, OnceLock};
 
 use arrayvec::ArrayVec;
@@ -276,8 +276,9 @@ pub(crate) struct AudioParamRaw {
     max_value: f32,     // immutable
     automation_rate: AutomationRate,
     automation_rate_constrained: bool,
+    current_value: Arc<AtomicF32>,
     // TODO Use `Weak` instead of `Arc`. The `AudioParamProcessor` is the owner.
-    shared_parts: Arc<AudioParamShared>,
+    // shared_parts: Arc<AudioParamShared>,
 }
 
 impl AudioNode for AudioParam {
@@ -366,7 +367,7 @@ impl AudioParam {
     //      test_exponential_ramp_a_rate_multiple_blocks
     //      test_exponential_ramp_k_rate_multiple_blocks
     pub fn value(&self) -> f32 {
-        self.raw_parts.shared_parts.load_current_value()
+        self.raw_parts.current_value.load(Ordering::Acquire)
     }
 
     /// Set the value of the `AudioParam`.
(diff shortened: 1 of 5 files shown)


src/param.rs
```diff
@@ -316,20 +320,23 @@ impl AudioNode for AudioParam {
...
+        self.raw_parts.automation_rate
     }
 
     /// Update the current value of the automation rate of the AudioParam
     ///
     /// # Panics
     ///
     /// Some nodes have automation rate constraints and may panic when updating the value.
-    pub fn set_automation_rate(&self, value: AutomationRate) {
+    pub fn set_automation_rate(&mut self, value: AutomationRate) -> AutomationRate {
```
b-ma, line 331: Actually this is the behaviour in JS, but I'm not sure this is what we want here.

I guess, either we generalize this behavior to all "raw" setters, either we decide they don't return a value?

(same for `set_value` actually)

src/param.rs
```diff
@@ -316,19 +320,20 @@ impl AudioNode for AudioParam {
...
+        self.raw_parts.automation_rate
     }
 
     /// Update the current value of the automation rate of the AudioParam
     ///
     /// # Panics
     ///
     /// Some nodes have automation rate constraints and may panic when updating the value.
-    pub fn set_automation_rate(&self, value: AutomationRate) {
+    pub fn set_automation_rate(&mut self, value: AutomationRate) {
```
orottier, line 331: I don't think marking these methods as `&mut self` will work in practise, because they are return as immutable refs from the AudioNodes:
`pub fn gain(&self) -> &AudioParam`

This could of course be changed to `pub fn gain(&mut self) -> &mut AudioParam` but that is a rather big change where we should assess pros and cons.

src/param.rs
```diff
@@ -272,9 +274,11 @@ pub(crate) struct AudioParamRaw {
     default_value: f32, // immutable
     min_value: f32,     // immutable
     max_value: f32,     // immutable
+    automation_rate: AutomationRate,
```
orottier, line 277: Including the non-Arc field here breaks the `Clone` implementation (which we need for the node bindings).
Calling set_automation_rate on a cloned AudioParam will not update the automation rate of the original AudioParam

orottier (commented):
Thanks for the changes.
I will probably pick this PR apart in the separate issues
- analyser panics, to be continued in #422 
- validation of audio param numerical values
- the issue of automation_rate_synchronicity (needs work)

b-ma:
Hum right, I didn't think about these points indeed (and sorry I indeed forked from the previous PR branch rather than from main)

For the automation rate problem, maybe the more straightforward solution is then just to update it through `AudioParamShared` when calling the method on control side rather than when the message is handled on render side, as in any case the renderer only uses it own `automation_rate` copy?

I let you cherry pick the changes, feel free to close the PR when needed

orottier:
Continued in #425 #422 and new issue https://github.com/orottier/web-audio-api-rs/issues/426

PR #427, opened 2024-01-04
test: fix #426

### LINKED ISSUE #426: AudioParam automationRate does not instantly update on the control thread
The following test fails:

    #[test]
    fn test_automation_rate_synchronicity() {
        let context = OfflineAudioContext::new(1, 0, 48000.);

        let opts = AudioParamDescriptor {
            name: String::new(),
            automation_rate: AutomationRate::A, // start with A
            default_value: 0.,
            min_value: 0.,
            max_value: 1.,
        };
        let (param, _render) = audio_param_pair(opts, context.mock_registration());

        // set to K
        param.set_automation_rate(AutomationRate::K);
        assert_eq!(param.automation_rate(), AutomationRate::K);
    }

This is quite tricky to fix. We could keep a local copy of the automation rate on the control thread, but because these methods all take `&self` it needs to be wrapped in an atomic and is susceptible to data races on concurrent calls. Best would be to require `&mut` on the methods but it is incompatible with how we return AudioParams from the AudioNodes (and the Clone impl)

b-ma:
Just to understand regarding #426, what's wrong with this solution?

### DIFF
--- a/src/param.rs
+++ b/src/param.rs
@@ -331,6 +331,7 @@ impl AudioParam {
             panic!("InvalidStateError: automation rate cannot be changed for this param");
         }
 
+        self.raw_parts.shared_parts.store_automation_rate(value);
         self.registration().post_message(value);
     }
 
@@ -719,7 +720,6 @@ impl AudioProcessor for AudioParamProcessor {
     fn onmessage(&mut self, msg: &mut dyn Any) {
         if let Some(automation_rate) = msg.downcast_ref::<AutomationRate>() {
             self.automation_rate = *automation_rate;
-            self.shared_parts.store_automation_rate(*automation_rate);
             return;
         }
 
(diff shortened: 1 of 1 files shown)


b-ma:
The failing check is not related to my change! :) cf. `tests/online.rs:99`

Just to mention, it's true that then our control and render nodes are slightly desynchronized, but we have already used this strategy, e.g. in `AudioBufferSourceNode`:

```rs
    pub fn set_loop(&mut self, value: bool) {
        self.loop_state.is_looping = value;
        self.registration.post_message(ControlMessage::Loop(value));
    }
```

orottier:
The problem is that concurrent calls to `set_automation_rate` with different values may result in an inconsistent state. Where the control thread will report that e.g. the automation rate = A while the render thread uses K.

The function signature `pub fn set_loop(&mut self, value: bool) {` uses `&mut` which prevents concurrent calls so there is no risk of inconsistencies.

The issue arises when
- thread 1 calls `set_automation_rate(A)`
- thread 2 concurrently calls `set_automation_rate(K)`
- thread 1 is first, so the `Arc<AudioParamShared>` is set to A-rate, however, at that point thread 1 could be pre-empted
- thread 2 will set `Arc<AudioParamShared>` to K-rate and posts a ctrl msg to the render thread to use K-rate processing 
- thread 1 is scheduled again, posts a ctrl msg to use A-rate processing
- result: control thread reporst K-rate, render thread uses A-rate

This issue was discussed in the past with uklotzde and that is why we made many methods use &mut to prevent this

b-ma:
Hum ok I understand now, thanks for the explanation! 

Seems to me the `Mutex` would be the right way to go then. As it is not used anyway by the render thread, it would not hurt too much (but maybe a bit ugly I understand :)

orottier:
Alright! This made me realize I need to improve https://github.com/orottier/web-audio-api-rs/pull/365 as well.
That PR fixes out of order arrival of channel config changes, but is in the current state susceptible to data races too

PR #428, opened 2024-01-04
Fix AudioParam automationRate does not instantly update on the control thread

### LINKED ISSUE #426: AudioParam automationRate does not instantly update on the control thread

orottier:
fixes #426

### DIFF
--- a/src/param.rs
+++ b/src/param.rs
@@ -327,12 +327,12 @@ impl AudioParam {
     /// # Panics
     ///
     /// Some nodes have automation rate constraints and may panic when updating the value.
-    pub fn set_automation_rate(&self, value: AutomationRate) {
+    pub fn set_automation_rate(&mut self, value: AutomationRate) {
         if self.raw_parts.automation_rate_constrained && value != self.automation_rate() {
             panic!("InvalidStateError: automation rate cannot be changed for this param");
         }
 
-        self.automation_rate = value;
+        self.raw_parts.automation_rate = value;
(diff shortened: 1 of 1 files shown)


src/param.rs
```diff
@@ -271,12 +271,12 @@ pub struct AudioParam {
 // helper struct to attach / detach to context (for borrow reasons)
 #[derive(Clone)]
 pub(crate) struct AudioParamRaw {
```
b-ma, line 273: Rename to `AudioParamInner`? Would be more consistent with the reset of the codebase

b-ma:
+1

PR #429, opened 2024-01-06
Fix: minimize floating point errors in AudioBufferSourceNode current_time calculation

b-ma:
Fixes the crashing the `the-audio-api/the-audiobuffersourcenode-interface/sample-accurate-scheduling.html` wpt test

### DIFF
--- a/src/node/audio_buffer_source.rs
+++ b/src/node/audio_buffer_source.rs
@@ -47,7 +47,7 @@ impl Default for AudioBufferSourceOptions {
     }
 }
 
-#[derive(Copy, Clone)]
+#[derive(Debug, Copy, Clone)]
 struct PlaybackInfo {
     prev_frame_index: usize,
     k: f32,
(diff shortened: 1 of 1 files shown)


src/node/audio_buffer_source.rs
```diff
@@ -649,19 +649,19 @@ impl AudioProcessor for AudioBufferSourceRenderer {
         let mut playback_infos = [None; RENDER_QUANTUM_SIZE];
 
         // compute position for each sample and store into `self.positions`
-        for playback_info in playback_infos.iter_mut() {
+        for (i, playback_info) in playback_infos.iter_mut().enumerate() {
+            let current_time = block_time + i as f64 * dt;
```
b-ma, line 653: This is actually the main change

src/node/audio_buffer_source.rs
```diff
@@ -649,19 +649,19 @@ impl AudioProcessor for AudioBufferSourceRenderer {
...
+
             if current_time < self.start_time
                 || current_time >= self.stop_time
                 || self.render_state.buffer_time_elapsed >= self.duration
             {
-                *playback_info = None;
-                current_time += dt;
```
b-ma, line 658: Rather than that

orottier (approved):
Thanks nice fix - could you have at my nitpick?

b-ma:
Done!

Issue #432, opened 2024-01-07
Unstable tests in tests/online.rs


orottier:
The event loop has unpredictable timings. Also the sleeps commands to measure progress of current time sometimes do not suffice

PR #433, opened 2024-01-13
Update pre-commit hooks

uklotzde:


orottier:
Thanks! Build fails due to #432 so I'm merging anyway

PR #435, opened 2024-01-13
Break up futures dependencies

uklotzde:


orottier:
Yup that's nicer. Thanks Uwe

PR #436, opened 2024-01-13
GitHub CI: Verify packaging

uklotzde:


orottier:
Thanks again. Did not know about this command, good to see it will extract the crate again and check if it builds

PR #440, opened 2024-01-30
Fix in ChannelMergerNode channel config & run assertions in constructor of nodes

orottier:


orottier:
```
  RESULTS:
  - # pass: 3939 (+83)
  - # fail: 1148 (-78)
  - # type error issues: 58 (-5)
 ```

b-ma:
Nice catches! All good to me

PR #441, opened 2024-01-30
Convolver: assert sample rate and number of channels on buffer

orottier:
```
  RESULTS:
  - # pass: 3970 (+31)
  - # fail: 1116 (-32)
  - # type error issues: 59 (+1)
 ```


b-ma:
+1

PR #442, opened 2024-01-31
Fix: AudioParam initial values should be visible immediately

orottier:
So use set_value instead of set_value_at_time in the node constructors

```
  RESULTS:
  - # pass: 3979 (+9)
  - # fail: 1107 (-9)
  - # type error issues: 59
```


b-ma:
+1

PR #443, opened 2024-01-31
Fix: do not upmix inputs to the AudioNode channelCount beforehand

orottier:
The upmixing will happen when the inputs are mixed together, and only when needed. This prevent unnecessary upmixes to stereo for many nodes with channelCountMode Max and ClampedMax.

This is pretty wild, it improves the wpt results with big margin. Mostly due to the PannerNode now picking mono-to-stereo by default - which fixes many assertions in the suite

```
  RESULTS:
  - # pass: 4306 (+327)
  - # fail: 780 (-327)
  - # type error issues: 59
```

### DIFF
--- a/src/render/graph.rs
+++ b/src/render/graph.rs
@@ -417,14 +417,6 @@ impl Graph {
             // acquire a mutable borrow of the current processing node
             let mut node = nodes[*index].borrow_mut();
 
-            // make sure all input buffers have the correct number of channels, this might not be
-            // the case if the node has no inputs connected or the channel count has just changed
-            let interpretation = node.channel_config.interpretation();
-            let count = node.channel_config.count();
-            node.inputs
-                .iter_mut()
-                .for_each(|i| i.mix(count, interpretation));
-
             // let the current node process (catch any panics that may occur)
             let params = AudioParamValues::from(nodes);
             scope.node_id.set(*index);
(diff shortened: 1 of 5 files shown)


src/render/graph.rs
```diff
@@ -417,14 +417,6 @@ impl Graph {
             // acquire a mutable borrow of the current processing node
             let mut node = nodes[*index].borrow_mut();
 
-            // make sure all input buffers have the correct number of channels, this might not be
```
orottier, line 420: This is the main change. I think I just misunderstood the spec 'channelCount' behaviour

tests/mixing.rs
```diff
@@ -93,7 +93,7 @@ fn test_stereo_to_discrete_stereo() {
 
     assert_eq!(output.number_of_channels(), 2);
     assert_float_eq!(output.get_channel_data(0), ONES, abs_all <= 0.);
-    assert_float_eq!(output.get_channel_data(1), ONES, abs_all <= 0.);
+    assert_float_eq!(output.get_channel_data(1), ZEROES, abs_all <= 0.);
```
orottier, line 96: This should indeed be ZEROES.
The intermediate nodes leaves the mono signal in place, the destination then upmixes to the Explicit 2 value but with Discrete interpretation. Meaning the missing 2nd channel will be filled with silence.

orottier:
Okay, the benchmark results are equally wild

orottier:
Okay the regression is gone but I don't understand why

orottier:
Okay, I reworked the iir implementation a bit to improve cache locality which improves performance and perhaps makes the bench more stable

orottier:
Final result

```
  RESULTS:
  - # pass: 4307 (+328)
  - # fail: 779 (-328)
  - # type error issues: 59
```

PR #444, opened 2024-02-01
Benchmarks: ensure src is looped for fair measurements

orottier:
src duration is 2 seconds but most benchmark duration are 10s


orottier:
Okay, benchmark job duration does not seem to increase 👍

b-ma:
Cool!

PR #447, opened 2024-02-05
MediaDevices: add support for MediaTrackConstraints.channelCount

### LINKED ISSUE #446: channel_count constraint of get_user_media_sync is not supported
I would like to suggest adding support for more than 2 channels for media streams. The WebAudio spec does include this in the form of specifying channelCount in the constraints.

At first I [noted](https://github.com/ircam-ismm/node-web-audio-api/issues/74) that this JavaScript wrapper for web-audio-api-rs didn't support it. But then I realized this is a feature request for the underlying rust implementation.

Here's a basic test case...

```Rust
use web_audio_api::context::AudioContext;
use web_audio_api::media_devices::{self, MediaTrackConstraints};
use web_audio_api::media_devices::MediaStreamConstraints;

fn main() {
	let context = AudioContext::default();

	let devices = media_devices::enumerate_devices_sync();
	for device in devices.iter() {
		println!( "{}", device.label() );

		let mut media_track_constraints = MediaTrackConstraints::default();
		media_track_constraints.device_id = Some( device.device_id().to_string() );

		// Ideally, this would be supported
		// media_track_constraints.channel_count = Some( 18 );

		let input = media_devices::get_user_media_sync(
			MediaStreamConstraints::AudioWithConstraints( media_track_constraints )
		);
		let stream = context.create_media_stream_source(&input);

		println!( "> {}", stream.channel_count() );
	}
}

```
Hopefully this could be added in the near future.

orottier:
Fixes #446

TODO: channelCount should be modeled as a ConstrainULong, not just u32.
TODO: cubeb support

### DIFF
--- a/src/io/mod.rs
+++ b/src/io/mod.rs
@@ -112,7 +112,10 @@ pub(crate) fn build_output(
 }
 
 /// Set up an input stream (microphone) bases on the selected features (cubeb/cpal/none)
-pub(crate) fn build_input(options: AudioContextOptions) -> MediaStream {
+pub(crate) fn build_input(
+    options: AudioContextOptions,
+    number_of_channels: Option<u32>,
+) -> MediaStream {
     #[cfg(all(not(feature = "cubeb"), not(feature = "cpal")))]
     {
         panic!("No audio backend available, enable the 'cpal' or 'cubeb' feature")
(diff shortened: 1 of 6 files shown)


src/io/mod.rs
```diff
@@ -112,7 +112,10 @@ pub(crate) fn build_output(
 }
 
 /// Set up an input stream (microphone) bases on the selected features (cubeb/cpal/none)
-pub(crate) fn build_input(options: AudioContextOptions) -> MediaStream {
+pub(crate) fn build_input(
+    options: AudioContextOptions,
```
b-ma, line 116: Not for this PR, but it is a bit weird to have this `AudioContextOptions` struct here

b-ma:
Cool, really nice you could make it so fast

src/io/mod.rs
orottier: I know, but it was just too convenient because we needed all the same fields to construct the input stream. But now we need additional fields so it makes less sense indeed

PR #449, opened 2024-02-11
Implement `Debug` for all public facing types

### LINKED ISSUE #418: Many public facing types do not implement Debug
Types such as `context::online::AudioContext`, `GainNode`, and likely others do not implement `std::fmt::Debug`. This makes implementing `Debug` for wrapping types cumbersome, and prevents the use of `expect` (among other things) involving these types.

I'm not sure if there are any major trade offs in terms of compile time or other metrics, but I'd personally prefer to have `Debug` regardless.

orottier:
Fixes #418 

Size of the compiled lib is now 5.1MB, was 5.0MB on main

cc @CorvusPrudens

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -56,10 +56,15 @@ pub use media_element::MediaElement;
 mod resampling;
 pub mod worklet;
 
-#[derive(Debug)]
 #[repr(transparent)]
 pub(crate) struct AtomicF32 {
     bits: AtomicU32,
+}
+
+impl std::fmt::Debug for AtomicF32 {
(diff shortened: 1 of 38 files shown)


src/lib.rs
```diff
@@ -4,6 +4,7 @@
 #![warn(clippy::missing_panics_doc)]
 #![warn(clippy::clone_on_ref_ptr)]
 #![deny(trivial_numeric_casts)]
+#![deny(missing_debug_implementations)]
```
orottier, line 7: We now enforce it at the crate level

PR #453, opened 2024-02-18
Hide AudioProcessor trait from public API and force users to use worklet

orottier:
Hey, this is meant as a discussion starter. I think our AudioWorkletNode implementation is mature enough to serve all custom processing needs for end users. Hiding our non-spec AudioProcessor keeps the API surface neat and tidy, and allows for further iteration without breaking changes.

As you can see from the latency_tester example, using the AudioWorkletProcessor reduces boilerplate without sacrifying features.

The only thing missing then is the MessagePort functionality. I have some work prepared how we could fill this in, but it does become a bit awkward due to lack of inheritance. Since rust has message passing natively built in, we could also opt to leave this functionality out (and provide useful examples to fill in the gap).


orottier:
The MessagePort implementation could look like https://github.com/orottier/web-audio-api-rs/compare/feature/hide-audioprocessor-trait...feature/message_port?expand=1
However I think the added complexity is not worth it (and there are still some issues to fix in that implementation)

b-ma:
Hey, looks like a pretty good idea!

Would be interesting to port other example, for example https://googlechromelabs.github.io/web-audio-samples/audio-worklet/ to see if everything looks good. I could try to have a shot on that when I find some time

For message port, I think this is quite interesting to have it as it would guarantee the order of the messages, but we could take some time to iterate of course

src/node/convolver.rs
```diff
@@ -155,7 +155,7 @@ impl ConvolverNode {
             channel_config,
         } = options;
 
-        let mut node = context.base().register(move |registration| {
+        let mut node = context.base().base().register(move |registration| {
```
b-ma, line 158: Do we need this `.base().base()`, that looks a bit weird

orottier: That was a runaway vim macro, will fix

orottier:
Thanks for the feedback. I will add an example with manual message passing via Rust primitives. Then I will probably merge and see if I can iterate on a better design of the message port. The `Box<dyn Any>` to allow for any message object type is very cumbersome

orottier:
Hey it took quite a bit of mental gymnastics but I'm quite happy with the result.
The AudioWorkletNode now has a spec compliant `port() -> MessagePort` method.
On the render thread we diverge a bit, and have a separate `AudioWorkletProcessor::onmessage` method a `RenderScope.post_message`. But still I think that is as clean and performant as it could be.

orottier:
The diff is quite large so I recommend you to look at the `examples/worklet_message_port.rs` example

src/render/processor.rs
```diff
@@ -38,6 +38,18 @@ impl std::fmt::Debug for RenderScope {
 }
 
 impl RenderScope {
+    /// Send a message to the corresponding AudioWorkletNode of this processor
+    ///
+    /// This method is just a shim of the full
+    /// [`MessagePort`](https://webaudio.github.io/web-audio-api/#dom-audioworkletprocessor-port)
+    /// `postMessage` functionality of the AudioWorkletProcessor.
+    pub fn post_message(&self, msg: Box<dyn Any + Send + 'static>) {
```
orottier, line 46: I have made the Box explicit here, so the user is forced to think about the allocation that occurs

orottier:
If you agree with all this @b-ma I will release a version 1.0 release candidate :)

b-ma:
Hey, I think it looks all good to me

> The AudioWorkletNode now has a spec compliant `port() -> MessagePort` method.
> On the render thread we diverge a bit, and have a separate `AudioWorkletProcessor::onmessage` method a `RenderScope.post_message`. But still I think that is as clean and performant as it could be.

Yeah a bit disappointing but I agree the API looks rather clean

> If you agree with all this @b-ma I will release a version 1.0 release candidate :)

It's up to you :) (...but I have to confess I start to have some hard time thinking of possible issues on that point)

PR #457, opened 2024-02-23
remove unneeded `&mut self`

b-ma:
Let me know if I miss something here


orottier:
> Let me know if I miss something here

nothing :)

PR #459, opened 2024-02-24
Ensure k-rate with connected inputs deliver a constant value per quantum

orottier:
Results of the-audioparam-interface

```
  RESULTS:
  - # pass: 1413 (+66)
  - # fail: 510 (-66)
  - # type error issues: 9
```


b-ma (approved):

b-ma:
Nice catch!

PR #460, opened 2024-02-24
AudioParam end time calculation, use round instead of ceil

orottier:
```
  RESULTS:
  - # pass: 1587 (+174)
  - # fail: 180
  - # type error issues: 9
```

### DIFF
--- a/src/param.rs
+++ b/src/param.rs
@@ -1060,7 +1060,7 @@ impl AudioParamProcessor {
 
         // fill buffer with current intrinsic value until `event.time`
         if infos.is_a_rate {
-            let end_index = ((time - infos.block_time).max(0.) / infos.dt) as usize;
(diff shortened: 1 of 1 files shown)


src/param.rs
```diff
@@ -1110,9 +1110,9 @@ impl AudioParamProcessor {
 
         if infos.is_a_rate {
             let start_index = self.buffer.len();
-            // we need to `ceil()` because if `end_time` is between two samples
+            // we need to `round()` because if `end_time` is between two samples
```
b-ma, line 1113: no sure these comments still hold then... which I consider is a bit strange actually

b-ma (approved):

b-ma:
Given the wpt results and according to the "Monkey sees wpt improvement, monkey approves" principle :), it's in any case good to merge, but I must say I'm a bit confused with that

orottier:
I have added a TODO with link to this PR to possibly look into it later

PR #463, opened 2024-02-26
Fix: check length > 0 in constructor

b-ma:


b-ma:
Was unsure how to handle that failling test ffd1691, feel free to change that if you have another idea on how to handle that

PR #465, opened 2024-02-26
AudioScheduledSourceNode can return tail_time false when not scheduled

### LINKED ISSUE #462: Free audio processor resources when the node will never start processing
For `AudioScheduledSourceNode`, renderer should be dropped when control side has been dropped and the source has not been started (it cannot be started anyway)

orottier:
The render thread will not garbage collect them as long as the control
handle is kept. Once the control handle is dropped, and the node is
still not scheduled to start (start_at != f64::MAX) we can safely drop
the node from the audio graph.

Fixes #462

Apologies I mixed in some housekeeping

### DIFF
--- a/src/node/audio_buffer_source.rs
+++ b/src/node/audio_buffer_source.rs
@@ -411,7 +411,7 @@ impl AudioProcessor for AudioBufferSourceRenderer {
         params: AudioParamValues<'_>,
         scope: &AudioWorkletGlobalScope,
     ) -> bool {
-        // single output node
+        // Single output node
         let output = &mut outputs[0];
 
         let sample_rate = scope.sample_rate as f64;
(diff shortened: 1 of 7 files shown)


src/node/audio_buffer_source.rs
```diff
@@ -55,7 +55,7 @@ struct PlaybackInfo {
     k: f32,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy)]
 struct LoopState {
```
b-ma, line 59: To continue housekeeping, I think we could just remove this struct altogether now that all updates are made through message passing and that it is not used anymore to save some `Arc`

b-ma (approved):

b-ma:
> Apologies I mixed in some housekeeping

No problem, always nice to clean the code

src/node/audio_buffer_source.rs
orottier: Ah, that's where it is coming from. But I do think it improves the readability slightly (makes the Renderer struct less daunting)

Issue #421, opened 2024-01-03
EventLoop never terminates


orottier:
A new EventLoop is spawned upon creation of a ConcreteBaseAudioContext, but it never terminates.
I should terminate when both the RenderThread and ConcreteBaseAudioContext is dropped (?)

PR #466, opened 2024-02-27
Online AudioContext improvements (continue playing after drop, terminate event loop)

### LINKED ISSUE #464: Should the AudioContext continue playing when it is dropped?
Currently we stop the audio stream when the context goes out of scope. This does not correspond to the JS behaviour where the context only halts playing when `close()` is called.

We should probably make a decision before releasing the v1 version

orottier:
Fixes #464 #421

### DIFF
--- a/examples/roundtrip_latency_test.rs
+++ b/examples/roundtrip_latency_test.rs
@@ -138,7 +138,7 @@ fn main() {
 
     let estimated_latency = Arc::new(AtomicF64::new(0.));
 
-    let _context = if test {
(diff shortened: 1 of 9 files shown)


b-ma:
Looks good to me! 

Just one question, what is the use case you envision for #464 ?

b-ma:
> Or do you think it is better to keep the previous behaviour?

I dont know actually, no strong opinion... I don't really understand why someone would do that but I don't think this hurts neither :)

Issue #468, opened 2024-02-28
Issue with NodeCollection


b-ma:
Just had this issue

```
thread '<unnamed>' panicked at /web-audio-api-rs/src/render/node_collection.rs:96:32:
Index 45 for dropped Node in NodeCollection
```

I really don't know where it comes from, the app is rather complex and dynamic. I will try to investigate next week and see if I can come with a minimal failing test bed

orottier:
Thanks, that would be very helpful.
Does your graph involve nodes connecting to audio params? E.g. constant src -> playback rate -> buffer src? I just realized this may trigger an issue with dynamic lifetimes

In any case I should improve the error message in these rare cases because the current message is not helpful. (Track caller does not appear to work on trait methods)

orottier:
Marking this one solved. For now..

b-ma:
> For now..

Huhu :)

PR #469, opened 2024-02-29
Dynamic lifetime issues and the node collection

orottier:
Apologies for the complex PR. I will let it simmer for a while and put some more thought in it, but wanted to share already.

Introduce AudioProcess::has_side_effects

A processor with no side effect will be dropped when the control handle
is gone and there are not output ports connected.

This makes the cleanup of AudioParams easier and more robust.

Relates to https://github.com/orottier/web-audio-api-rs/issues/397 and https://github.com/orottier/web-audio-api-rs/issues/468

### DIFF
--- a/src/node/delay.rs
+++ b/src/node/delay.rs
@@ -374,6 +374,10 @@ impl AudioProcessor for DelayWriter {
         // let the node be decommisioned if it has no input left
         false
     }
+
+    fn has_side_effects(&self) -> bool {
+        true // message passing
+    }
 }
 
 impl DelayWriter {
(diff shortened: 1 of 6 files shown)


orottier:
Right, I think we should probably merge this as it solves a panic in the audio param dynamic lifetimes.
@b-ma did you manage to reproduce the panic you experienced in your complex graph? I hope this PR would solve it.

b-ma:
No I didn't have time to make a simpler test bed, but I can test my app against this branch right now, the panic was rather easy to get, I let you now in the afternoon

b-ma:
Just tested and I didn't manager to make it crash!

b-ma (approved):

b-ma:
Did understood all the details, but I will check that when needed and it looks all good to me

orottier:
Cool, good to hear!

PR #470, opened 2024-03-06
AudioParam validations, compound limits, optimizations for single-valued params

orottier:


orottier:
No changes in WPT score

src/node/constant_source.rs
```diff
@@ -201,23 +201,33 @@ impl AudioProcessor for ConstantSourceRenderer {
...
-                    *o = value;
-                }
-
-                current_time += dt;
-            });
+
+        // fast path
+        if offset.len() == 1
+            && self.start_time <= scope.current_time
+            && self.stop_time > next_block_time
```
b-ma, line 208: I guess it could be `self.stop_time >= next_block_time`, if stop time == next block time we can still fill the current buffer with constant value, no?

b-ma (approved):

b-ma:
Really nice!

src/node/constant_source.rs
orottier: Hmm, I think it would then not correspond to the check below:
```
if current_time < self.start_time || current_time >= self.stop_time {
    *o = 0.;
}
```


The spec says:
An [AudioScheduledSourceNode](https://webaudio.github.io/web-audio-api/#AudioScheduledSourceNode) is said to be playing when its associated [BaseAudioContext](https://webaudio.github.io/web-audio-api/#BaseAudioContext)'s [currentTime](https://webaudio.github.io/web-audio-api/#BaseAudioContext) is greater or equal to the time the [AudioScheduledSourceNode](https://webaudio.github.io/web-audio-api/#AudioScheduledSourceNode) is set to start, and less than the time it’s set to stop.

So only non-zero output for `start <= time < stop`?

b-ma: > So only non-zero output for start <= time < stop?

Yup agreed, but not sure this is contradictory actually

if `self.stop_time == next_block_time`, last sample time for the current block is still below `self.stop_time`
then in first sample of next block `current_time == self.stop_time` so it passes the check you mention

Or am I missing something ?

PR #473, opened 2024-03-11
WPT fixes: channel splitter/merger error msg, offline ctx channel count mode

orottier:
Big win with the AudioNode tests because of the fixed panic in the destination node

```
  RESULTS:
  - # pass: 5199 (+190)
  - # fail: 640 (-19)
  - # type error issues: 62 (+8)
```

### DIFF
--- a/src/node/channel_merger.rs
+++ b/src/node/channel_merger.rs
@@ -4,10 +4,30 @@ use crate::context::{AudioContextRegistration, BaseAudioContext};
 use crate::render::{
     AudioParamValues, AudioProcessor, AudioRenderQuantum, AudioWorkletGlobalScope,
 };
+use crate::MAX_CHANNELS;
 
 use super::{
     AudioNode, ChannelConfig, ChannelConfigOptions, ChannelCountMode, ChannelInterpretation,
 };
+
+/// Assert that the given number of channels is valid for a ChannelMergerNode
+///
+/// # Panics
+///
+/// This function will panic if:
+/// - the given number of channels is outside the [1, 32] range,
+/// 32 being defined by the MAX_CHANNELS constant.
+///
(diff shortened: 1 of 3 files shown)


src/node/channel_merger.rs
```diff
@@ -4,11 +4,31 @@ use crate::context::{AudioContextRegistration, BaseAudioContext};
...
+///
+/// # Panics
+///
+/// This function will panic if:
+/// - the given number of channels is outside the [1, 32] range,
+/// 32 being defined by the MAX_CHANNELS constant.
+///
+#[track_caller]
+#[inline(always)]
+pub(crate) fn assert_valid_number_of_channels(number_of_channels: usize) {
```
b-ma, line 23: This one is already defined in `lib.rs` but with `NotSupportedError` instead of `IndexSizeError`, weird

src/node/destination.rs
```diff
@@ -74,7 +74,7 @@ impl AudioNode for AudioDestinationNode {
         // OfflineAudioContext, then the channel count mode cannot be changed.
         // An InvalidStateError exception MUST be thrown for any attempt to change the value.
         assert!(
-            !self.registration.context().offline(),
+            !self.registration.context().offline() || v == ChannelCountMode::Explicit,
```
b-ma, line 77: Cannot find where it is explained in the spec, maybe add something as comment ?

b-ma:
Apart my small comment, looks all good to me

src/node/destination.rs
orottier: Good point, added. I also found a spec issue related to the channelCount. Will double check and raise an issue at the spec repo

b-ma (approved):

PR #476, opened 2024-03-30
Rename ChannelConfigOptions -> AudioNodeOptions

### LINKED ISSUE #474: Rename ChannelConfigOptions to AudioNodeOptions

orottier:
Fixes #474

### DIFF
--- a/examples/worklet_message_port.rs
+++ b/examples/worklet_message_port.rs
@@ -42,7 +42,7 @@ impl WhiteNoiseNode {
             output_channel_count: vec![1],
             parameter_data: HashMap::new(),
             processor_options: (),
-            channel_config: AudioNodeOptions::default(),
+            audio_node_options: AudioNodeOptions::default(),
         };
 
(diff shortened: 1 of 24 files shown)


examples/worklet_message_port.rs
```diff
@@ -42,7 +42,7 @@ impl WhiteNoiseNode {
             output_channel_count: vec![1],
             parameter_data: HashMap::new(),
             processor_options: (),
-            channel_config: ChannelConfigOptions::default(),
+            channel_config: AudioNodeOptions::default(),
```
b-ma, line 45: Just a bit confused with the field name now, don't have a better idea though calling it `audio_node_options` would look a bit silly maybe...

b-ma:
Except my comment, looks good to me (I'm surprised we didn't catch this one earlier)

examples/worklet_message_port.rs
orottier: Very good point, since we use composition instead of inheritance all public `channel_config` keys should be renamed to `audio_context_options`

PR #478, opened 2024-03-31
Fix: do not assume curve is centered around zero

b-ma:
This fixes the `the-waveshapernode-interface/silent-inputs.html` wpt test

### DIFF
--- a/src/node/waveshaper.rs
+++ b/src/node/waveshaper.rs
@@ -393,11 +393,6 @@ impl AudioProcessor for WaveShaperRenderer {
         let input = &inputs[0];
         let output = &mut outputs[0];
 
-        if input.is_silent() {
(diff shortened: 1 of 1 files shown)


b-ma:
Actually I just realize that we could be a bit smarter and check if the curve value for 0 signal is zero, then we could could still propagate silent input, what do you think?

orottier (approved):
Thanks.

I realize we have an omission in our implementation regarding https://webaudio.github.io/web-audio-api/#AudioNode-actively-processing

Currently, the graph `[WaveShaper] -> [Destination]` with an non-centric curve will emit output in our lib after merging this. But since the WaveShaper is not fed with a source node, the spec requires it to be silent. Not sure if this is important enough to write a ticket though.

PR #480, opened 2024-04-01
Also run the EventLoop for OfflineAudioContext

orottier:
Iteration 1: just enable the event loop

### DIFF
--- a/src/context/concrete_base.rs
+++ b/src/context/concrete_base.rs
@@ -108,7 +108,7 @@ struct ConcreteBaseAudioContextInner {
     /// Stores the event handlers
     event_loop: EventLoop,
     /// Sender for events that will be handled by the EventLoop
-    event_send: Option<Sender<EventDispatch>>,
+    event_send: Sender<EventDispatch>,
 }
 
 impl BaseAudioContext for ConcreteBaseAudioContext {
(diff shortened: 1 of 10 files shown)


orottier:
So spawning a thread per OfflineAudioContext does not really contribute to the number of instructions/cycles.
However with `cargo bench` we see an interesting measurement:

```
bench_audio_buffer_decode
                        time:   [15.106 µs 15.118 µs 15.130 µs]
                        change: [+83.155% +83.804% +84.604%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 12 outliers among 100 measurements (12.00%)
```

And a less severe one:

```
bench_analyser_node     time:   [2.4068 ms 2.4148 ms 2.4242 ms]
                        change: [+10.787% +11.168% +11.606%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 10 outliers among 100 measurements (10.00%)
  6 (6.00%) high mild
  4 (4.00%) high severe
```

`cargo run --release --example benchmarks` does not show any relevant changes though so I'm a bit confused..

Next step: experiment with a global event loop (instead of per context)

b-ma:
So far this looks like a simplification, quite nice IMO :)

orottier:
Yeah yeah, I agree.

I have a new idea to prevent a thread per OfflineAudioContext: after each render quantum run the event loop for all pending events. This both prevents the performance impact of an extra thread, plus it solves the issue that the ScriptProcessorNode faces when rendering too fast. As you discovered, Chrome blocks the rendering to await the output_buffer. This reflects in the wpt results: https://wpt.fyi/results/webaudio/the-audio-api/the-scriptprocessornode-interface/simple-input-output.html?label=experimental&label=master&aligned&q=webaudio
We currently face the same issue as Firefox with the current draft PR.

Running the event loop after each render quantum also corresponds to the spec's `Perform a microtask checkpoint.` mentioned in https://webaudio.github.io/web-audio-api/#rendering-loop

orottier:
Iteration 2: run the event loop in the same thread as the offline audiocontext, handling events after each render quantum

orottier:
This iteration fixes the large performance regression in  cargo bench `bench_audio_buffer_decode` and `bench_analyser_node` but has a footprint on all other benches (because we call the event loop on each quantum)

orottier:
The fast path trick mitigates most of the performance issues but I'm still not really happy
- it's a bit of a dirty trick
- fast path is not really fair because all events are only flushed after rendering is done
- for graphs like granular synthesis many events will be buffered and take memory

b-ma:
Actually, I think the numbers without the fast path trick were not that bad, also considering that:
1. It's a kind of structural impact due to new feature
2. It only affect offline processing were performance is a bit less important
3. We know this does not impact the actual audio processing, which is the big deal

So maybe we could just accept the small regression?

orottier:
You are probably right. But I couldn't leave it so I just wrote an unrelated performance optimization to ease my mind.

/bench

PR #482, opened 2024-04-05
Fix: more compliant option checks for ChannelSplitterNode

b-ma:
Fixes small wpt errors in wpt/webaudio/the-audio-api/the-channelsplitternode-interface/ctor-channelsplitter.html


orottier:
I think I already touched these files quite a few times, but happy to see another improvement!

PR #485, opened 2024-04-07
Fix - Add upper bound for valid sample rates

b-ma:
cf. wpt/webaudio/the-audio-api/the-audiocontext-interface/audiocontextoptions.html

Also modified a bit the lower bound so that we have:

```
min_sample_rate = min_required_in_spec / 2
max_sample_rate = max_required_in_spec * 2
```

This is a bit arbitrary, let me know if it's ok for you

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -137,10 +137,12 @@ impl AtomicF64 {
 #[track_caller]
 #[inline(always)]
 pub(crate) fn assert_valid_sample_rate(sample_rate: f32) {
+    // Arbitrary cutoffs defined as:
+    // min_sample_rate = min_required_in_spec / 2
+    // max_sample_rate = max_required_in_spec * 2
     let min_sample_rate = 4_000.;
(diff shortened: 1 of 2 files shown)


orottier:
Hey thanks. I agree some limits would be good to prevent running with a typo'd sample rate (of which I am apparently guilty).

However it would be sad to prevent some use cases without good reason - perhaps you just want to analyse electromagnetic coil whine instead of voice/music.

Chrome says: The hardware sample rate provided (1000) is outside the range [3000, 768000].
Firefox says: AudioContext constructor: Sample rate 1000 is not in the range [8000, 192000]
Safari supports 3000 to about 380000

So, eh, how about 3000 to 384000?

b-ma:
Just put

```
// min_sample_rate = min_required_in_spec / 4
// max_sample_rate = max_required_in_spec * 4
let min_sample_rate = 2_000.;
let max_sample_rate = 384_000.;
```

so that values do not look that arbitrary :)

PR #486, opened 2024-04-08
Fix - DelayNode ring buffer size

b-ma:
There was an issue with how the ring buffer size was calculated, this fixes all failing wpt tests for the delay node


orottier:
Great, thanks

PR #487, opened 2024-04-08
Fix - BiquadFilterNode gain max value

b-ma:
very small one, cf. https://webaudio.github.io/web-audio-api/#dom-biquadfilternode-gain


orottier:
👍

Issue #489, opened 2024-04-09
Implement onended events for more cases


orottier:
_Originally posted by @b-ma in https://github.com/ircam-ismm/node-web-audio-api/issues/95#issuecomment-2044245304_

But there are still some issues. In short we have two cases we don't handle correctly (just take an `OscillatorNode` for the example, but this is the same behaviour for all `AudioScheduledSourceNode`)

1. Explicit stop before the end of rendering **works**

```js
const context = new OfflineAudioContext(1, 48000, 48000);
const src = new ConstantSourceNode(context);
src.connect(context.destination);
src.onended = () => console.log('onended fired 1');

src.start();
src.stop(0.9);

context.startRendering();
```

2. Explicit stop at end of the rendering **fails**

```js
const context = new OfflineAudioContext(1, 48000, 48000);
const src = new ConstantSourceNode(context);
src.connect(context.destination);
src.onended = () => console.log('onended fired 1');

src.start();
src.stop(1.);

context.startRendering();
```

This one is rather logical, as we fire the ended event on next render quantum, which in this case is never computed. We could fix it either by a) reviewing when the `ended` events are sent, or b) by doing a last dry run at the end of the offline context (maybe a bit dirty I agree :)

3. Implicit stop **fails**

```js
const context = new OfflineAudioContext(1, 48000, 48000);
const src = new ConstantSourceNode(context);
src.connect(context.destination);
src.onended = () => console.log('onended fired 1');

src.start();
// no explicit stop

context.startRendering();
```

Not sure how to handle that case (is present in wpt, cf. the-audiobuffersourcenode-interface/audiobuffersource-ended.html), maybe have the offline duration in render scope and use it to clamp stop time?

PR #490, opened 2024-04-10
Introduce AudioProcessor::before_drop pre-destructor hook

orottier:
This allows the processor to send some final events, or move some items to the garbage collector thread. This will help with https://github.com/orottier/web-audio-api-rs/pull/368 

In this commit we use the before_drop hook to ensure that the ended event is sent for all AudioScheduledSourceNodes when the stop time is implicit. This fixes https://github.com/orottier/web-audio-api-rs/issues/489


b-ma (approved):

b-ma:
Really nice!

PR #491, opened 2024-04-11
Code reorganization of src/node/mod.rs

orottier:
No changes to the public or private API


b-ma:
Cool

Would be nice to do the same for `AudioWorkletGlobalScope`, I always spend some time to find it

orottier:
> Would be nice to do the same for `AudioWorkletGlobalScope`, I always spend some time to find it

I just don't know where to put it, it belongs both in src/render and with worklet.rs.. I could move it to the root?

b-ma:
> > Would be nice to do the same for `AudioWorkletGlobalScope`, I always spend some time to find it
> 
> I just don't know where to put it, it belongs both in src/render and with worklet.rs.. I could move it to the root?

Ehm true... don't really know neither... maybe having a proper `worklet` mod would make sens as it is where it is publicly exposed?

PR #493, opened 2024-04-14
AudioNode disconnect methods

### LINKED ISSUE #471: Review AudioNode connect/disconnect methods
We currently have
- `connect(node)`
- `connect_at(node, output, input)`
- `disconnect()`
- `disconnect_from(node)`

I think we at least also need:
- `disconnect_at(output)`
- `disconnect_from_at(node, output, input)`

For reference, the wasm bindings project model the methods like https://docs.rs/web-sys/latest/web_sys/struct.AudioNode.html

Then for maximum fun, I just read that the `disconnect` method should throw exceptions if they are trying to remove non-existing edges. https://webaudio.github.io/web-audio-api/#dom-audionode-disconnect-destinationnode
This means the control side needs to have graph knowledge as well..

orottier:
Fixes #471 

As you can see we get quite a few methods extra, this is conform https://webaudio.github.io/web-audio-api/#AudioNode . Unlike connect, disconnect does not have optional arguments (instead, we treat a missing value as wildcard when disconnecting). I'm up for debate for new names though, or possibly exposing the `Option` arguments...

We now bookkeep connection on the control side, and throw when one tries to disconnect a non-existent connection.

Bonus fix: disconnect without args would previously clear all connections of a node, but the spec mandates that only outgoing connections are cleared!

The node-lib needs some work after merging this to support stuff like `node.disconnect(number)`

### DIFF
--- a/src/context/concrete_base.rs
+++ b/src/context/concrete_base.rs
@@ -426,24 +426,29 @@ impl ConcreteBaseAudioContext {
         self.inner.queued_messages.lock().unwrap().push(message);
     }
 
-    /// Disconnects all outputs of the audio node that go to a specific destination node.
-    ///
-    /// # Panics
-    ///
-    /// Panics if this node was not connected to the target node
-    pub(crate) fn disconnect_from(&self, from: AudioNodeId, to: AudioNodeId) {
+    /// Disconnects outputs of the audio node, possibly filtered by output node, input, output.
+    pub(crate) fn disconnect(
+        &self,
+        from: AudioNodeId,
+        output: Option<usize>,
+        to: Option<AudioNodeId>,
+        input: Option<usize>,
+    ) {
         // check if the node was connected, otherwise panic
         let mut has_disconnected = false;
         let mut connections = self.inner.connections.lock().unwrap();
-        connections.retain(|&(c_from, output, c_to, input)| {
-            let retain = c_from != from || c_to != to;
+        connections.retain(|&(c_from, c_output, c_to, c_input)| {
+            let retain = c_from != from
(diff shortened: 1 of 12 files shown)


b-ma:
Cool, really nice!

> The node-lib needs some work after merging this to support stuff like node.disconnect(number)

Yup, but this should be rather straightforward (I'm starting to be quite fluent in parsing IDL now :)

For the naming what about?

```rs
node.connect(dest)
node.connect_dest_from_output_to_input(dest, output, input)
// or (...I think I prefer the first one for the symmetry with disconnect, 
// but it is more consistent according to `connect`)
node.connect_from_output_to_input(dest, output, input)

node.disconnect()
node.disconnect_output(output)
node.disconnect_dest_from_output(dest, output)
node.disconnect_dest_from_output_to_input(dest, output, input)
```

This is a bit more verbose, but I find naming like `disconnect_dest_output` a bit confusing as `output` seems to refer to `dest`

src/context/concrete_base.rs
```diff
@@ -283,17 +287,23 @@ impl ConcreteBaseAudioContext {
...
-        if !magic {
-            let message = ControlMessage::ControlHandleDropped { id };
-            self.send_control_msg(message);
+        // Ignore magic nodes
+        if id == DESTINATION_NODE_ID || id == LISTENER_NODE_ID || LISTENER_PARAM_IDS.contains(&id.0)
+        {
+            return;
         }
+
+        // Inform render thread that the control thread AudioNode no langer has any handles
```
b-ma, line 297: typo `longer` instead of `langer`

orottier:
Thanks for the suggestions, I applied them

PR #499, opened 2024-04-18
Fix hanging UI thread for certain AudioContext::suspend/stop actions

orottier:
There were quite a few issues with
- multiple subsequent suspend() calls
- multiple subsequent close() calls
- calling close() on a suspended context

There are regression tests for these cases now.

CC @ggadwa

### DIFF
--- a/src/context/concrete_base.rs
+++ b/src/context/concrete_base.rs
@@ -351,10 +351,8 @@ impl ConcreteBaseAudioContext {
 
     /// Updates state of current context
     pub(super) fn set_state(&self, state: AudioContextState) {
-        // Only to be used from OfflineAudioContext, the online one should emit the state changes
-        // from the render thread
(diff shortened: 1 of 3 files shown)


src/context/concrete_base.rs
```diff
@@ -351,10 +351,8 @@ impl ConcreteBaseAudioContext {
 
     /// Updates state of current context
     pub(super) fn set_state(&self, state: AudioContextState) {
-        // Only to be used from OfflineAudioContext, the online one should emit the state changes
-        // from the render thread
-        debug_assert!(self.offline());
-
+        // Only used from OfflineAudioContext or suspended AudioContext, otherwise the state
+        // changed ar spawned from the render thread
```
b-ma, line 355:
```suggestion
        // changed are spawned from the render thread
```

b-ma (approved):

Issue #423, opened 2024-01-03
Add async method for `BaseAudioContext::decodeAudioData`


orottier:
See #412 for some initial inspiration.
Async versions of suspend and resume were introduced in #419 and #412 
We don't need the async_trait crate anymore because of https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html

PR #500, opened 2024-04-20
Implement async method decode_audio_data (however still blocking IO)

orottier:
Fixes https://github.com/orottier/web-audio-api-rs/issues/423


orottier:
Build fails because we need to bump the MSRV to Rust 1.75 - https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html
1.75 was released end of december 2023
I don't think this PR is very important, so maybe let's wait till June so we can keep our MSRV to lag 6 months behind

PR #504, opened 2024-04-28
RenderCapacity - return owned value

b-ma:
Started to have a look to create the JS bindings for `RenderCapacity` and found this was returned as reference, which has been found always very complicated to wrap

Also look a bit more more consistent in terms of API according to listener and destination.

Let me know if you see any issue with that


orottier:
Hey I understand, and I think this api change is fine. I will update your PR a bit later and take the chance to add some tests too, but feel free to continue the work on the node bindings assuming the owned value

b-ma:
Ok nice!

PR #511, opened 2024-05-16
Experiment with thread local pool of audio render quanta

orottier:


orottier:
Interesting regression on the delay nodes between version one and two.. 🤔

orottier:
I can't reproduce the regression locally (not with cargo bench, and also the example/benchmarks.rs shows improvement for the delay test..)

Let's try again /bench

orottier:
Hey @b-ma , the idea for this change was growing for a long time on me. Happy to see some benchmark improvements again. Could you have a look? I'm certainly interested what `cargo bench` reports for you (run first on main, then on this branch) for the delay tests. The IAI numbers look bad but I don't understand why and `cargo bench` and the benchmarks example.rs show improvements also for delay tests..

b-ma:
Hey, I'm a bit out of networks this week-end, I'll have a look on Monday

b-ma:
Hey, so on my side:

1. With the `benchmarks` examples, all numbers seems to improve
2. With `cargo` bench, it seems I have some small regression but on benches that involve a `sine`, that's a bit weird...

```
bench_sine              time:   [2.8590 ms 2.8640 ms 2.8685 ms]
                        change: [+2.4613% +3.0625% +3.6813%] (p = 0.00 < 0.05)
                        Performance has regressed.

bench_sine_gain         time:   [3.0442 ms 3.0501 ms 3.0555 ms]
                        change: [+2.7440% +3.4710% +4.2233%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

do we agree that I have run `cargo bench` on main first and that the diff is made according to that? It seems logical but I'm not sure actually

src/render/quantum.rs
```diff
@@ -1,75 +1,46 @@
...
+    use std::cell::RefCell;
 
-    pub fn silence(&self) -> AudioRenderQuantumChannel {
-        AudioRenderQuantumChannel {
-            data: Rc::clone(&self.inner.zeroes),
-            alloc: Rc::clone(&self.inner),
-        }
+    thread_local! {
+        /// Thread local pool of pre-allocated sample slice that can be reused after being dropped
+        static POOL: RefCell<Vec<Rc<[f32; RENDER_QUANTUM_SIZE]>>> = RefCell::new(Vec::with_capacity(32));
```
b-ma, line 21: I wonder if we shouldn't have a larger initial pool? As a stereo delay of 1 second (default `max_delay_time` uses 375 * 2 `AudioRenderQuantumChannel`, having something like 1024 preallocated buffers would guarantees we don't allocate at least when creating one delay

orottier: Good point. We could definitely do it but we should realize this could happen on any thread and the memory is never reclaimed, so to be a bit conservative would not be bad.
I'm having a bit more thoughts about this setup though, when using async rendering the OfflineAudioContext can move to any thread of the executor and we may end up leaking a lot of memory.
I could specialize it to be of different size for offline and online rendering. Starting with a small pool for offline rendering is not bad because we can safely allocate more during rendering.

orottier:
> do we agree that I have run `cargo bench` on main first and that the diff is made according to that? It seems logical but I'm not sure actually

Yup that's how it works. Weird that your sines seem to be affected because none of this PR should matter there..

PR #512, opened 2024-05-30
Fix a panic in AudioWorklet rendering with zero outputs

orottier:


b-ma:
Hey, let's good to me!

(apart this weird test fail which I doesn't seem related...)

PR #514, opened 2024-06-10
Experiment: python bindings

orottier:
Don't take it too seriously, just wanted to put my mind to something else.

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -160,7 +160,7 @@ jobs:
   release:
     name: Release
     runs-on: ubuntu-latest
-    # if: "startsWith(github.ref, 'refs/tags/')"
+    if: "startsWith(github.ref, 'refs/tags/')"
     # needs: [linux, musllinux, windows, macos, sdist] # musl and linux ignored for now
     needs: [windows, macos, sdist]
     steps:


b-ma:
> Don't take it too seriously, just wanted to put my mind to something else.

Héhé nice! No risk I take it too seriously, I hate python :) but let me know if you want me to ask around for testers at some point (students seems to like python for some strange reason...)

orottier:
Thanks for the offer!
Let me fight with the workflows for a few more hours, and then add a few more nodes and we should be good to go :)

orottier:
Okay that was terrible but I managed to publish mac and windows binaries for https://pypi.org/project/web-audio-api/
I will have another look at linux later.

orottier:
Hey @lordmauve that would be great! I will polish and merge this later today. A lot of work is to be done on the actual binding and fixing the inheritance of classes, but perhaps together we can figure that out

orottier:
Okay something broke the builds since last year so I probably should not merge this as is. Something with the python version. I can look into it but it won't be today. @lordmauve feel free to build upen this work in a separate PR. I'm also not sure if python bindings should go into the main repo, or if a separate one makes more sense..

b-ma:
Hey, cool to see this going forward!

> not sure if python bindings should go into the main repo, or if a separate one makes more sense..

Just my two cents but I think another repo would be more legible

orottier:
Alright, closing in favour of https://github.com/orottier/py-web-audio-api

Issue #508, opened 2024-05-07
MediaRecorder: support more mimeTypes than audio/wav


orottier:
First #507 should be implemented.
Then we can switch out `hound` for the encoding and use `symphonia`

orottier:
Encoding is not provided by Symphonia so we either need to find a new general purpose encoder or use dedicated crates per type

PR #519, opened 2024-07-24
MediaRecorder spec compliance

### LINKED ISSUE #507: MediaRecorder spec compliance: add options and isTypeSupported
To future proof the v1 API we should change the constructor from

`pub fn new(stream: &MediaStream) -> MediaRecorder`

to

`pub fn new(stream: &MediaStream, options: MediaRecorderOptions) -> Result<MediaRecorder, DomException>`

with

```
#[non_exhaustive]
#[derive(Debug, Clone, Default)]
pub struct DomException {
    pub name: String,
    pub message: String,
}

#[non_exhaustive]
#[derive(Debug, Clone, Default)]
pub struct MediaRecorderOptions {
    pub mime_type: String, // use "" as default
}

fn is_type_supported(mime_type: String) -> bool;

orottier:
Hello, it has been a while

Even though we only support audio/wav I wanted to push this one out early because it is a breaking API change

Fixes #507 
To be continued with #508

### DIFF
--- a/examples/recorder.rs
+++ b/examples/recorder.rs
@@ -2,7 +2,7 @@ use std::io::Write;
 use web_audio_api::context::{
     AudioContext, AudioContextLatencyCategory, AudioContextOptions, BaseAudioContext,
 };
-use web_audio_api::media_recorder::MediaRecorder;
+use web_audio_api::media_recorder::{MediaRecorder, MediaRecorderOptions};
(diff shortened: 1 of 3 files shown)


b-ma:
Hey, yup indeed! (...many things to do before holidays on my side :)

Looks good to me

PR #523, opened 2024-09-05
Rethrow panics in nodes during user tests

### LINKED ISSUE #522: Possible issue with `catch_unwind` in renderer for unit tests
Hey,

I just realized today that the `catch_unwind` in unit test can probably lead to false positive tests

```rs
match panic::catch_unwind(catch_me) {
    Ok(tail_time) => (true, tail_time),
    Err(e) => {
        node.outgoing_edges.clear();
        scope.report_error(e);
        (false, false)
    }
}
```

For example, if we want an empty buffer as expected result, but the node just crashes (and is removed from graph), then the resulting buffer is empty so the test is passing... And we can only see the problem if we run the tests with the `--nocapture` flag.

I checked very very rapidly and maybe we already have such issues, e.g. `node::waveshaper::tests::test_user_defined_options` (TBC)

orottier:
Fixes #522

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -191,6 +191,13 @@ impl EventLoop {
             result = ControlFlow::Break(());
         }
 
+        // In unit tests, we rethrow panics to avoid missing critical node exceptions
+        // https://github.com/orottier/web-audio-api-rs/issues/522
+        #[cfg(test)]
+        if let EventPayload::ProcessorError(e) = event.payload {
+            panic!("Rethrowing exception during tests: {:?}", e);
+        }
+
         let mut event_handler_lock = self.event_handlers.lock().unwrap();
         let callback_option = event_handler_lock.remove(&event.type_);
         drop(event_handler_lock); // release Mutex while running callback


orottier:
@b-ma maybe you are more familiar with the inner workings of the WaveShaperNode?
The issue in the tests is

```
---- node::waveshaper::tests::test_user_defined_options stdout ----
thread 'node::waveshaper::tests::test_user_defined_options' panicked at src/node/waveshaper.rs:328:46:
assertion `left == right` failed
  left: 128 --->  the channel length
 right: 256 ---->  processor.input_frames_next()

in self.upsampler_x2.process(channels):

   6: web_audio_api::node::waveshaper::Resampler::process
             at ./src/node/waveshaper.rs:328:9
   7: <web_audio_api::node::waveshaper::WaveShaperRenderer as web_audio_api::render::processor::AudioProcessor>::process
             at ./src/node/waveshaper.rs:429:21
```

I don't see why the upsampler x2 should have 256 as input size...
```
impl ResamplerConfig {
    fn upsample_x2(channels: usize, sample_rate: usize) -> Self {
        let chunk_size_in = RENDER_QUANTUM_SIZE * 2;
```

Could you shed some light?

b-ma:
Hum I don't know actually, looks like either it has always been there and the underlying lib has been fixed somehow, or I introduced that by mistake in some refactoring... I can have a look

orottier:
Okay thanks. Let me know how it goes, if it takes more time I can release a fix for the buffer loop separately

b-ma:
cf #526

orottier:
Thanks. I will release tonight. V-1.0.0 probably, why not? :)

Issue #525, opened 2024-09-10
AudioBufferSourceNode: inconsistencies in fast and slow track


orottier:
When I force all processing in slow-track, with

```
+++ b/src/node/audio_buffer_source.rs
@@ -528,7 +528,7 @@ impl AudioProcessor for AudioBufferSourceRenderer {
             self.render_state.is_aligned = false;
         }

-        if self.render_state.is_aligned {
+        if false && self.render_state.is_aligned {
             // ---------------------------------------------------------------
             // Fast track
             // ---------------------------------------------------------------
```

I run into a few test failures:
```
failures:
    node::audio_buffer_source::tests::test_end_of_file_fast_track_2
    node::audio_buffer_source::tests::test_fast_track_loop_mono
    node::audio_buffer_source::tests::test_fast_track_loop_stereo
    node::delay::tests::test_max_delay
```

`test_max_delay` appears to be a slight numerical discrepancy, but the other three hint at possible bugs in one of the two tracks.

PR #527, opened 2024-09-13
AudioBufferSourceNode: inconsistencies in fast and slow track

b-ma:
Work in progress for #525

The failing tests are passing now but it introduced / revealed another issue, `test_loop_hangs` is now failing for another reason....

The semantics of loop start and end are not clear and they are thus not applied properly in edge cases I think, I need / will take some time to review all this properly

### DIFF
--- a/src/node/audio_buffer_source.rs
+++ b/src/node/audio_buffer_source.rs
@@ -411,6 +411,7 @@ impl AudioBufferSourceRenderer {
             if self.loop_state.start < 0. {
                 self.loop_state.start = 0.;
             }
+
             if self.loop_state.start > duration {
                 self.loop_state.start = duration;
             }
@@ -455,8 +456,8 @@ impl AudioProcessor for AudioBufferSourceRenderer {
         let buffer = match &self.buffer {
             None => {
                 output.make_silent();
-                // #462 like the above arm, we can safely return tail_time false if this node has
-                // no buffer set.
+                // #462 like the above arm, we can safely return tail_time false
+                // if this node has no buffer set.
                 return false;
             }
             Some(b) => b,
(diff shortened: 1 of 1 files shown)


orottier:
Right, I understand. Let me know if I can help somehow!

b-ma:
Ok, I think the failing test was faulty actually, I renamed it to `test_loop_out_of_bounds` and added some comment, could you confirm you read it the same as me?

I left a bunch of todos I will try to tackle in the next days

src/node/audio_buffer_source.rs
```diff
@@ -1461,14 +1590,25 @@ mod tests {
...
 
-        assert_float_eq!(channel[0], 1.0, abs_all <= 0.);
-        assert_float_eq!(channel[1..], [0.; 48_000 - 1][..], abs_all <= 0.);
+        // Both loop points will be clamped to buffer duration due to rules defined at
+        // https://webaudio.github.io/web-audio-api/#dom-audiobuffersourcenode-loopstart
+        // https://webaudio.github.io/web-audio-api/#dom-audiobuffersourcenode-loopend
+        // Thus it violates the rule defined in
+        // https://webaudio.github.io/web-audio-api/#playback-AudioBufferSourceNode
+        // `loopStart >= 0 && loopEnd > 0 && loopStart < loopEnd`
+        // Hence the whole buffer should be looped
```
orottier, line 1604: Yes, I agree. I did not put too much thought into the expected result earlier, just focused on fixing the indefinite loop. Thanks!

b-ma:
Ok I think it's all good to me for this one

orottier:
Benchmarks are currently failing #529 
I was interested in the performance impact for the interpolation moving to f64. It requires some casting back and forth and increases the size of a 'hot' object. I will run the locally instead. Anyway, since this PR improves correctness, so the results of benchmarks are not too important

orottier:
Okay maybe we don't have benchmarks for slow-track processing anyway.. oh well

PR #530, opened 2024-09-17
Fix absn sub sample scheduling

b-ma:
Some other AudioBufferSourceNode fixes, based on #527 

Mainly makes all remaining tests from the-audiobuffersourcenode-interface/sub-sample-scheduling.html test pass


orottier:
I'm releasing everything in v1.0.1 now. Thanks!

PR #531, opened 2024-09-19
Fix `AudioBufferSourceNode` stitching

b-ma:
This fixes some other wpt test (which were quite tricky)

Results for `the-audiobuffersourcenode-interface`:

#### Before
```
RESULTS:
  - # pass: 601
  - # fail: 31
  - # type error issues: 0
```

#### After (edit: better with the actual new results :)
```
RESULTS:
  - # pass: 617
  - # fail: 15
  - # type error issues: 0
```

I added a new dependency to test for floating point computation errors, I didn't feel like implementing this myself :) 

I think it is ok as this could help us in other part of the code where we test against absolute values which is not very clean, let me know what you think!

I tried to comment the code but let me know if something is not clear, or doesn't feel right to you

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ include = [
 rust-version = "1.71"
 
 [dependencies]
+almost = "0.2.0"
 arc-swap = "1.6"
 arrayvec = "0.7"
 cpal = { version = "0.15", optional = true }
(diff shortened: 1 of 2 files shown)


orottier:
Okay cool. Looks like good improvements. I'm a bit worried about performance regression due to extra loops, float accuracy etc but this seems all slow-track stuff anyway

PR #541, opened 2024-12-09
Fix ABSN backward playback

b-ma:
Hey,

This fix an issue that has been introduced somehow in earlier refactoring (probably in #530), breaking reverse playback rate. I also added a test to prevent future regression

No change in wpt results (meaning there is no test for that in wpt...?)

### DIFF
--- a/src/node/audio_buffer_source.rs
+++ b/src/node/audio_buffer_source.rs
@@ -1223,6 +1223,42 @@ mod tests {
     }
 
     #[test]
+    fn test_negative_playback_rate() {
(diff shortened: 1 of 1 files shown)


orottier:
Great, thanks! We could consider adding a test to wpt :)

b-ma:
> We could consider adding a test to wpt :)

Yup :), actually this seems to be not implemented neither in Chrome (<https://issues.chromium.org/issues/40719652>) nor Firefox (<https://bugzilla.mozilla.org/show_bug.cgi?id=1308438>)

PR #545, opened 2025-01-12
Fix: panics if `AudioContextOptions#sinkId` is invalid

b-ma:
cf. https://webaudio.github.io/web-audio-api/#validating-sink-identifier - step 4.

It's indeed a little change in the API but that is the spec and the behaviour that is actually documented. 
I guess it's ok as it doesn't break builds, but only raise a clippy warning eventually?


orottier:
Thanks! Let's consider this a bugfix/behavioural change, not an API change (as the function call remains the same). I will put in a next minor release.

Issue #542, opened 2024-12-12
Multichannel support for ConvolverNode


orottier:
The current implementation only supports mono processing. See #537 for some discussion

orottier:
Multi channel is now supported, however we need to add one bit still:

> There is still a todo regarding input number of channel changes (i.e. from stereo to mono), I guess we need to add a bit of bookkeeping to properly handle the tail of the channel that disappeared, but not completely sure how to do that yet.

PR #546, opened 2025-01-12
Feat: implement convolution channel configuration

b-ma:
This implements the multichannel stuff for the `ConvolverNode`
cf. https://webaudio.github.io/web-audio-api/#Convolution-channel-configurations

There is still a todo regarding input number of channel changes (i.e. from stereo to mono), I guess we need to add a bit of bookkeeping to properly handle the tail of the channel that disappeared, but not completely sure how to do that yet.

### todos

- [ ] input channel number change (delegate to another PR?)
- [x] wpt check

### DIFF
--- a/src/node/convolver.rs
+++ b/src/node/convolver.rs
@@ -223,6 +223,7 @@ impl ConvolverNode {
             let renderer = ConvolverRenderer {
                 convolvers: None,
                 impulse_length: 0,
+                impulse_number_of_channels: 0,
                 tail_count: 0,
             };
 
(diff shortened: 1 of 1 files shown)


b-ma:
WPT results:

#### before
```
- # pass: 260
- # fail: 68
- # type error issues: 0
```

#### after
```
- # pass: 322
- # fail: 6
- # type error issues: 0
```

orottier:
Alright, that looks nice.
I took me a while to understand the spec regarding multi channel, but as I see it know the number of input channels and the number of convolver channels determine the output channel count. And you create a second (identical) convolver in case the IR buffer has only one channel to handle stereo inputs.

orottier:
Relates to #542

PR #552, opened 2025-09-05
Echo cancellation

remorses:
Still not ready for review.

In preparation of https://github.com/ircam-ismm/node-web-audio-api/issues/155#issuecomment-3257327103

I used the aec crate to implement echo cancellation: https://github.com/thewh1teagle/aec

I also updated the microphone example to enable echo cancellation by default


orottier:
Hey just let us know when you want us to have a look. We really do not mind looking at work in progress and provide early feedback

PR #559, opened 2026-04-12
AudioContext::try_new, propagate cpal/cubeb errors into online audio context

orottier:
Only change to the public API is the addition of `AudioContext::try_new`
Other ones still panic (resume, suspend, latency), could make those public later.

Warning: AI assisted coding, but I reviewed thoroughly

### DIFF
--- a/.github/workflows/msrv.yaml
+++ b/.github/workflows/msrv.yaml
@@ -25,7 +25,7 @@ jobs:
 
       - name: Install Rust toolchain
         # Aligned with `rust-version` in `Cargo.toml`
-        uses: dtolnay/rust-toolchain@1.81
+        uses: dtolnay/rust-toolchain@1.85
 
       - name: Check out repository
         uses: actions/checkout@v3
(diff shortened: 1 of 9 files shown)


b-ma:
Cool, interesting ! 

I'm more and more convinced this "panic everywhere" paradigm we choose (...and I was advocating for...) is a bit problematic in the end

> Warning: AI assisted coding, but I reviewed thoroughly

(I'll have to pretend I didn't see that :)

orottier:
> I'm more and more convinced this "panic everywhere" paradigm we choose (...and I was advocating for...) is a bit problematic in the end

Yeah, I guess, but we can tackle these one by one. This is a nice start.

> (I'll have to pretend I didn't see that :)

I have tried to hold it off for as long as I could! But it is nice to have some assistance for tasks that were long overdue and you cannot get yourself to start on it

orottier:
I now remember the spec editors agree with you @b-ma , and that is why they introduced https://github.com/WebAudio/web-audio-api/issues/2567 
Let me check if we can incorporate this, and if we then still need the `try_new` constructor.

PR #561, opened 2026-04-13
Fix 558: cpal::Stream::suspend() stops the render callback causing control messages to stall

### LINKED ISSUE #558: cpal::Stream::suspend() stops the render callback causing control messages to stall
****Problem****

When suspending the AudioContext, the CPAL backend currently calls cpal::Stream::suspend(). However, when the stream is paused, the CPAL output callback stops being invoked. Since the RenderThread logic runs inside this callback, control messages (including resume) are no longer processed.

The AudioContextRegistration destructor sends a ControlMessage::ControlHandleDropped. Depending on timing, the following call:
```
       let result = self.inner.render_channel.read().unwrap().send(msg);
```
can get stuck if the channel is full, since the render thread isn’t consuming messages while paused.

Currently, _thread.rs_ already handles suspension in the callback:

```
        // if the thread is still booting, suspended, or shutting down, fill with silence
        if self.suspended || !self.graph.as_ref().is_some_and(Graph::is_active) {
            output_buffer.fill(S::from_sample_(0.));
            return;
        }
```
**Possible solution:**

Avoid pausing the CPAL stream when suspending; instead, continue invoking the output callback and fill with silence.

_cpal.rs_

Image

orottier:
This would be the most easy fix. I need to think about it a bit more. Probably refactor that giant ConcreteBaseAudioContextInner a bit..

Fixes #558

### DIFF
--- a/src/context/concrete_base.rs
+++ b/src/context/concrete_base.rs
@@ -289,7 +289,7 @@ impl ConcreteBaseAudioContext {
     }
 
     pub(crate) fn suspend_control_msgs(&self, msg: ControlMessage) {
-        let sender = self.inner.render_channel.write().unwrap();
+        let sender = self.inner.render_channel.read().unwrap();
         *self.inner.suspended_messages.lock().unwrap() = Some(Vec::new());
         if sender.send(msg).is_err() {
             log::warn!("Discarding control message - render thread is closed");
@@ -297,7 +297,7 @@ impl ConcreteBaseAudioContext {
     }
 
     pub(crate) fn resume_control_msgs(&self, msg: ControlMessage) {
-        let sender = self.inner.render_channel.write().unwrap();
+        let sender = self.inner.render_channel.read().unwrap();
         let messages = self
             .inner
             .suspended_messages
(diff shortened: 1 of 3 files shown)


src/context/concrete_base.rs
```diff
@@ -272,13 +275,49 @@ impl ConcreteBaseAudioContext {
     /// emitted.
     pub(crate) fn send_control_msg(&self, msg: ControlMessage) {
         if self.state() != AudioContextState::Closed {
-            let result = self.inner.render_channel.read().unwrap().send(msg);
+            let sender = self.inner.render_channel.read().unwrap();
+            if let Some(queued) = self.inner.suspended_messages.lock().unwrap().as_mut() {
```
orottier, line 280:
```suggestion
            // if the context is suspended, buffer the message and don't send it
            if let Some(queued) = self.inner.suspended_messages.lock().unwrap().as_mut() {
```

orottier:
I don't see a way to simplify ConcreteBaseAudioContextInner right now. So I will leave it like this

PR #565, opened 2026-04-15
Audio playback stats (alongside RenderCapacity)

### LINKED ISSUE #564: Add AudioPlayoutStats interface
Via https://github.com/WebAudio/web-audio-api/issues/2642 and https://github.com/WebAudio/web-audio-api/pull/2645

orottier:
Fixes #564 

Warning, some AI-assisted coding

### DIFF
--- a/src/context/online.rs
+++ b/src/context/online.rs
@@ -583,6 +583,9 @@ impl AudioContext {
             return;
         }
 
+        // Stop AudioRenderCapacity before closing so no capacity events are queued during shutdown.
+        self.render_capacity.stop();
+
         if self.state() == AudioContextState::Running {
             // First, stop rendering via a control message
             let (sender, receiver) = oneshot::channel();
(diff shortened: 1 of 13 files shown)


src/context/online.rs
```diff
@@ -257,7 +259,8 @@ impl AudioContext {
 
         // Setup AudioRenderCapacity for this context
         let base_clone = base.clone();
-        let render_capacity = AudioRenderCapacity::new(base_clone, load_value_recv);
+        let render_capacity = AudioRenderCapacity::new(base_clone, stats.clone());
```
b-ma: really cosmetics but having
```rs
let render_capacity = AudioRenderCapacity::new(base.clone(), stats.clone());
```
for consistency with next line would be nice

b-ma:
Cool!

Somehow related question: I think since a while that as the `AudioRenderCapacity` is not in the spec anymore we could maybe abuse it to expose more stats helpful for debugging (eventually behind a compilation flag)? 
For example, I once found really helpful the be able to track the size the graph in real time to debug JS issues with GC, but I had to hack directly in the source code to do that. What do you think, should I open a dedicated issue for that or do your think there is a better design / alternative?

orottier:
That could be nice. Maybe you can open an issue with the desired diagnostics? 
Instead of a compiler flag it could be an env var to opt in. This means you dont need to recompile

PR #567, opened 2026-04-15
Fix unstable tests for AudioContext

### LINKED ISSUE #432: Unstable tests in tests/online.rs

orottier:
Fixes #432

### DIFF
--- a/tests/online.rs
+++ b/tests/online.rs
@@ -12,7 +12,36 @@ use std::thread;
 use std::time::Duration;
 use web_audio_api::MAX_CHANNELS;
 
+const TEST_TIMEOUT: Duration = Duration::from_secs(1);
+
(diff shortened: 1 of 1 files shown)


tests/online.rs
```diff
@@ -8,13 +8,41 @@ use web_audio_api::context::{
 };
 use web_audio_api::node::AudioNode;
 
-use std::sync::atomic::{AtomicBool, AtomicU32, Ordering};
 use std::thread;
 use std::time::Duration;
 use web_audio_api::MAX_CHANNELS;
 
+const TEST_TIMEOUT: Duration = Duration::from_secs(1);
```
orottier:
```suggestion
const TEST_TIMEOUT: Duration = Duration::from_secs(2);
```

Funny we had another failure on 1 second, lets try double

PR #568, opened 2026-04-15
Fix and modernize the benchmark github action

orottier:


orottier:
Lets see: /bench

orottier:
Ugh, this needs to be on main to be tested.

PR #578, opened 2026-04-18
fix: accept buffer_size 0 in createScriptProcessorNode

### LINKED ISSUE #576: Accept createScriptProcessorNode(0)
From https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-createscriptprocessor

The [bufferSize](https://webaudio.github.io/web-audio-api/#dom-scriptprocessornode-buffersize) parameter determines the buffer size in units of sample-frames. If it’s not passed in, or if the value is 0, then the implementation will choose the best buffer size for the given environment, which will be constant power of 2 throughout the lifetime of the node.

--

Let's opt for 0 ==> 256 for now? Or based on the latency setting, use more

mvanhorn:
Per the [Web Audio spec](https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-createscriptprocessor), a `bufferSize` of 0 means the implementation should choose the best buffer size. Previously, passing 0 triggered the assertion because `0 / 256 = 0` and `0.is_power_of_two()` is false.

Now defaults to 256 when 0 is passed, before the validation check runs.

Closes #576

### DIFF
--- a/src/node/script_processor.rs
+++ b/src/node/script_processor.rs
@@ -72,7 +72,7 @@ impl ScriptProcessorNode {
     /// # Panics
     ///
     /// This function panics if:
-    /// - `buffer_size` is not 256, 512, 1024, 2048, 4096, 8192, or 16384
+    /// - `buffer_size` is not 0, 256, 512, 1024, 2048, 4096, 8192, or 16384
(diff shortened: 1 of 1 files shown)


orottier:
Thanks!

Issue #115, opened 2022-02-18
Add missing docs and examples


orottier:
We aim to have an example usage per audio node type, and all methods documented.
Check e.g. DelayNode for a decent top level doc and example usage: https://docs.rs/web-audio-api/latest/web_audio_api/node/struct.DelayNode.html
Notice we currently have no documentation and examples for the delaynode methods. In this case, I think the top level example suffices.

PR #586, opened 2026-04-28
docs(node): add top-level + method docs for GainNode and ConstantSourceNode

mvanhorn:
## What

Brings two of the smaller node modules up to the doc standard already used by `DelayNode` and `StereoPannerNode` (which #115 cites as the reference quality bar).

### Before

```rust
/// AudioNode for volume control
pub struct GainNode { ... }

impl GainNode {
    pub fn new<C: BaseAudioContext>(context: &C, options: GainOptions) -> Self { ... }
    pub fn gain(&self) -> &AudioParam { ... }
}
```

### After

- Top-level explainer for `GainNode` covering what the node is good for, MDN + spec + see-also links, a runnable `# Usage` example (one-second linear fade-in), and `# Examples` pointing at `amplitude_modulation` / `feedback_delay`.
- Method docs on `GainNode::new`, `GainNode::gain`, `ConstantSourceNode::new`, and `ConstantSourceNode::offset` covering defaults, typical usage, and the automation surface (a-rate, ramp methods).
- Field-level docstrings on private struct members of `GainNode` to mirror `BiquadFilterNode` / `StereoPannerNode`.

## Why

#115 calls out `DelayNode` as the reference style and notes most other nodes haven't reached that bar. `GainNode` was the most barebones (single-line struct doc, no method docs at all), so it was the highest-leverage place to start. `ConstantSourceNode` already had a top-level example but its two methods were undocumented; folded that in to keep this PR a single coherent unit.

Other nodes (e.g. `analyser`, `panner`) have many more methods and would each be their own follow-up PR. Happy to take a swing at one of those next if helpful.

## Test plan

- [x] `cargo doc --no-default-features --no-deps` -- 0 errors, builds cleanly
- [x] `cargo test --no-default-features --doc` -- 29 passed (the new `# Usage` example compiles via `no_run` like the existing ones)

Refs #115

### DIFF
--- a/src/node/constant_source.rs
+++ b/src/node/constant_source.rs
@@ -128,6 +128,15 @@ impl AudioScheduledSourceNode for ConstantSourceNode {
 }
 
 impl ConstantSourceNode {
+    /// Constructs a new `ConstantSourceNode` from explicit options.
+    ///
+    /// Most callers should prefer [`BaseAudioContext::create_constant_source`],
+    /// which applies the spec defaults (`offset = 1.0`).
+    ///
+    /// # Arguments
+    ///
+    /// * `context` - audio context in which the audio node will live
+    /// * `options` - initial value of the offset parameter
     pub fn new<C: BaseAudioContext>(context: &C, options: ConstantSourceOptions) -> Self {
         context.base().register(move |registration| {
             let ConstantSourceOptions { offset } = options;
@@ -160,6 +169,13 @@ impl ConstantSourceNode {
         })
     }
 
+    /// Returns the offset `AudioParam`.
+    ///
+    /// The default value is `1.0`. Treat the node as a *constructible*
+    /// `AudioParam`: connect `offset()` to one or more sink params (e.g.
+    /// the `gain` of several `GainNode`s) and automate `offset()` once to
+    /// drive every connected sink in lockstep.
+    #[must_use]
     pub fn offset(&self) -> &AudioParam {
         &self.offset
     }
--- a/src/node/constant_source.rs
+++ b/src/node/constant_source.rs
@@ -37,8 +37,9 @@ enum Schedule {
     Stop(f64),
 }
 
-/// Audio source whose output is nominally a constant value. A `ConstantSourceNode`
-/// can be used as a constructible `AudioParam` by automating the value of its offset.
+/// Audio source whose output is nominally a constant value.
+///
+/// Can be used as a constructible `AudioParam` by automating the value of its offset.
 ///
 /// - MDN documentation: <https://developer.mozilla.org/en-US/docs/Web/API/ConstantSourceNode>
 /// - specification: <https://webaudio.github.io/web-audio-api/#ConstantSourceNode>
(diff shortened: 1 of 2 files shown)


src/node/constant_source.rs
```diff
@@ -128,6 +128,15 @@ impl AudioScheduledSourceNode for ConstantSourceNode {
 }
 
 impl ConstantSourceNode {
+    /// Constructs a new `ConstantSourceNode` from explicit options.
+    ///
+    /// Most callers should prefer [`BaseAudioContext::create_constant_source`],
```
orottier: Better to frame as an 'alternative' than a pref

src/node/constant_source.rs
```diff
@@ -160,6 +169,13 @@ impl ConstantSourceNode {
         })
     }
 
+    /// Returns the offset `AudioParam`.
+    ///
+    /// The default value is `1.0`. Treat the node as a *constructible*
```
orottier: Some slop in here, keep it tight

src/node/gain.rs
```diff
@@ -25,11 +25,50 @@ impl Default for GainOptions {
     }
 }
 
-/// AudioNode for volume control
+/// `GainNode` applies a single gain (volume) value to its incoming audio
```
orottier: For rustdocs, put linebreaks after the first line so it shows that first line as summary

src/node/gain.rs
```diff
@@ -25,11 +25,50 @@ impl Default for GainOptions {
...
+///
+/// osc.connect(&gain);
+/// gain.connect(&context.destination());
+/// osc.start();
+/// ```
+///
+/// # Examples
+///
+/// - `cargo run --release --example amplitude_modulation`
+/// - `cargo run --release --example feedback_delay`
```
orottier: not the best example

src/node/gain.rs
```diff
@@ -52,6 +91,15 @@ impl AudioNode for GainNode {
 }
 
 impl GainNode {
+    /// Constructs a new `GainNode` from explicit options.
+    ///
+    /// Most callers should prefer [`BaseAudioContext::create_gain`], which
```
orottier: Better to frame as an 'alternative' than a pref

orottier (commented):
Can I get another pass? The CI failure is unrelated

orottier (changes requested):

mvanhorn:
Thanks @orottier, pushed 283077b. Walked through your inline comments:

- gain.rs line 28 / constant_source.rs line 40: split the summary line out so rustdoc treats the first sentence as the index summary.
- gain.rs line 63: dropped `feedback_delay` from the examples list; kept the more direct `amplitude_modulation`.
- gain.rs line 96 / constant_source.rs line 133: reframed `BaseAudioContext::create_gain` / `create_constant_source` as alternatives that apply spec defaults rather than the preferred path.
- constant_source.rs line 174 (`offset()`): tightened to a one-line summary plus one sentence on use as a constructible AudioParam. Dropped the *constructible* emphasis and the GainNode example you flagged as slop.

`cargo check` and `cargo doc --no-deps -D warnings` both pass locally.

orottier:
Thanks for your contribution!

PR #588, opened 2026-05-08
Fix StereoPannerNode: accurate computation of right and left gains

b-ma:
This removes the table lookup that was both wrong and inaccurate enough

The WPT tests that are passing now:
- the-stereopanner-interface/no-dezippering.html
- the-stereopanner-interface/stereopannernode-panning.html

Which means the `StereoPannerNode` is fully compliant:

```
  RESULTS:
  - # pass: 121
  - # fail: 0
  - # type error issues: 0
```


orottier:
Funny 🤔

PR #590, opened 2026-05-12
fix: validate sample_rate in AudioContext constructor (#589)

### LINKED ISSUE #589: Validate `sample_rate` in AudioContext constructor
Somehow we only validate this way down the line when bootstrapping cpal, so this tests actually does not panic.

```
    #[test]
    #[should_panic]
    fn test_construct_bad_sample_rate() {
        let options = AudioContextOptions {
            sample_rate: Some(0.),
            sink_id: "none".into(),
            ..AudioContextOptions::default()
        };
        let _ = AudioContext::new(options);
    }
```

kimjune01:
## Summary

- Validates `sample_rate` in `AudioContext::try_new_inner` before building the output stream, returning `Err(InvalidSampleRate)` for values outside [3000, 768000] per the W3C Web Audio API spec
- `AudioContext::new` panics on invalid sample rate (consistent with existing `SinkNotFound` pattern)
- Adds 9 tests covering zero, too-low, too-high, NaN, Infinity, negative infinity, negative values, boundary values (3000/768000), and the `try_new` error path

Fixes #589

## Test plan

- [x] `cargo test` -- 438/438 tests pass
- [x] Boundary tests verify both 3000.0 and 768000.0 are accepted
- [x] NaN, Inf, -Inf all correctly rejected (the `>=`/`<=` comparison handles NaN by returning false)

### DIFF
--- a/src/context/online.rs
+++ b/src/context/online.rs
@@ -33,6 +33,7 @@ fn is_valid_sink_id(sink_id: &str) -> bool {
 #[derive(Debug)]
 enum AudioContextError {
     SinkNotFound { sink_id: String },
+    InvalidSampleRate { sample_rate: f32 },
     Backend { error: io::AudioBackendError },
 }
 
@@ -41,6 +42,12 @@ impl std::fmt::Display for AudioContextError {
         match self {
             Self::SinkNotFound { sink_id } => {
                 write!(f, "NotFoundError - Invalid sinkId: {sink_id:?}")
+            }
+            Self::InvalidSampleRate { sample_rate } => {
+                write!(
+                    f,
+                    "NotSupportedError - Invalid sample rate: {sample_rate}, should be in the range [3000.0, 768000.0]"
+                )
             }
             Self::Backend { error } => write!(f, "InvalidStateError - {error}"),
         }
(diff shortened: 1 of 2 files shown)


src/context/online.rs
```diff
@@ -225,6 +234,16 @@ impl AudioContext {
             });
         }
 
+        // Validate sample_rate if provided
+        // https://webaudio.github.io/web-audio-api/#sample-rates
+        if let Some(sample_rate) = options.sample_rate {
```
orottier, line 239: The `assert_valid_sample_rate` method panics, which is undesired, so I get that the method is inlined here. But let's refactor, add an is_valid_sample_rate method and make this call site andn `assert_valid_sample_rate` use that. This avoid sprawling these magic numbers around the codebase

src/context/online.rs
```diff
@@ -906,6 +925,114 @@ mod tests {
         require_send_sync(context.close());
     }
 
+    #[test]
```
orottier: One test will suffice. With a bad sample rate. We have plenty of other tests validating rates in the valid range

orottier (changes requested):
Thanks could you pick up these comments?

kimjune01:
Thanks @orottier — both addressed:

- Magic-number sprawl: extracted `is_valid_sample_rate` (plus `MIN_SAMPLE_RATE` / `MAX_SAMPLE_RATE` consts) in `lib.rs`; `assert_valid_sample_rate` and the `AudioContext::try_new` validation now both call it. Range lives in one place. (5852c62)
- Test sprawl: collapsed the per-edge `should_panic` tests (zero / too_low / too_high / NaN / infinities / negative) and the redundant valid-boundary test down to a single `test_try_new_invalid_sample_rate` exercising the new error path. (6b63cb6)

`cargo test --lib` green.

kimjune01:
Both addressed in the latest pushes:

- 5852c62 extracts `is_valid_sample_rate` and reuses it from `assert_valid_sample_rate` and the `AudioContext::try_new` call site, so the spec range constants live in one place.
- 6b63cb6 collapses to a single `test_try_new_invalid_sample_rate` with `sample_rate: Some(0.)`. Valid-range cases are covered elsewhere.

Ready for another look.

kimjune01:
PR lacks a test exercising the change, so drafting. Please close or take it over.

orottier:
Thanks for your contribution!

Issue #593, opened 2026-05-18
Update symphonia to version 0.6


orottier:
@dependabot ignore this minor version

It's a big change: https://github.com/pdeljanov/Symphonia/blob/main/docs/guides/migration/0p6.md let's track it in a separate issue

_Originally posted by @orottier in https://github.com/orottier/web-audio-api-rs/issues/591#issuecomment-4475978989_

b-ma:
With this update, would be a nice occasion to:
- support more format (e.g. aiff which is heavily used by mac users...) and 
- see if we can remove the silence at the beginning of mp3

orottier:
Sure, good points.
Also I think it would be a good idea to remove our codec feature zoo
```
default = ["mp3", "ogg", "flac", "wav", "m4a", "alac", "cpal"]
mp3 = ["symphonia/mp3", "creek/decode-mp3"]
ogg = [
    "symphonia/ogg",
    "symphonia/vorbis",
    "creek/decode-ogg",
    "creek/decode-vorbis",
]
flac = ["symphonia/flac", "creek/decode-flac"]
wav = ["symphonia/wav", "symphonia/pcm", "creek/decode-wav", "creek/decode-pcm"]
aac = ["symphonia/aac", "creek/decode-aac"]
m4a = ["aac", "symphonia/isomp4", "creek/decode-isomp4"]
alac = [
    "symphonia/alac",
    "symphonia/isomp4",
    "creek/decode-alac",
    "creek/decode-isomp4",
]
```
It is not very easy and sustainable, and I think we are not a lightweight lib anyway that benefits from scraping a few lines of code..

orottier:
> see if we can remove the silence at the beginning of mp3

It seems this is fixed in 0.6 - or at least the decoded mp3's are a bit shorter. Perhaps you can check if you have time?

> support more format (e.g. aiff which is heavily used by mac users...) 

works now - we enable all symphonia features
```
> playing file: "samples/sample.aiff"
> duration: 3.2241950113378683
> length: 142187
> channels: 2
> sample rate: 44100.0
```

b-ma:
Cool thanks ! I will have a look on the mp3 (hopefully today or tomorrow)

b-ma:
Hey, I had a look and yes `aiff` works on my side too and both `ogg` and `mp3` looks to have the exact same length / duration as the `wav` file now, pretty cool!

[...]

PR #594, opened 2026-05-20
Symphonia v0.6, include all decoders/encoders always, refactors

orottier:
Note that creek still pulls in 0.5 - I will try to make a patch on that project


orottier:
Ready for review. Fixes #593

b-ma:
Looks good to me (as far I understand :)

Maybe add 'CAF' to the `.codespellignore` file?

PR #595, opened 2026-05-23
Reduce oscillator sine table from 8192 to 2048 samples

### LINKED ISSUE #309: Improve pre-computed `SINETABLE`
```
let table: Vec<f32> = (0..TABLE_LENGTH_USIZE)
    .map(|x| ((x as f32) * 2.0 * PI * (1. / (TABLE_LENGTH_F32))).sin())
    .collect();
```

This computes the full sine wave from t=0 to t=2PI.
We can use the symmetries of the sine wave to store only the part of t=0 to t=PI/2 and derive the other ranges from it:
https://www.electronics-notes.com/images/waveform-sine-wave-02.png

This will reduce the general memory usage so we can utilize the caches for other things which may improve the libs performance

orottier:
Fixes #309

2048 samples and linear interpolation are good enough to reach WPT precision requirements. The smaller table size will reduce the cache pressure.

Keep the 8192-sample table length for PeriodicWave rendering (per the spec). Removed the consts and methods from src/node/mod.rs

Benchmarking showed that using the full 2 pi phase in the lookup table is fastest - so we do not use cleverness around the sine function symmetry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Decoders are enabled always - the legacy cargo features mp3/ogg/flac/wav/.. are now no-op
 - Added .aiff decoder support
 - Fix: mp3 decoding would sometimes insert leading silent frames
+- Improved sine oscillator performance
 
 ## Version 1.4.0 (2026-05-18)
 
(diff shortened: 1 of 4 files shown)


orottier:
Draft because I need to run WPT again - my workspace is a total mess atm

orottier:
Yeah WPT is stable
```
  RESULTS:
  - # pass: 98
  - # fail: 6
  - # type error issues: 0
  ```

PR #597, opened 2026-05-24
Fix - Biquad filter coefs and stability

b-ma:
This review the way the coefs are calculated (especially at limits) using the functions from WPT (cf. /wpt/webaudio/resources/biquad-filters.js) 

Also check and flush to zero any non normal value in `process` which should (hopefully) improve stability. Maybe it can introduce some discontinuity, but this is anyway better than the application becoming silent forever due to some NaN in filter history.

### before
```
RESULTS:
  - # pass: 266
  - # fail: 17
  - # type error issues: 0
```

### after
```
RESULTS:
- # pass: 283
- # fail: 0
- # type error issues: 0
```


orottier:
Monkey sees WPT improvement, monkey merges 

However, let's /bench first

orottier:
Good enough for me

Thanks!

PR #600, opened 2026-05-25
Fix - AudioWorklet `inputs[n].len()` should be zero when source is not actively processing

b-ma:
cf. https://webaudio.github.io/web-audio-api/#audioworkletprocess-callback-parameters 

Allows to notify the worklet that it's source is not [actively processing](https://webaudio.github.io/web-audio-api/#actively-processing), e.g.

```rs
fn process(/* ...*/) -> bool {
  if inputs[0].is_empty() {
    // do nothing
  }
}
```

Semantic is kind of similar to `AudioRenderQuantum::is_silent()`


b-ma:
hum weird... this ci fail does not seem to be related to this PR at all

PR #602, opened 2026-05-26
WIP - experimenting with  `device disconnect`

orottier:
Create a loopback virtual device
Run `RUST_LOG=debug cargo run --example sink_id`
Choose virtual device
While running, remove the device in loopback
Observe output

Bluetooth disconnect does not trigger cpal err_fn
Changing output device from OSX settings does not trigger either

Perhaps interesting read
https://github.com/RustAudio/cpal/issues/373

Needs more polishing!

### DIFF
--- a/examples/sink_id.rs
+++ b/examples/sink_id.rs
@@ -57,10 +57,16 @@ fn main() {
     osc.start();
 
     loop {
-	println!();
-        loop {
-		println!("Current time is now {:<3}\r", context.current_time());
-                std::thread::sleep(std::time::Duration::from_millis(100));
+        let sink_id = ask_sink_id();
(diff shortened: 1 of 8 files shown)


orottier:
`cpal` on OSX does strange things. When I use the default sink (`''`) and start with speakers, then plug in my headphones - the sounds moves gracefully and the render thread needs no reboot. I can even plug in my headphones back in and the sound move again.
The other way around does not work, start with headphones, unplug - cpal registers no stream error even.

orottier:
Alright, none of this is needed because #616  handle default device following very well.
Although this was nice research, the err_fn improvements should definitely land in a future version

PR #603, opened 2026-05-26
Fix `ConvolverNode` active processing

b-ma:
Just like the title indicates :) 

Mark the node as not actively processsing (i.e. `output.make_silent()`) if input is silent and we reached the end of the convolution.

Fix `wpt/webaudio/the-audio-api/the-convolvernode-interface/active-processing.https.html` (visible only given #600, as wpt relies on this to test active processing)

### DIFF
--- a/src/node/convolver.rs
+++ b/src/node/convolver.rs
@@ -354,15 +354,15 @@ impl AudioProcessor for ConvolverRenderer {
 
         // handle tail time and active processing, return early if input is silent
         // and we reached the end of the convolution
-        let is_silent_input = input.is_silent();
-
-        if is_silent_input {
+        if input.is_silent() {
             if self.tail_count > self.impulse_length {
                 output.make_silent();
                 return false;
             }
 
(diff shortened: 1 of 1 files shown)


src/node/convolver.rs
```diff
@@ -351,7 +351,19 @@ impl AudioProcessor for ConvolverRenderer {
         // single input/output node
         let input = &inputs[0];
         let output = &mut outputs[0];
-        output.force_mono();
+
+        // handle tail time and active processing
+        // return early if input is silent and we reached the end of the convolution
+        if input.is_silent() {
+            if self.tail_count > self.impulse_length {
```
orottier: This might need to be `>=`, the earlier check was
`return self.tail_count < self.impulse_length;`

orottier:
```suggestion
            if self.tail_count >= self.impulse_length {
```

orottier:
Thanks! Also for the previous PR

PR #605, opened 2026-05-29
Some fixes on the `AudioBufferSourceNode`

b-ma:
Fixes all wpt test in:

- the-audiobuffersourcenode-interface/audiobuffersource-duration-loop.html
- the-audiobuffersourcenode-interface/audiobuffersource-playbackrate-negative.html
- the-audiobuffersourcenode-interface/audiobuffersource-start-null-buffer.html

Except for the last one, it mainly consists in recovering from floating point errors

### before
```
- # pass: 292
- # fail: 19
- # type error issues: 0
```

### after
```
- # pass: 303
- # fail: 7
- # type error issues: 0
```


orottier:
What a minefield :/ but alright. I wrote some tests and think we have another reverse playback issue. PR incoming

b-ma:
> What a minefield :/

Haha sure :)

PR #606, opened 2026-05-31
Add AudioBufferSourceNode regression tests

orottier:
And fix negative playback not adding to duration

Spec text 

The duration parameter describes the duration of sound to be played, expressed as seconds of total buffer content to be output, including any whole or partial loop iterations. The units of duration are independent of the effects of playbackRate. For example, a duration of 5 seconds with a playback rate of 0.5 will output 5 seconds of buffer content at half speed, producing 10 seconds of audible output. A RangeError exception MUST be thrown if duration is negative

### DIFF
--- a/src/node/audio_buffer_source.rs
+++ b/src/node/audio_buffer_source.rs
@@ -686,7 +686,7 @@ impl AudioProcessor for AudioBufferSourceRenderer {
                     }
 
                     buffer_time = self.offset;
-                    self.render_state.buffer_time_elapsed = delta * computed_playback_rate;
+                    self.render_state.buffer_time_elapsed = (delta * computed_playback_rate).abs();
                     self.render_state.started = true;
                 }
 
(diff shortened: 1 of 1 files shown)


b-ma:
Thanks for the regression tests (...I have been a bit lazy on that :)

And really nice catch! 

I just ported the test code in JS to run it chrome and firefox and interestingly they both output complete silence, they just behave as if playbackRate was not taken into account, that's weird... maybe due to the fact that run it in the console or maybe I messed up something in my code, but cannot figure what it could be:

```js
const sampleRate = 48000;
const context = new OfflineAudioContext(1, 128, sampleRate);

const buffer = context.createBuffer(1, 5, sampleRate);
buffer.copyToChannel(new Float32Array([1, 2, 3, 4, 5]), 0);

const src = context.createBufferSource();
src.connect(context.destination);
src.buffer = buffer;
src.playbackRate.value = -1;
src.start(0, buffer.duration, 2 / sampleRate);

const result = await context.startRendering();
console.log(result.getChannelData(0));
```

In any case, this is not covered in `wpt/webaudio/the-audiobuffersourcenode-interface/audiobuffersource-playbackrate-negative.html`, could be a good addition I think

b-ma:
Hum actually they are not very good at negative playback rates : https://wpt.fyi/results/webaudio/the-audio-api/the-audiobuffersourcenode-interface/audiobuffersource-playbackrate-negative.html?label=experimental&label=master&aligned

PR #607, opened 2026-06-03
Fix - IIRFilterNode with feedback and feedforward coefs of different lengths

b-ma:
(+ prevent propagation of denormals into the graph)

Now fully WPT compliant!

```
RESULTS:
- # pass: 128
- # fail: 0
- # type error issues: 0
 ```

### DIFF
--- a/src/node/iir_filter.rs
+++ b/src/node/iir_filter.rs
@@ -386,7 +386,11 @@ impl AudioProcessor for IirFilterRenderer {
                 let input = f64::from(i);
                 let b0 = self.norm_coeffs[0].0;
                 let last_state = channel_state[0];
-                let mut output = b0.mul_add(input, last_state);
+                let mut output = b0.mul_add(input, last_state);// do not propagate denormals into graph and state
+
+                if !output.is_normal() {
+                    output = 0.;
+                }
 
                 // update states for next call
                 for (i, (b, a)) in self.norm_coeffs.iter().skip(1).enumerate() {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/node/iir_filter.rs [hunk 1]
```diff
@@ -386,7 +386,11 @@ impl AudioProcessor for IirFilterRenderer {
                 let input = f64::from(i);
                 let b0 = self.norm_coeffs[0].0;
                 let last_state = channel_state[0];
-                let mut output = b0.mul_add(input, last_state);
+                let mut output = b0.mul_add(input, last_state);// do not propagate denormals into graph and state
```
orottier: I thought we flushed denormals already on the CPU level? So this is basically an extra guard for NaN and Infinity? It's quite a heavy performance hit with 11% extra instruction count but I'm fine with taking it. NaNs in your graph are terrible

b-ma: Hum, I'm not sure actually (is it done on all CPU?), we should investigate this question extensively at some point (...or just wait Rust for proper fast math in stable :) 
We've got some issues with unstable biquads (I think, I don't see any other place this could come from), breaking the sound for the whole machine: all apps went silent until we killed the faulty one..., so I've became a bit defensive with these at that moment (but maybe I'm overreacting too... :)

orottier: flushing denormals is good for performance, flushing Inf/NaN will fix silence, I think.
We flush denormals on aarch64/x86_64/x86 globally, not on other platforms (it is not easy)
So I think your addition is good!

orottier:
This seems within noise so I would say keep that, because it skips allocations in the render thread

b-ma:
Ok, all good to me, wpt is still happy for both this one and #610, thanks for the improvements

PR #608, opened 2026-06-03
Fix: mixing rules for non standard speakers layout i.e. 3 and 5

b-ma:
Fix all tests in `the-audionode-interface/audionode-channel-rules.html` (and simplifies the code by the way)

I can't really understand what happened here, weird... maybe the spec changed (but the strange rules were not even in the tests...). Anyway, it works now :)


orottier:
Weird! How did we get to those rules then? Anyway

PR #609, opened 2026-06-04
Fix - `OscillatorNode` compliance

b-ma:
Small fixes for the `OscillatorNode`
- computed frequencies >= nyquist should output 0.
- proper support for negative frequencies

Now fully WPT compliant

```
RESULTS:
- # pass: 104
- # fail: 0
- # type error issues: 0
```

### DIFF
--- a/src/node/oscillator.rs
+++ b/src/node/oscillator.rs
@@ -27,11 +27,11 @@ fn precomputed_sine_table() -> &'static [f32] {
     })
 }
 
-fn get_phase_incr(freq: f32, detune: f32, sample_rate: f64) -> f64 {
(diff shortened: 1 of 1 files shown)


orottier:
Good stuff. I'm working on some code quality on top of this, and I'm investigating computedFreq moving over `-nquist`

b-ma:
Ok thanks, I just realise it can't be simplified indeed :) (...we could just fill output with zeros in fast path for example)

> and I'm investigating computedFreq moving over -nquist

Haha nice catch, forgot this one

orottier:
Superseded by #610

PR #610, opened 2026-06-09
More OscillatorNode work

orottier:
Sorry I went a bit wild. Built further upon your work
- handle -nyquist values
- fix as possible panic path with large phase incr value
- add a fast path for not-automation frequence/detune
- added an even faster path for 'fully active' (oscillator is ringing during the whole render block)
- float value calculation optimization
- some tests and a benchmark for the slow path


b-ma:
Haha given the numbers, this seems welcome! I'll have a look tomorrow

b-ma:
All good for me, thanks!

PR #615, opened 2026-06-15
Revamp AudioContext::run_diagnostics (structured output, feature-gated)

### LINKED ISSUE #611: Improve `AudioContext::run_diagnostics` utility
Hey,

As discussed in #565, now that we have the spec compliant `AudioPlaybackStats` and that `AudioRenderCapacity` is officially abandoned, we could abuse it to report useful debug information. 

So far, the main thing that would be really helpful, especially within Node.js, would be to track the number of nodes currently in the graph. With the GC, the behavior is quite different there, and this really helped me to track down memory issues once or twice.

This could be hidden behind an environment variable (maybe the whole `AudioRenderCapacity` could be in next major version)

orottier:
Fixes #611

### DIFF
--- a/src/context/online.rs
+++ b/src/context/online.rs
@@ -6,7 +6,9 @@ use std::sync::{Arc, Mutex};
 #[cfg(feature = "diagnostics")]
 use crate::context::{AudioBackendDiagnostics, AudioContextDiagnostics};
 use crate::context::{AudioContextState, BaseAudioContext, ConcreteBaseAudioContext};
-use crate::events::{EventDispatch, EventHandler, EventLoop, EventPayload, EventType};
+#[cfg(feature = "diagnostics")]
+use crate::events::EventPayload;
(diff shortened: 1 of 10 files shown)


src/context/online.rs
```diff
@@ -3,6 +3,8 @@ use std::error::Error;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 
+#[cfg(feature = "diagnostics")]
+use crate::context::{AudioBackendDiagnostics, AudioContextDiagnostics};
 use crate::context::{AudioContextState, BaseAudioContext, ConcreteBaseAudioContext};
 use crate::events::{EventDispatch, EventHandler, EventLoop, EventPayload, EventType};
```
b-ma: `EventPayload` clippy seems to complain here

b-ma:
Looks pretty cool!

PR #617, opened 2026-06-15
Feat - add native pipewire host support

### LINKED ISSUE #614: Implement cpal PipeWire host
Version 0.18 of cpal seems to provide proper PipeWire integration (https://github.com/RustAudio/cpal/blob/master/CHANGELOG.md#0180---2026-06-06):

> PipeWire: New host for Linux and BSD via the native PipeWire API. 

As Pipewire is now the default on most linux distribution (debian, unbuntu, fedora, etc...), and will probably stand as the default standard for linux audio in the future, would be great to provide a backend for it

b-ma:
Fixes  #614, seems to work as expected on my side, tested on RPi 4 with latest Raspberry Pi OS (based on debian Trixie)

(I also checked the `realtime` flag but my system is quite configured already and I'm not sure it does something, needs further investigations)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,7 +22,7 @@ jobs:
     steps:
       - name: Install ALSA and Jack dependencies
         run: |
-          sudo apt-get update && sudo apt-get install -y libasound2-dev libjack-jackd2-dev cmake
+          sudo apt-get update && sudo apt-get install -y libasound2-dev libjack-jackd2-dev libpipewire-0.3-dev cmake clang
 
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@stable
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Install ALSA and Jack dependencies
+      - name: Install ALSA, Jack and PipeWire dependencies
         run: |
           sudo apt-get update && sudo apt-get install -y libasound2-dev libjack-jackd2-dev libpipewire-0.3-dev cmake clang
 
@@ -51,9 +51,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Install ALSA and Jack dependencies
+      - name: Install ALSA, Jack and PipeWire dependencies
         run: |
-          sudo apt-get update && sudo apt-get install -y libasound2-dev libjack-jackd2-dev cmake
+          sudo apt-get update && sudo apt-get install -y libasound2-dev libjack-jackd2-dev libpipewire-0.3-dev cmake clang
 
       - name: Install Rust toolchain
         # Aligned with `rust-version` in `Cargo.toml`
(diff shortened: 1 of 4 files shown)


orottier:
Nice work!
I am a bit wary of all these optional features that we add. What if we add cpal pipewire and jack always? That will not break anything right? Or do they need to be installed on linux in order to compile?

b-ma:
Hum, actually you need to install some lib*-dev with `apt` or something in order to compile (which is the case also for ALSA by the way):

```
# ALSA
sudo apt install libasound2-dev
# Jack
sudo apt install libjack-jackd2-dev
# PipeWire
sudo apt install libpipewire-0.3-dev clang
```

(we could put that in the README probably)

> What if we add cpal pipewire and jack always?

I don't have a strong opinion on that, it seems to me that the whole Linux ecosystem is moving to PipeWire now, but this is quite a slow process and still in the middle of the transition I think.

As we didn't provide the pulseaudio host so far, the only alternative to ALSA we had so far was Jack (which is really geared toward realtime and professional use cases). So I guess we have kind of two paths now:
- provide support for all hosts: ALSA, pulseaudio, Jack, PipeWire (backward compatible, quite hard to maintain and above all to test I think)
- drop Jack support and go only for PipeWire as default with a fallback to ALSA (more opinionated, still better than what we have now for regular users, but probably requires a major version)

I would go the second one personally I thing (Linux audio was such a mess before PipeWire that we should spend too much time maintaining it :)

orottier:
Okay I had a look too, it appears we cannot just throw host features to the wall and see what sticks. Users will run into compile/runtime errors. So our shape with passing features to cpal is right. Sadly we cannot publish a pipewire powered npm package then..

During my tests I prepared another commit
```
    cpal host selection choices

    - prefer pipewire over jack if both features are selected
    - fix compile error outside linux when jack/pipewire feature is enabled
    - log and fall back to default host when specific host is not available
 ```
    
But maybe we do prefer compile errors when users select an unavailable host? That makes it nice explicit - you asked for jack but we cant honor it. What do you think?

b-ma:
Hey, 

I think I got your point, but then I would say the best thing would be to think about the different users (I don't know how much work it is and maybe it is unrealistic... but let's put this aside for now):

Let's consider a developper user of the lib ("user A" e.g. developing some game or so), this user can't afford to release different builds of her app (built with different config flags) depending of the host installed on the end-user (user B) machine (maybe B doesn't even have any idea about his audio backend). 

Then, maybe this is our job (and not the job of A) to propose some best effort to support any host installed on B machine? 

If this holds true, I think we should therefore go for something like this:
- try PipeWire (kind of future standard as it seems, modern and efficient), 
- fallback to Jack (older system, but user B is actively interested in having efficient audio)
- fallback to PulseAudio (older system too, but user B belongs to a more general audience)
- fallback to Alsa (last effort, only one app can hold this host at the same time or it will crash)
- fallback to runtime error or no sound, e.g. similar to `sinkId: { type:'none' }` (if Alsa doesn't work, nothing else can work anyway...)

Then this is just a matter of properly documenting that for Linux builds you need to have all the `lib*-dev` installed, which is not something so terrible for a user of kind A I think

> Sadly we cannot publish a pipewire powered npm package then..

No worries, I'm a bit more agressive with such things and already publish the linux version with the jack feature on, I could just switch the flag to pipewire :)

_**edit:** maybe looks like the same question as for the different audio codecs finally: no gated feature, just provide support for everything we can_

PR #623, opened 2026-07-15
Fix flaky test_none_sink_id under load

cmzy:
## Problem

`cargo test --all-targets` intermittently fails with:

```
---- test_none_sink_id stdout ----
thread 'test_none_sink_id' panicked at tests/online.rs:126:5:
timed out waiting for started state change
```

The test passes when run on its own but fails when the machine is busy, which is exactly what `--all-targets` does: it builds and runs several test binaries concurrently.

## Cause

`test_none_sink_id` installs its `onstatechange` handler *after* constructing the `AudioContext`, then blocks waiting for the `suspended -> running` event:

```rust
let context = AudioContext::new(options);
// ...
context.set_onstatechange(move |_| { let _ = state_change_tx.send(()); });
assert!(state_change_rx.recv_timeout(TEST_TIMEOUT).is_ok(), "timed out waiting for started state change");
```

That transition is emitted from the render thread the moment it adopts the initial graph (`Startup` message -> `set_state(Running)`). For the `none` backend the render thread starts in the running state, so this happens within the first render quantum. If it fires before the handler is installed, `EventLoop::handle_event` finds no handler registered and drops the event. It is never re-sent, so `recv_timeout` waits out the full 3s and the test fails.

Under no load the main thread installs the handler within microseconds, well before the render thread's first quantum, so the event is virtually always caught. Under load the main thread can be preempted in the window between `AudioContext::new` and `set_onstatechange`, letting the render/event threads deliver-and-drop the event first.

This is reproducible on demand by sleeping 100ms between construction and `set_onstatechange`: the started event is then always dropped and the test fails every time.

## Fix

Wait for the observable running state instead of the one-shot event:

```rust
let deadline = Instant::now() + TEST_TIMEOUT;
while context.state() != AudioContextState::Running {
    assert!(Instant::now() < deadline, "timed out waiting for started state change");
    thread::sleep(Duration::from_millis(5));
}
```

The later `suspend`/`resume`/`close` transitions are still asserted through the event channel. Those are triggered by explicit calls made *after* the handler is installed, so they are not subject to the same race and keep exercising the `onstatechange` delivery path.

## Verification

- `cargo test --all-targets` is green.
- Running 16 copies of the `online` test binary in parallel to force contention: all pass (previously ~2 in 8 failed, and even the passers sat right at the 3s timeout).
- With a 100ms sleep injected to guarantee the started event is dropped: fails every time before this change, passes every time after.


cmzy:
Opened by mistake — this only carried a test-stability tweak. Superseded by a PR from the fixes/v1.6.0 branch.

orottier:
Merged via 01ff21dd4e3f05

PR #624, opened 2026-07-15
Fixes for v1.6.0: hang safety, AudioParam automation, and graph handling

cmzy:
This branch collects a set of fixes on top of `main`, grouped by area below. Each change is a separate commit with a full description and, where applicable, tests.

### Hang safety

- **Never block the control thread indefinitely on a stalled render thread.** `send_control_msg` and the suspend/resume/close acknowledgements now wait on render-thread liveness (playhead progress) rather than blocking forever when the output device disappears mid-session.
- **Do not hang or silently kill the context when a sink change fails.**
- **Never block in `MediaStreamAudioDestinationNode`'s stream provider.**
- **Fall back to silence-flag suspension when native pause is unsupported**, so `suspend` still takes effect on backends without a native pause.

### AudioParam automation

- **Compute k-rate params at the first frame of the render quantum.**
- **Settle first-frame events before the constant-block check**, so a block whose only event lands on its first frame is classified as constant and produces a single-valued array (matches the WPT `audioworklet-audioparam-size` expectation).
- **Fix `cancelAndHoldAtTime` interactions with `setTargetAtTime` and `setValueCurveAtTime`.**
- **Cancel the whole `setValueCurve` when `cancelScheduledValues` lands inside it.**
- **Avoid FMA in linear-ramp sample computation** for reproducible ramp values.
- **Include source params in the EqualPower single-valued check.**

### Graph and rendering

- **Ignore repeated `connect()` calls with identical termini** instead of creating duplicate edges.
- **Use sets for visited/cycle bookkeeping in graph ordering.**
- **Restrict the identical-channels fast path in `add()` to mono/stereo.**

### API

- **Add an incremental rendering API to `OfflineAudioContext`.**

### Misc

- **Use the standard separator in stop-before-start panic messages.**

### DIFF
--- a/src/context/concrete_base.rs
+++ b/src/context/concrete_base.rs
@@ -13,10 +13,11 @@ use crate::spatial::AudioListenerParams;
 
 use crate::AudioListener;
 
-use crossbeam_channel::{SendError, Sender};
+use crossbeam_channel::{Receiver, RecvTimeoutError, SendError, SendTimeoutError, Sender};
 use std::collections::HashSet;
 use std::sync::atomic::{AtomicU64, AtomicU8, Ordering};
 use std::sync::{Arc, Mutex, RwLock, RwLockWriteGuard};
+use std::time::Duration;
 
 /// This struct assigns new [`AudioNodeId`]s for [`AudioNode`]s
 ///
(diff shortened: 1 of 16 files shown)


orottier:
Hi @cmzy , thank you for your contribution. These large AI diffs are impossible to review. However I think there are some good fixes in there. Could you split the work in separate and independent PRs, so we can go over them one by one.
For incremental rendering, we would have to follow the specification that is currently discussed over at the official web audio api repo.

cmzy:
Hi @orottier Thank you for your contribution! I'm currently building a game engine that uses your project as its audio backend. However, while running the W3C Web Audio test suite, I encountered these issues. Although the fixes were generated with AI assistance, I have verified that the underlying problems are real and valid, so I went ahead and fixed them. I will be submitting a separate PR for each commit.

PR #628, opened 2026-07-16
Fall back to silence-flag suspension when native pause is unsupported

cmzy:
CpalBackend::suspend() forwarded directly to cpal's Stream::pause(),
which on Linux maps to snd_pcm_pause. Pausing is an optional ALSA
hardware feature: dmix, the null device and many plug devices reject it
with "Unknown errno (77)". Two user-visible consequences:

1. suspend() errors out on such devices, so AudioContext::suspend_sync()
   fails on perfectly normal configurations.
2. set_sink_id_sync() replays the suspended state onto the freshly built
   backend (`backend_manager_guard.suspend()?`). When the new device does
   not support pausing, the error propagates out of the middle of the
   hotswap while internal state is torn down, and the subsequent unwind
   poisons the backend manager mutex - every later call touching the
   backend (including further setSinkId attempts) then fails with a
   PoisonError, effectively bricking the context. This is directly
   observable through WPT audiocontext-sinkid-state-change.https.html:
   suspend a context, call setSinkId(), and the whole battery collapses.

Reproduction (Linux, any device without pause support - the ALSA "null"
device is a reliable one):

    let context = AudioContext::new(AudioContextOptions::default());
    context.suspend_sync();                 // Err: snd_pcm_pause failed
    context.set_sink_id_sync("...".into()); // PoisonError afterwards

Fix: try the native pause first and fall back to a "suspended" flag when
it fails. The output callback checks the flag, emits equilibrium
(silence) and returns without invoking the renderer, so the rendering
clock stops - semantically equivalent to a paused stream, and the same
strategy browsers use for devices without hardware pause. resume()
clears the flag before attempting play(), and tolerates a play() failure
for flag-suspended devices whose stream never actually stopped.

A regression test running against the default output device is included;
it skips silently when no device is available (headless CI).

### DIFF
--- a/src/io/cpal.rs
+++ b/src/io/cpal.rs
@@ -1,5 +1,5 @@
 //! Audio IO management API
-use std::sync::atomic::Ordering;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::sync::Mutex;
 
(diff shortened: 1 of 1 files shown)


orottier:
This improvement is very welcome but it will have to be reconciled with the RenderThread.suspended feature. The cubeb host could also need fixes.

PR #632, opened 2026-07-16
Avoid FMA in linear ramp sample computation

cmzy:
compute_linear_ramp_sample() used f32::mul_add, which contracts the
multiply and add into a single fused operation with one rounding step.
Browsers evaluate the spec formula v(t) = V0 + (V1 - V0) * k as a plain
multiply followed by an add, i.e. with two rounding steps. The two forms
differ by 1 ulp for many phase values whenever V0 != 0.

This is observable: the WPT k-rate connection suites (for example
k-rate-dynamics-compressor-connections.html) compare a param driven by a
ramp automation from min to max against a param sitting at its intrinsic
min with an audio-rate signal ramping from 0 to max - min connected to
it, and assert *sample-exact* equality. The audio-input path composes
min + round(diff * phase) while the automated path computed
fma(diff, phase, min); with the compressor's ratio (min = 1) and
threshold (min = -100) the outputs diverge by 1 ulp deep into the
render and the exact comparison fails.

Reproduction: render a ramp from -100 to 19 and separately a ramp from
0 to 119, then compare sample-wise `ramp1[i]` against `-100 + ramp2[i]`
with bitwise equality - with FMA they differ on many samples.

Spelling the expression out as `diff * phase + start_value` restores the
two-rounding form; Rust does not re-fuse explicit multiply-add without
fast-math, so the behavior is guaranteed. A bitwise composition test is
included.

### DIFF
--- a/src/param.rs
+++ b/src/param.rs
@@ -61,7 +61,19 @@ fn assert_sequence_length(values: &[f32]) {
     );
 }
 
-// 𝑣(𝑡) = 𝑉0 + (𝑉1−𝑉0) * ((𝑡−𝑇0) / (𝑇1−𝑇0))
+// v(t) = V0 + (V1 - V0) * ((t - T0) / (T1 - T0))
+//
+// Deliberately NOT mul_add: an FMA rounds once, while the natural
+// multiply-then-add rounds twice. WPT's k-rate connection suites (e.g.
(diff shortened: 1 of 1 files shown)


orottier:
This sounds more like a WPT issue. The spec is leading and I don't think we deviate from it

PR #636, opened 2026-07-16
Use the standard separator in stop-before-start panic messages

cmzy:
All validation panics follow the "<DOMException name> - <detail>" shape
(e.g. "InvalidStateError - cannot call `startRendering` twice"), which
embedders rely on to map a caught panic back to the DOMException they
must surface to script. The stop-before-start assertions in
OscillatorNode, ConstantSourceNode and AudioBufferSourceNode read
"InvalidStateError cannot stop before start" - without the " - "
separator - so a parser of that shape cannot split the exception name
from the detail and misclassifies the error as a generic one.

Reproduce: call stop() without a prior start() on any of the three
nodes and inspect the panic message.

Each node gains a should_panic test pinning the exact message.


orottier:
Merged via rebase here https://github.com/orottier/web-audio-api-rs/commit/bb8e373e7b40869b1ed8fb22e3dc23dac2573b4f

PR #639, opened 2026-08-07
Fix - Improve microphone number of channels handling

b-ma:
Hey,

This fixes the https://github.com/ircam-ismm/node-web-audio-api/issues/159 issue

I tried to upgrade cubeb as well but it seems they don't expose frame format other than mono and stereo for now (cf. https://docs.rs/cubeb/latest/cubeb/trait.Frame.html) so I just kept it as it was.

Also added a simple Max patch to test multichannel input.

I had a look that reporting the number of channels of the stream / media stream track, cf.
> "The number of channels of the output corresponds to the number of channels of the MediaStreamTrack."
https://webaudio.github.io/web-audio-api/#MediaStreamAudioSourceNode
https://webaudio.github.io/web-audio-api/#MediaStreamTrackAudioSourceNode

~But I'm not sure how it should be properly handled (maybe through `MediaStreamRenderer`?), probably better if you have a look. In any case this is a different problem that can be handled in another PR I will open an issue~

**edit** I'm may have misunderstood the spec, not sure this has to be reported in control thread actually, let's forget that


orottier:
Thanks!

PR #645, opened 2026-09-03
Remove 'static bound from `decode_audio_data` and `decode_audio_data_sync`

DouglasDwyer:
### Summary

`decode_audio_data` and `decode_audio_data_sync` required `R: 'static`, which forced callers to hand over an owned reader even though the reader is only ever read from during the call. This is overly-restrictive: Symphonia's `MediaSourceStream` and `FormatReader` are lifetime-generic, so nothing actually needs the `'static` bound. This PR removes the bound, so it's possible to decode types like `&[u8]`.

### Changes

- `MediaDecoder` gains a lifetime parameter and holds `Box<dyn FormatReader + 'a>`.
- `decode_media_data`, `decode_audio_data_sync` and `decode_audio_data` drop the `'static` bound on `R`. `decode_audio_data` returns a future bound by the reader's lifetime instead of `'static`.

### Notes

When the reader is `'static` (a `File`, an owned `Vec`/`Cursor`, ...) the future is still `'static` and `Send`, so `tokio::spawn` and the existing thread-pool pattern keep working unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Version History
 
+## Unreleased
+
+- `decode_audio_data` and `decode_audio_data_sync` no longer require the reader to be `'static`,
+  so an audio source that borrows from the stack can be decoded
+
 ## Version 1.7.0 (2026-08-06)
 
 - Fix: Deduplicate connections between the same nodes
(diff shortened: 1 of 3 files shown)


orottier:
Thanks!

PR #651, opened 2026-09-14
Make decode_media_data and AudioBuffer::resample public

DouglasDwyer:
## Summary

I would like a bit more flexibility in how audio is loaded from files. Right now, the main function is [`decode_audio_data_sync`](https://docs.rs/web-audio-api/latest/web_audio_api/context/trait.BaseAudioContext.html#method.decode_audio_data_sync). But this has some limitations:

- It requires having an active `AudioContext` object
- It always resamples the audio buffer to the `AudioContext` playback rate (there is no way to load an `AudioBuffer` at its native sample rate)

As far as I can tell, nothing in `AudioBuffer` needs to be tied to the `AudioContext` playback rate. `AudioBufferSourceNode` is capable of performing resampling (and it resamples using the exact same linear interpolation as `AudioBuffer::resample`) so wanting to load `AudioBuffer`s at their native rate is very reasonable. This also makes it easier from a code architecture perspective because I don't need to have access to an `AudioContext` in order to load audio data (which I want to do from inside the asset system in my game engine).

## Changes

- Make the `decoding` module public
- Expose the `decode_media_data` as a public function
- Expose `AudioBuffer::resample` as a public function


b-ma:
Hey,

> As far as I can tell, nothing in AudioBuffer needs to be tied to the AudioContext playback rate.

Sure, it's more a compromise between having the workload beforehand or at runtime (in some cases).

However it seems to me that rather than exposing (non-spec) methods, the preferred pattern would be to use a dedicated `OfflineAudioContext` configured at whichever sample rate you want in your asset system to load the `AudioBuffer`s. `AudioBuffer`s are not tied to their `BaseAudioContext`, so they can perfectly be decoded / created and played back by 2 different contexts.

DouglasDwyer:
> Hey,
> 
> > As far as I can tell, nothing in AudioBuffer needs to be tied to the AudioContext playback rate.
> 
> Sure, it's more a compromise between having the workload beforehand or at runtime (in some cases).
> 
> However it seems to me that rather than exposing (non-spec) methods, the preferred pattern would be to use a dedicated `OfflineAudioContext` configured at whichever sample rate you want in your asset system to load the `AudioBuffer`s. `AudioBuffer`s are not tied to their `BaseAudioContext`, so they can perfectly be decoded / created and played back by 2 different contexts.

I would be okay with this solution if there were a way to know the source sample rate in advance. There's no way to ask "what is the bitrate of this MP3 file" even with `OfflineAudioContext`, right? Would you be open to exposing a way to do that instead?

DouglasDwyer:
I guess I should also mention that in general it's problematic that resampling is forced. As an example, imagine trying to load two MP3s that have different biteates. It seems wasteful to me to allocate the machinery of two separate `OfflineAudioContext`s just to load each of those files. So I'm not entirely sold on that solution either.

I totally understand wanting to keep close to the WebAudio spec. One other alternative on my end is to use Symphonia directly, I guess. But this is a case where WebAudio falls short - people have requested the feature and it got [put into WebCodecs instead](https://github.com/WebAudio/web-audio-api/issues/30#issuecomment-613351725).

